Issue #21, opened 2022-11-02
Add support to py 3.10


johnnv1:
Related to https://github.com/kornia/kornia/issues/1979

Apparently, `kornia-rs` is not available for python3.10

edgarriba:
i can take this action

johnnv1:
can we also try to have a release for 3.11? So when the torch has a version for it, it is already in plan for us

edgarriba:
Working on this these days

johnnv1:
Done on 0.0.8 version

PR #22, opened 2022-11-03
update packaging for python 3.10

edgarriba:
fixes https://github.com/kornia/kornia-rs/issues/21


johnnv1:
Does it seem not falling because if py310. I think it can be merged and then release a new version for kornia_rs -- and that clears the way for ensure py39 and py310 for kornia

edgarriba:
@johnnv1 i'm checking if i can quickly fix this one but it needs a bit of investigation since the libjpegturbo package is causing some issues.

PR #23, opened 2022-12-31
Default jpegturbo, udpate version to 0.0.6 and more improvements

edgarriba:
- add support to python 3.11
- turbojpeg is not optional anymore
- bump version package to 0.0.6
- update to latest all the requirements
- release image uses now `quay.io/pypa/manylinux2014_x86_64`
- improve ownership of PyCapsule
- Include new structures: `ImageSize`, `ImageEncoder`, `ImageDecoder`
- refactor `read_image_jpeg`, `write_image_jpeg`


edgarriba:
@johnnv1 still need to polish a few things but happy to hear api feedback

.github/workflows/release.yml
```diff
@@ -12,7 +12,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        python-version: [3.7, 3.8, 3.9]
+        python-version: [3.7, 3.8, 3.9, 3.10]
```
edgarriba, line 15:
```suggestion
        python-version: [3.7, 3.8, 3.9, 3.10.0]
```

.github/workflows/release.yml
```diff
@@ -38,7 +38,7 @@ jobs:
     runs-on: macos-latest
     strategy:
       matrix:
-        python-version: [3.7, 3.8, 3.9]
+        python-version: [3.7, 3.8, 3.9, 3.10]
```
edgarriba, line 41:
```suggestion
        python-version: [3.7, 3.8, 3.9, 3.10.0]
```

.github/workflows/release.yml
```diff
@@ -65,7 +65,7 @@ jobs:
     runs-on: windows-latest
     strategy:
       matrix:
-        python-version: [3.7, 3.8, 3.9]
+        python-version: [3.7, 3.8, 3.9, 3.10]
```
edgarriba, line 68:
```suggestion
        python-version: [3.7, 3.8, 3.9, 3.10.0]
```

johnnv1: why `.0`?

edgarriba: seems that somehow gh actions detects 3.1 if i don't explicitly set the patch version

src/io.rs
```diff
@@ -1,51 +1,122 @@
...
+
+#[pymethods]
+impl ImageDecoder {
+    #[new]
+    pub fn new() -> ImageDecoder {
+        let decompressor = turbojpeg::Decompressor::new().unwrap();
+        ImageDecoder { decompressor }
+    }
+
+    pub fn read_header(&mut self, jpeg_data: &[u8]) -> ImageSize {
```
johnnv1, line 64: This will be useful outside the class? (also, not sure if this suggestion works here)
```suggestion
    pub(self) fn read_header(&mut self, jpeg_data: &[u8]) -> ImageSize {
```

src/lib.rs
```diff
@@ -9,11 +9,12 @@ pub mod io;
 #[allow(dead_code)]
 pub mod tensor;
 
-use crate::dlpack_py::cvtensor_to_dlpack;
-use crate::io::read_image_rs;
-
-#[cfg(feature = "libjpeg-turbo")]
 use crate::io::read_image_jpeg;
+use crate::io::read_image_rs;
+use crate::io::write_image_jpeg;
```
johnnv1, line 14: Personally, I would like to have simpler names for these methods in kornia_rs -- like just `read_image` and `write_image` and have a flag/enum/anything to choice the encoder type

src/lib.rs
```diff
@@ -37,12 +38,13 @@ pub fn get_version() -> String {
...
     m.add("__version__", get_version())?;
+    m.add_function(wrap_pyfunction!(read_image_jpeg, m)?)?;
+    m.add_function(wrap_pyfunction!(write_image_jpeg, m)?)?;
     m.add_function(wrap_pyfunction!(read_image_rs, m)?)?;
-    m.add_function(wrap_pyfunction!(cvtensor_to_dlpack, m)?)?;
     m.add_class::<tensor::cv::Tensor>()?;
-
-    #[cfg(feature = "libjpeg-turbo")]
-    m.add_function(wrap_pyfunction!(read_image_jpeg, m)?)?;
+    m.add_class::<ImageSize>()?;
```
johnnv1, line 45: Will be useful for something in the outside?

test/test_io.py
```diff
@@ -35,6 +30,24 @@ def test_read_image_rs():
...
     # convert to dlpack to import to torch
-    dlpack = K.cvtensor_to_dlpack(cv_tensor)
-    th_tensor = torch.utils.dlpack.from_dlpack(dlpack)
+    th_tensor = torch.utils.dlpack.from_dlpack(cv_tensor)
     assert th_tensor.shape == (195, 258, 3)
+
+
+def test_decompress():
+    # load an image with libjpeg-turbo
+    img_path: Path = DATA_DIR / "dog.jpeg"
```
johnnv1, line 39: Have a fixture for this path will be better to maintain

test/test_io.py
```diff
@@ -35,6 +30,24 @@ def test_read_image_rs():
...
+    tensor_rs = K.read_image_jpeg(str(img_path))
+    image = np.from_dlpack(tensor_rs)
```
johnnv1, line 41: We can have an assert here? or maybe in another test with a small and simpler image (a black/white square)

test/test_io.py
```diff
@@ -35,6 +30,24 @@ def test_read_image_rs():
...
+    image_encoded: list = image_encoder.encode(image.tobytes(), image.shape)
+    K.write_image_jpeg(str(DATA_DIR/"dog_rs.jpeg"), image_encoded)
+
+    image_decoder = K.ImageDecoder()
+    image_size: K.ImageSize = image_decoder.read_header(bytes(image_encoded))
+    assert image_size.width == 258
+    assert image_size.height == 195
+
+    decoded_tensor: K.cvTensor = image_decoder.decode(bytes(image_encoded))
+    decoded_image = np.from_dlpack(decoded_tensor)
```
johnnv1, line 53: same for the decoder

src/linalg.rs
```diff
@@ -0,0 +1,69 @@
+use pyo3::prelude::*;
```
johnnv1, line 1: This isn't in the kornia_rs module, and also don't have tests

johnnv1 (commented):
I just know the basic of rust, so, just some comments/questions :)

.github/workflows/release.yml
```diff
@@ -12,22 +12,19 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        python-version: [3.7, 3.8, 3.9]
+        python-version: ["3.7", "3.8", "3.9", "3.10", "3.11"]
         target: [x86_64]
     steps:
+      - uses: egor-tensin/setup-clang@v1
```
edgarriba, line 18:
```suggestion
```

.github/workflows/release.yml
edgarriba: the trick was to define as strings

.github/workflows/release.yml
```diff
@@ -12,7 +12,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        python-version: [3.7, 3.8, 3.9]
+        python-version: [3.7, 3.8, 3.9, 3.10.0]
```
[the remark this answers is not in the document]
edgarriba: done

src/io.rs
```diff
@@ -1,51 +1,122 @@
...
+            pitch: 3 * image_size.width, // we use no padding between rows
+            height: image_size.height,
+            format: turbojpeg::PixelFormat::RGB,
+        };
+
+        // decompress the JPEG data
+        self.decompressor.decompress(jpeg_data, image).unwrap();
+
+        // return the raw pixel data and shape as Tensor
+        let shape = vec![image_size.height as i64, image_size.width as i64, 3];
```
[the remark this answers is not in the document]
edgarriba: it;s a requirement from dlpack protocol

src/io.rs
```diff
@@ -1,51 +1,122 @@
...
+}
+
+impl Default for ImageEncoder {
+    fn default() -> Self {
+        Self::new()
+    }
+}
+
+#[pymethods]
+impl ImageDecoder {
```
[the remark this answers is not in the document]
edgarriba: not now but later maybe yes. The initial intention was to directly expose (de)compressor but it throw too many errors, so i went for this light version

src/io.rs
edgarriba: i have some uses cases at work where i want to just check image sizes from the bytes stream and don't want to decode the whole stream

src/lib.rs
edgarriba: my idea idea for kornia_rs was to provide very low level and explicit operators so that is easier later to be implemented in any python library (in the direction of low level dll ops libraries). Also i believe it's easier from kornia python side to have high level `read_image` or `write_image`  and we dispatch based on filenames or similar

src/lib.rs
edgarriba: for now is only to handle the result from header. But i have also in mind to wrap somehow image class that could use this. We need to iterate yet on that api

test/test_io.py
edgarriba: i will redo this tests possibly creating a synthetic image or something. It can be tricky thought becouse of the compression quality

src/linalg.rs
edgarriba: yeah. that's an artifact for a subsequent pr i'll prepare. Wanna explore to expose some basic ops in torch via binings

.pre-commit-hooks.yaml
```diff
@@ -0,0 +1,22 @@
+- id: fmt
```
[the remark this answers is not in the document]
edgarriba: still playing with it but this can be tricky because some rust check requires compiling inside the devel docker container

src/lib.rs
johnnv1: fair

.pre-commit-hooks.yaml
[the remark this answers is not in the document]
johnnv1: in this case, maybe the best way is to keep with your original script

.pre-commit-config.yaml
```diff
@@ -0,0 +1,41 @@
...
+      types: [rust]
+      pass_filenames: false
+    - id: clippy
+      name: clippy
+      description: Lint rust sources
+      entry: cargo clippy
+      language: system
+      args: ["--", "-D", "warnings"]
+      types: [rust]
+      pass_filenames: false
```
johnnv1, line 41:
```suggestion
  - repo: https://github.com/doublify/pre-commit-rust
    rev: v1.0
    hooks:
    - id: fmt
    - id: cargo-chec
    - id: clippy
```

.github/workflows/ci.yml
```diff
@@ -18,62 +18,64 @@ jobs:
...
+      credentials:
+          username: ${{ github.actor }}
+          password: ${{ secrets.GITHUB_TOKEN }}
+    steps:
+      -
+        name: Checkout the repository
+        uses: actions/checkout@v3
+      -
+        name: Run cargo check
+        run: pre-commit run cargo-check --all-files
```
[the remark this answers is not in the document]
edgarriba: trying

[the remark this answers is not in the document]
edgarriba: ok, i need to create an image that has bash entrypoint. Will do in a next PR when i support aarch64

PR #61, opened 2024-03-23
Add `io::WebcamCapture` via gstreamer

edgarriba:
- implements a simple gstreamer pipeline to capture frames from a webcam
- the code lives in `webcam.rs` as `WebcamCapture` with its own builder


Cargo.toml
```diff
@@ -10,12 +10,14 @@ license-file = "LICENSE"
...
+fast_image_resize = "3.0.4"
+gst = { version = "0.22.2", package = "gstreamer" }
+gst-app = { version = "0.22.0", package = "gstreamer-app" }
 image = { version = "0.25.0" }
-turbojpeg = {version = "1.0.0"}
 memmap2 = "0.9.4"
 ndarray = { version = "0.15.6", features = ["rayon"] }
-fast_image_resize = "3.0.4"
+num-traits = "0.2.17"
+turbojpeg = {version = "1.0.0"}
```
Wumpf, line 20: might be worth considering putting different image file format supports behind feature flags so one can opt out of some of these dependencies (jpeg would be a good candidate for a default feature ofc)

examples/webcam.rs
```diff
@@ -0,0 +1,47 @@
+use kornia_rs::{image::ImageSize, io::webcam::WebcamCaptureBuilder};
+
+fn main() -> Result<(), Box<dyn std::error::Error>> {
+    // start the recording stream
+    let rec = rerun::RecordingStreamBuilder::new("Kornia App")
+        // set the batcher config to flush every 10 bytes to improve performance
+        .batcher_config(rerun::log::DataTableBatcherConfig {
+            flush_num_bytes: 10,
```
Wumpf, line 8: 10bytes is a bit extreme, that effectively disables the batcher, meaning every interaction with the recording stream will right away send to the viewer which can be overall much slower.
The default is to send either every 1mb or ever 8ms. So if the default is still a problem running in release mode it might make actually more sense to lower `flush_tick`.
I guess alternatively you could also make it very explicit and go with `rerun::log::DataTableBatcherConfig::ALWAYS`, that's more honest than 10bytes 😄

examples/webcam.rs
```diff
@@ -0,0 +1,47 @@
...
+
+        // convert the image to f32 and normalize before processing
+        let img = img.cast_and_scale::<f32>(1. / 255.).unwrap();
+
+        // convert the image to grayscale and binarize
+        let gray = kornia_rs::color::gray_from_rgb(&img).unwrap();
+        let bin = kornia_rs::threshold::threshold_binary(&gray, 0.5, 1.0).unwrap();
+
+        // log the image
+        let _ = rec.log("binary", &rerun::Image::try_from(bin.data).unwrap());
```
Wumpf, line 43: is it intentional to swallow the error here via `let _ =`? Generally, `.ok()` is better imho since it's more explicit about ignoring the error.

On a general note all these `.unwrap` in this file make me nervous. Having everything be a hard crash is pretty awful imho. it would be nice if `webcam.run` supports forwarding errors for when they're critical enough to stop the running and otherwise handle the errors in place e.g. with logging a message etc. so that the application can continue when it makes sense.

src/io/webcam.rs
```diff
@@ -0,0 +1,247 @@
...
+    ///
+    /// # Arguments
+    ///
+    /// * `camera_id` - The camera id used for capturing images
+    /// * `size` - The image size used for resizing directly from the camera
+    ///
+    /// # Returns
+    ///
+    /// A WebcamCapture object
+    fn new(camera_id: usize, size: Option<ImageSize>) -> Result<Self> {
```
Wumpf, line 102: to an extent a matter of taste, but I'd only reach to `anyhow::Result` definitely no longer to be handled programmatically and 100% a user or log facing message. The problem with anyhow is that one can't match on the different error outcomes. Instead, I prefer `thiserror` errors whenever it makes sense, and I think here it would be a good fit.

src/io/webcam.rs
```diff
@@ -0,0 +1,247 @@
...
+
+        // Event loop or processing here
+        // Set pipeline state to Null when done
+
+        let bus = pipeline
+            .bus()
+            .ok_or_else(|| anyhow::anyhow!("Failed to get bus"))?;
+
+        let handle = std::thread::spawn(move || {
+            for msg in bus.iter_timed(gst::ClockTime::NONE) {
```
Wumpf, line 148: what does `ClockTime::NONE` do? Hope that's not busy looping 😄

src/io/webcam.rs
```diff
@@ -0,0 +1,247 @@
...
+            .dynamic_cast::<gst_app::AppSink>()
+            .map_err(|_| anyhow::anyhow!("Failed to cast to AppSink"))?;
+
+        let (tx, rx) = std::sync::mpsc::channel();
+
+        appsink.set_callbacks(
+            gst_app::AppSinkCallbacks::builder()
+                .new_sample(move |sink| match Self::extract_image_frame(sink) {
+                    Ok(frame) => {
+                        if let Err(e) = tx.send(frame) {
```
Wumpf, line 123: if one doesn't start processing messages via `run`, would this keep on populating the channel until the application runs out of memory?

src/io/webcam.rs
```diff
@@ -0,0 +1,247 @@
...
+    ///
+    /// * `f` - A function that takes an image frame
+    pub fn run<F>(&self, f: F) -> Result<()>
+    where
+        F: Fn(Image<u8, 3>),
+    {
+        let receiver = &self.receiver;
+        while let Ok(img) = receiver.recv() {
+            f(img);
+        }
```
Wumpf, line 184: looks like there's no way for `f` to stop this right now, that seems problematic!

src/io/webcam.rs
```diff
@@ -0,0 +1,247 @@
...
+    /// Extracts an image frame from the appsink
+    ///
+    /// # Arguments
+    ///
+    /// * `appsink` - The AppSink
+    ///
+    /// # Returns
+    ///
+    /// An image frame
+    fn extract_image_frame(appsink: &gst_app::AppSink) -> Result<Image<u8, 3>> {
```
Wumpf, line 221: some concern about anyhow::Result applies here

src/io/webcam.rs
```diff
@@ -0,0 +1,247 @@
...
+            handle: Some(handle),
+        })
+    }
+
+    /// Runs the webcam capture object and grabs frames from the camera
+    ///
+    /// # Arguments
+    ///
+    /// * `f` - A function that takes an image frame
+    pub fn run<F>(&self, f: F) -> Result<()>
```
Wumpf, line 177: Right now this can't fail so why the Result. (I do agree this should be able to fail :))

Wumpf (commented):
basic structure lgtm given my little understanding of this domain 😄 
the error handling and the lack of easy cancel & memory control makes me a bit uneasy though

src/io/webcam.rs
edgarriba: i've been [experimenting](https://github.com/kornia/kornia-rs/compare/this-error?expand=1) a bit with thiserror and i feel  that is a bit impractical here because i have to be mapping all the time all the defined errors by the methods of the other libraries. It give more semantics yes, but seems a bit of a blocker for development at least now. And specially because this crate is more an aggregation of libraries, and thiserror seems pretty well defined when you start a very foundational crate with almost none dependencies.

edgarriba:
@Wumpf @gau-nernst would be great if you could try the examples in your computer

PR #64, opened 2024-03-25
add `compute_histogram` python bindings

edgarriba:
- expose `compute_histogram` to python
- rename `py-kornia` -> `kornia-py` in the direction to homogenize internal crates


johnnv1 (commented):
nice

edgarriba:
@johnnv1 have you noticed somewhere else torch with macOS failing to download?

Issue #47, opened 2024-03-12
implement `warp_affine` / `warp_perspective`


edgarriba:
reference:
- [x] warp_affine: https://github.com/kornia/kornia/blob/main/kornia/geometry/transform/imgwarp.py#L131
- [x] warp_perspective: https://github.com/kornia/kornia/blob/main/kornia/geometry/transform/imgwarp.py#L48

needs (to abstract / reuse code)
- [x] grid_sample / map_coordinates

gau-nernst:
I recently had to implement warp_affine from scratch in Dart to do image transformation in Flutter without OpenCV. I could contribute warp_affine here if you are okay with that. It's probably not gonna be super fast.

grid_sample is not needed. The current `bilinear_interpolation()` function should be enough.

edgarriba:
feel free to go for it -- about grid_sample, it will basically a way to reuse code. If you see in the [resize_native](https://github.com/kornia/kornia-rs/blob/main/src/resize.rs#L159) function the `ndarray::Zip` is basically parallelizing over the destination points. In any case you will have to reimplement that to make it efficient, we don't want regular looping but instead iterators (which is what's happening in that code already.

gau-nernst:
Interesting. Do you have benchmarks between naive for-loop version and the current version?

Some observations
- Currently we materialize the whole index array (width x height). This is quite wasteful.
- There is probably overhead for parallelization (from what I see, ndarray uses rayon https://docs.rs/ndarray/latest/ndarray/parallel/index.html). Perhaps we can parallelize over the rows only instead of all the points to reduce overhead.

In any case, let's get a working implementation first before looking more into the speed.

edgarriba:
done

PR #67, opened 2024-03-29
Add Warp affine

gau-nernst:
Partially address #47

Questions and problems:

1. From Python, do we support images with n_channels != 3? The following code fails

```python
import kornia_rs as K
import numpy as np

img = np.random.randint(0, 255, size=(256, 256, 1), dtype=np.uint8)
K.resize(img, (40, 40), "bilinear")
```

Error

```
Traceback (most recent call last):
  File "[redacted]/kornia-rs/kornia-py/test.py", line 8, in <module>
    K.resize(img, (40, 40), "bilinear")
Exception: Data length (65536) does not match the image size (196608)
```

Currently my warp_affine python function also only works for 3-channel, since the compiler forces me to add type annotation when doing conversion `PyImage` -> `Image`. 

```rust
    let image: Image<u8, 3> = Image::from_pyimage(image)
        .map_err(|e| PyErr::new::<pyo3::exceptions::PyException, _>(format!("{}", e)))?;
```

I think this is caused by image type casting, since your resize function doesn't need annotation? (I need to cast image from `u8` to `f32` since our native interpolation only works for `f32`).

2. Continue from above, the native interpolation should work for both `u8` and `f32` (and potentially other numeric types like `u16`?). For this to happen, we can create a trait that can cast into `f32` (when reading value from array) and cast back to the original type (when writing value to array). I looked at some existing crates for numeric traits but none seem to have this specific trait. We can create our own trait for this. There is also `From` and `Into` (need to check overflow/underflow behavior).

3. For resizing, it seems like your logic is similar to PyTorch's `align_corners=True`? I think my implementation for warp affine is not quite `align_corners=True`? Need to check more and check against other implementations.

For future extension, we can implement a struct `AffineTransform` so that we can do computation with it easier (i.e. invert, apply multiple affine transforms, extract rotation/scale param....). Right now I use a tuple of 6 numbers so that the Python binding doesn't need to depend on `ndarray` crate.

Some outputs

```python
import kornia_rs as K
import cv2

import numpy as np

img_path = "../tests/data/dog.jpeg"
img: np.ndarray = K.read_image_jpeg(img_path)

# check the image properties
assert img.shape == (195, 258, 3)

matrix = (1.0, 0.0, -20.0, 0.0, 1.0, 10.0)  # translation
# matrix = (2.0, 0.0, 0.0, 0.0, 2.0, 0.0)  # scale from (0, 0)
# matrix = tuple(cv2.getRotationMatrix2D((258 / 2, 195 / 2), 45.0, 1.0).flatten())  # rotation around center

img_resized = K.warp_affine(img, matrix, (195, 258), "bilinear")
```

image
image
image


src/geometry/transform.rs
```diff
@@ -0,0 +1,191 @@
+use crate::image::{Image, ImageSize};
+use crate::resize::{interpolate, meshgrid, InterpolationMode};
+use anyhow::Result;
+use ndarray::stack;
+
+type AffineMatrix = (f32, f32, f32, f32, f32, f32);
```
edgarriba, line 6: For small matrix multiplication I was planning to use the nalgebra crate which seems well optimizadles for that

src/geometry/transform.rs
```diff
@@ -0,0 +1,191 @@
...
+    src: &Image<f32, CHANNELS>,
+    m: AffineMatrix,
+    new_size: ImageSize,
+    interpolation: InterpolationMode,
+) -> Result<Image<f32, CHANNELS>> {
+    // invert affine transform matrix to find corresponding positions in src from dst
+    let m_inv = invert_affine_transform(m);
+
+    // create the output image
+    let mut output = Image::from_size_val(new_size, 0.0)?;
```
edgarriba, line 40: If want to to get your hand here , I was planning to add a from_size_empty

src/geometry/transform.rs
```diff
@@ -0,0 +1,191 @@
...
+
+    // create a grid of x and y coordinates for the output image
+    let x = ndarray::Array::range(0.0, new_size.width as f32, 1.0).insert_axis(ndarray::Axis(0));
```
edgarriba, line 43: Add a todo for adding a create mesh grid function to be reused

src/resize.rs
```diff
@@ -116,6 +106,19 @@ pub enum InterpolationMode {
     Nearest,
 }
 
+pub(crate) fn interpolate(
```
edgarriba, line 109:
```suggestion
pub(crate) fn interpolate_pixel(
```
I’m trying to avoid high level names for names. Generic apis names will come later once everything is mature

kornia-py/src/geometry/transform.rs
edgarriba, line 1: About namespace, I want to make a distinction because in future there will be 3d or calibration stuff. Don’t want to copy Kornia torch namespaces, it got to convoluted there. That’s why I’m putting for now everything in the top level and layer in future versions we’ll group based on the evolution of the library. I’d keep for now as ´warp.rs’

kornia-py/src/geometry/transform.rs
```diff
@@ -0,0 +1,42 @@
...
+        }
+    };
+
+    // we need to cast to f32 for now since kornia-rs interpolation function only works with f32
+    let image = image.cast::<f32>().unwrap();
+    let image = kornia_rs::geometry::transform::warp_affine(&image, m, new_size, interpolation)
+        .map_err(|e| PyErr::new::<pyo3::exceptions::PyException, _>(format!("{}", e)))?;
+
+    // NOTE: for bicubic interpolation (not implemented yet), f32 may overshoot 255
+    let image = image.cast::<u8>().unwrap();
```
edgarriba, line 39: No unwraps

kornia-py/src/geometry/transform.rs
```diff
@@ -0,0 +1,42 @@
...
+
+#[pyfunction]
+pub fn warp_affine(
+    image: PyImage,
+    m: (f32, f32, f32, f32, f32, f32),
+    new_size: (usize, usize),
+    interpolation: &str,
+) -> PyResult<PyImage> {
+    // have to add annotation Image<u8, 3>, otherwise the compiler will complain
+    // NOTE: do we support images with channels != 3?
```
edgarriba, line 14: I didn’t need but we can adjust as we need here

edgarriba (changes requested):

edgarriba (commented):
Can you add a bench for both python and rust ? I’m measuring a bit everything I put here

kornia-py/src/geometry/transform.rs
gau-nernst: without unwraps it won't compile

gau-nernst:
Added benchmark. Initial results are pretty positive considering the implementation is pretty naive. (I run `maturin develop -r` to install "release" version)

```
OpenCV: 0.22 ms
PIL: 0.61 ms
Kornia: 0.50 ms
```

It might not be fair with OpenCV and PIL, since ndarray::zip uses multi-threading via rayon, while I think OpenCV and PIL don't. For an image processing pipeline, single-thread performance might be more important, since there will be multiple processes running concurrently.

kornia-py/src/warp.rs
```diff
@@ -0,0 +1,42 @@
...
+    interpolation: &str,
+) -> PyResult<PyImage> {
+    // have to add annotation Image<u8, 3>, otherwise the compiler will complain
+    // NOTE: do we support images with channels != 3?
+    let image: Image<u8, 3> = Image::from_pyimage(image)
+        .map_err(|e| PyErr::new::<pyo3::exceptions::PyException, _>(format!("{}", e)))?;
+
+    let new_size = kornia_rs::image::ImageSize {
+        width: new_size.0,
+        height: new_size.1,
```
gau-nernst, line 20: I copied this from `resize.rs`. It was `height: new_size.0, width: new_size.1` originally. Was this a typo? i.e. For the `new_size`, do you use (width, height) or (height, width)?

edgarriba: it should be height width

kornia-py/src/geometry/transform.rs
edgarriba: You need to use propagate the error with ‘?’

gau-nernst: Just updated. Is that what you want? I don't have much experience with rust. I hope you could be more explicit with what you want.

edgarriba:
@gau-nernst for benchmark use RUSTFLAGS="-C target-cpu=native" cargo bench read_jpeg for better optimiser. Regarding parallelism I still exploring what’s the best strategy, I’m using mostly the lib in rust which that’s not an issue for me. Opencv I think by default uses Opencl which is  way to implement parallelism. I’m exploring in the background also a safe multithreaded PyTorch dataloader for rust that will be exposed to python. But happy to investigate and put numbers to decide the best’s strategy, maybe we need different backend implementations depending on the use case.

edgarriba:
In the long run I think that for speed we’ll implement the algorithms when cuda in rust is officially supported

edgarriba (approved):

PR #68, opened 2024-03-29
[prototype] `ImageDtype` trait to implement resize for various dtypes

gau-nernst:
This is my prototype to support resize with different dtypes. Justifications for trait

- `Copy` trait: to get the value out of `ndarray::Array<T, _>`
- `Default` trait: to get 0 value
- `Into<f32>` trait: to convert to f32 for internal computation i.e. bilinear/bicubic
- `Send` and `Sync` trait: required for `ndarray::Zip::par_for_each()`
- `from_f32()` function: convert from internal computation dtype f32 back to original dtype. We can't use `From<f32>` since it is undefined-behavior with underflow and overflow values i.e. we need to clamp the value by ourselves.

I think `ImageDtype` trait can be moved to `image.rs` and data within `Image` must implement `ImageDtype` trait i.e. these are the dtype that we support -> `pub struct Image<T: ImageDtype, const CHANNELS: usize>`

### DIFF
--- a/src/resize.rs
+++ b/src/resize.rs
@@ -39,6 +39,29 @@ pub fn meshgrid(x: &Array2<f32>, y: &Array2<f32>) -> (Array2<f32>, Array2<f32>) 
     (xx, yy)
 }
 
+trait ImageDtype {
+    fn to_f32(&self) -> f32;
+    fn from_f32(x: f32) -> Self;
+}
+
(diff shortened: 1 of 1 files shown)


edgarriba (changes requested):
Let’s somehow verify we don’t break anything , maybe add different types test

edgarriba (approved):

PR #69, opened 2024-04-02
add rotate example

edgarriba:
https://github.com/kornia/kornia-rs/assets/5157099/59cb8fef-0e5c-4cd8-b502-a1fbae0c4af2

### DIFF
--- a/examples/rotate.rs
+++ b/examples/rotate.rs
@@ -1,0 +1,38 @@
+use kornia_rs::image::Image;
+use kornia_rs::io::functional as F;
+
+fn main() -> Result<(), Box<dyn std::error::Error>> {
+    // read the image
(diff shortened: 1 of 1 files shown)


edgarriba:
@gau-nernst

gau-nernst:
There is one problem with my implementation that I forgot to mention. As you can see, the edges are jagged. This is because I skip pixels if they are outside of the original image, which will create aliasing effect. To make smooth edges, we need add different modes of handling the border (1) clamp to edge of the image (as right now) is suitable for resizing, and (2) return zero so that we can still interpolate the few pixels outside the image.
I think for now we can just open an issue to keep track of the problem.
It will be good to also look into OpenCV implementation.

edgarriba:
Yeah, I’m also looking at https://github.com/rust-lang/portable-simd because there are a lot of operations that can be vectorized

PR #75, opened 2024-05-31
implement warp_perspective

edgarriba:
- make `warp` as submodule
- create `warp_perspective` as public function


src/warp/affine.rs
```diff
@@ -142,6 +142,7 @@ pub fn warp_affine<const CHANNELS: usize>(
             let u_src = m_inv.0 * u + m_inv.1 * v + m_inv.2;
             let v_src = m_inv.3 * u + m_inv.4 * v + m_inv.5;
 
+            // TODO: remove -- this is already done in interpolate_pixel
```
edgarriba, line 145: @gau-nernst found this during the implementation of this patch

PR #88, opened 2024-07-01
add zeros, and compute offset

edgarriba:
Implement:
- `from_shape_val`
- `from_shape_fn`
- `get_iter_offset`
- `get_unchecked`
- `zeros`
- `map`
- `cast`
- aliases for `Tensor1`, `Tensor2`, `Tensor3`, `Tensor4`


src/tensor/base.rs
```diff
@@ -75,12 +75,108 @@ impl<T, const N: usize> Tensor<T, N> {
...
+    /// ```
+    /// use kornia_rs::tensor::Tensor;
+    ///
+    /// let t = Tensor::<u8, 1>::from_shape_fn([4], |[i]| i as u8);
+    /// assert_eq!(t.data, vec![0, 1, 2, 3]);
+    ///
+    /// let t = Tensor::<u8, 2>::from_shape_fn([2, 2], |[i, j]| (i * 2 + j) as u8);
+    /// assert_eq!(t.data, vec![0, 1, 2, 3]);
+    /// ```
+    pub fn from_shape_fn<F>(shape: [usize; N], f: F) -> Self
```
johnnv1, line 141: nice

PR #97, opened 2024-07-11
Refactor crates

edgarriba:
- this is a **BREAKING COMMIT** towards 0.2
- converted package to `workspace`
- main crates now are `kornia-core` / `kornia-image` / `kornia-io` / `kornia-imgproc` / `kornia`
- the main namespace `kornia_rs` is now just `kornia`
- moved the main crates into `crates/*`
- examples have been factored to their own crates
- benches have been moved to their own subcrates


examples/normalize.rs
```diff
@@ -1,5 +1,5 @@
-use kornia_rs::image::Image;
-use kornia_rs::io::functional as F;
+use kornia_image::Image;
+use kornia_io::functional as F;
```
johnnv1, line 2: why not keep using the top lib in the examples?

kornia-serve/Cargo.toml
```diff
@@ -6,7 +6,7 @@ edition = "2021"
 [dependencies]
 axum = "0.7.5"
 env_logger = "0.11.3"
-kornia-rs = "0.1.3"
+kornia-rs = { workspace = true }
```
johnnv1, line 9: when shipping it, will it be the right version this way?

benches/bench_color.rs
```diff
@@ -1,7 +1,7 @@
 use criterion::{black_box, criterion_group, criterion_main, BenchmarkId, Criterion};
```
johnnv1, line 1: I would separate the benchmarks into subdirectories for each crate hre

.gitignore
```diff
@@ -3,3 +3,4 @@
 target/
 docker/
 examples/*/Cargo.lock
+Cargo.lock
```
johnnv1, line 6: it's here, but the file is still on git

examples/normalize.rs
edgarriba: still figuring out how to recover the top level api -- maybe it's gonna be a big breaking change. Not clear yet

edgarriba: Fixed

README.md
```diff
@@ -123,7 +123,7 @@ fn main() -> Result<(), Box<dyn std::error::Error>> {
 
     let gray_resize: Image<f32, 1> = kornia_rs::resize::resize_native(
         &gray,
-        kornia_rs::image::ImageSize {
+        kornia_image::ImageSize {
```
edgarriba, line 126: Check imports in docs

crates/kornia-imgproc/src/color/gray.rs
```diff
@@ -33,7 +33,7 @@ const BW: f64 = 0.114;
 ///     vec![0f32; 4 * 5 * 3],
 /// )
 /// .unwrap();
-/// let gray: Image<f32, 1> = kornia_rs::color::gray_from_rgb(&image).unwrap();
+/// let gray: Image<f32, 1> = kornia_imgproc::color::gray_from_rgb(&image).unwrap();
```
edgarriba, line 36: Check Import

mweinelt:
Was the removal of Cargo.lock intentional? For nixpkgs this creates a problem, since we now need to provide our own to be able to reproducibly fetch the same dependencies.

edgarriba:
We can add again

mweinelt:
That would be appreciated.

PR #104, opened 2024-08-17
onnxruntime example for rtdetr

edgarriba:
07/09/2024: finally working with rtdetr

image (2)


examples/onnx/src/main.rs
```diff
@@ -0,0 +1,63 @@
...
+        InterpolationMode::Bilinear,
+    )?;
+
+    let image = image_resized.clone().cast_and_scale::<f32>(1. / 255.)?;
+
+    let image_ncwh = image.to_tensor_nchw();
+    let image_ncwh = image_ncwh.as_standard_layout().to_owned();
+
+    // run the model
+    let outputs = model.run(ort::inputs![
```
johnnv1, line 49: further we should timeit to check how it goes

edgarriba: sure, in principle shouldn't be much different that the python one since its the same c++ backend in the end

PR #116, opened 2024-08-31
Ndarray deprecation 

edgarriba:
- complete deprecation of ndarray in replacement of Tensor

### DIFF
--- a/crates/kornia-image/src/image.rs
+++ b/crates/kornia-image/src/image.rs
@@ -1,5 +1,3 @@
-use std::ops;
-
 use kornia_core::{CpuAllocator, SafeTensorType, Tensor3};
 
 use crate::error::ImageError;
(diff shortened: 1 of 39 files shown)


crates/kornia-image/src/image.rs
```diff
@@ -145,16 +122,16 @@ where
     /// ```
     pub fn new(size: ImageSize, data: Vec<T>) -> Result<Self, ImageError> {
         // check if the data length matches the image size
-        if data.len() != size.width * size.height * CHANNELS {
+        if data.len() != size.width * size.height * C {
```
edgarriba, line 125: Use .rows , .cols

crates/kornia-image/src/image.rs
```diff
@@ -316,9 +293,9 @@ where
         self.rows()
     }
 
-    /// Get the number of channels in the image.
+    /// Get the number of C in the image.
```
edgarriba, line 296:
```suggestion
    /// Get the number of channels in the image.
```

Issue #103, opened 2024-08-17
Update hflip and vflip with mutable signatures 


edgarriba:
After this refactor https://github.com/kornia/kornia-rs/pull/102 `hflip a and `vflip` were pending to be completed

emilmgeorge:
Hi, 

I'm interested in taking this up. 

As I understand from the linked PR, we have to use a mutable `dst` parameter for the output instead of returning, and also adjust the loop slightly for that. Please let me know if there is anything else I should consider. I have started with hflip function and hope to have something for your review soon.

edgarriba:
@emilmgeorge yes, that's it. Beware that this morning i merged the last of series of PR to ndarray as dep. See: https://github.com/kornia/kornia-rs/pull/116

you might have to rebase

emilmgeorge:
Thanks, yes I saw it just after I posted my prev comment. I have since rebased and updated my changes. Will PR now.

edgarriba:
merged in #117

PR #117, opened 2024-09-01
Refactor hflip, vflip functions to allow preallocation

emilmgeorge:
For #103


crates/kornia-imgproc/src/flip.rs
```diff
@@ -97,31 +100,44 @@ where
...
+    if src.size() != dst.size() {
+        return Err(ImageError::InvalidImageSize(
+            src.cols(),
+            src.rows(),
+            dst.cols(),
+            dst.rows(),
+        ));
+    }
 
     // TODO: improve this implementation
```
emilmgeorge, line 128: I retained this for your review. (I'm very new to rayon so I'm not sure if the below logic has any inefficiencies).

emilmgeorge:
Fixed formatting.

crates/kornia-imgproc/src/flip.rs
```diff
@@ -27,43 +33,60 @@ use rayon::{iter::ParallelIterator, slice::ParallelSliceMut};
 /// )
 /// .unwrap();
 ///
-/// let flipped: Image<f32, 3> = horizontal_flip(&image).unwrap();
+/// let mut flipped = Image::<f32, 3>::from_size_val(
```
edgarriba, line 36:
```suggestion
/// let mut flipped = Image::<f32, 3>::from_size_val(image.size()
```
im adding this as a good practice to avoid shape issues

crates/kornia-imgproc/src/flip.rs
```diff
@@ -27,43 +33,60 @@ use rayon::{iter::ParallelIterator, slice::ParallelSliceMut};
...
+        ));
+    }
 
     dst.as_slice_mut()
         .par_chunks_exact_mut(src.cols() * C)
-        .for_each(|row| {
-            let mut i = 0;
-            let mut j = src.cols() - 1;
-            while i < j {
+        .zip_eq(src.as_slice().par_chunks_exact(src.cols() * C))
```
edgarriba, line 65: cool! i wasn't aware of this. I think that could be another good task to improve across the operators

crates/kornia-imgproc/src/flip.rs
```diff
@@ -27,43 +33,60 @@ use rayon::{iter::ParallelIterator, slice::ParallelSliceMut};
...
+        .for_each(|(dst_row, src_row)| {
+            let n = src.cols() - 1;
+            for i in 0..=n {
```
edgarriba, line 68: the previous two pointer approach i believe it should be fast as you will cut by half the num of iterations here

edgarriba: and one more improvement is to make the swap of the pixel [T; C] at once using `copy_from_slice` or similar in order to remove the inner 0..C loop

crates/kornia-imgproc/src/flip.rs
```diff
@@ -80,31 +103,47 @@ where
...
-                let idx_i = i * C + c + k * src.cols() * C;
-                let idx_j = i * C + c + j * src.cols() * C;
-                dst.as_slice_mut().swap(idx_i, idx_j);
+    dst.as_slice_mut()
+        .par_chunks_exact_mut(src.cols() * C)
+        .zip_eq(src.as_slice().par_chunks_exact(src.cols() * C).rev())
+        .for_each(|(dst_row, src_row)| {
+            for i in 0..src.cols() {
+                for c in 0..C {
+                    dst_row[i * C + c] = src_row[i * C + c];
```
edgarriba, line 141: similar a above

edgarriba (changes requested):

crates/kornia-imgproc/src/flip.rs
emilmgeorge: > the previous two pointer approach i believe it should be fast as you will cut by half the num of iterations here

The reason I had to change the loop (full iteration and adding src) was because I removed `let mut dst = src.clone();` so that we can take in the dst as a parameter.  The other alternative is to keep the loop as as, but initially copy from src to dst. Is this what you prefer?

> and one more improvement is to make the swap of the pixel [T; C] at once using `copy_from_slice` or similar in order to remove the inner 0..C loop

This is what I could come up with:

    dst.as_slice_mut()
        .par_chunks_exact_mut(src.cols() * C)
        .zip_eq(src.as_slice().par_chunks_exact(src.cols() * C))
        .for_each(|(dst_row, src_row)| {
            dst_row
                .chunks_exact_mut(C)
                .zip(src_row.chunks_exact(C).rev())
                .for_each(|(dst_pixel, src_pixel)| {
                    dst_pixel.copy_from_slice(src_pixel);
                })
        });
If it is the above, then the change for `vertical_flip` is just to put `.rev()` in the first zip instead of the second.

--
Am I missing your point or going in the wrong direction? Please let me know which you prefer regarding both points.

edgarriba: This might be valid too. An alternative could be what you had but use `n//2` instead, then
```rust
let (idx_i, idx_j) = ( i * C + c, (n-1) * C + c);
dst_row[idx_i] = src_row[idx_i];
dst_row[idx_j] = src_row[idx_j];
```
and somehow avoid the C inner loop using the built in slice copy.

Last, to make sure what’s the best, extend the benchmarks and compare the implementations. Would be an interesting exercise to understand how rust behaves in this case iterators vs index access. Maybe add a `bench_flip.rs` in https://github.com/kornia/kornia-rs/tree/main/crates/kornia-imgproc/benches

emilmgeorge: Thanks, I hadn't thought of using the half loop like that.
I'll try benchmarking.

emilmgeorge:
I [benchmarked](https://github.com/kornia/kornia-rs/pull/117/commits/f5eda2d177264a1282cc9f6996a34277d2047708) horizontal_flip with the following code variations:

1. `par_par_slicecopy` => Row: parallel Iter -- Column: Parallel Iter -- Channel: SliceCopy
2. `par_loop_loop` => Row: parallel Iter -- Column: HalfLoop -- Channel: Loop
3. `par_loop_slicecopy` => Row: parallel Iter -- Column: HalfLoop -- Channel: SliceCopy
4. `par_seq_slicecopy` => Row: parallel Iter -- Column: Normal (Sequential) Iter -- Channel: SliceCopy
5. `kornia` => (The first code in this PR with the full loop) Row: parallel Iter -- Column: FullLoop -- Channel: Loop

The general trend is that (4) usually comes out on top, followed closely by (3), and then (2), (5) and (1) at the last. In case of larger images (1024x896), the difference between the (4), (3) and (2) is usually small and inconsistent between runs.

Criterion.rs Line chart (open in new tab if in dark mode):
image

The average time value from the last 1024x896 run are:
(1)  1.3298 ms
(2)  1.2114 ms
(3)  1.2118 ms
(4)  1.2062 ms
(5)  1.2484 ms

Here's the full report generated by criterion:
[criterion.zip](https://github.com/user-attachments/files/16947913/criterion.zip)

Please let me know your thoughts on this or if you have any other ideas to compare.

edgarriba:
let's keep `par_seq_slicecopy`as it uses iterators and speed perf seems reasonable ?

emilmgeorge:
Done.

edgarriba (approved):

edgarriba:
Merging this, actually would be great as a post pr, to verify against known libraries like image-rs

emilmgeorge:
I tried including image_rs in the benchmark (similar to how it is in bench_resize.rs), but my PC (16GB RAM) does not have enough memory for the full test. It gets terminated by the linux OOM killer before the 1024x896 test. The resize benchmark also crashes midway on my PC.

For 256x224 and 512x448, image-rs performs much worse (eg. 1.0548 **ms** for `image_rs` vs 161.53 **us** for `par_seq_slicecopy`).

For ref, here's the patch (Let me know if you want me to include it in this PR):
[0001-Add-image_rs-to-the-flip-benchmark.patch.txt](https://github.com/user-attachments/files/16965815/0001-Add-image_rs-to-the-flip-benchmark.patch.txt)

PR #125, opened 2024-09-05
implement Tensor into_vec

edgarriba:
- implement `Tensor` / `TensorStorage`  -- `into_vec()` to accomplish zero-copies between other libraries e.g `ort` @decahedron1

### DIFF
--- a/crates/kornia-core/src/storage.rs
+++ b/crates/kornia-core/src/storage.rs
@@ -83,14 +83,14 @@ where
     pub fn from_vec(vec: Vec<T>, alloc: A) -> Self {
         // NOTE: this is a temporary solution until we have a custom allocator for the buffer
         // create immutable buffer from vec
-        //let buffer = unsafe {
-        //    // SAFETY: `vec` is properly aligned and has the correct length.
-        //    Buffer::from_custom_allocation(
-        //        NonNull::new_unchecked(vec.as_ptr() as *mut u8),
-        //        vec.len() * std::mem::size_of::<T>(),
-        //        Arc::new(vec),
-        //    )
-        //};
+        // let _buffer = unsafe {
+        //     // SAFETY: `vec` is properly aligned and has the correct length.
+        //     Buffer::from_custom_allocation(
+        //         NonNull::new_unchecked(vec.as_ptr() as *mut u8),
+        //         vec.len() * std::mem::size_of::<T>(),
+        //         Arc::new(vec),
+        //     )
+        // };
 
         // create immutable buffer from vec
         // NOTE: this is a temporary solution until we have a custom allocator for the buffer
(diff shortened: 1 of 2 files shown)


crates/kornia-core/src/storage.rs
```diff
@@ -313,4 +343,21 @@ mod tests {
...
+        let original_vec_capacity = original_vec.capacity();
+
+        let storage = TensorStorage::<i32, _>::from_vec(original_vec, allocator);
+
+        // Convert the storage back to a vector
+        let result_vec = storage.into_vec();
+
+        // check NO copy
+        assert_eq!(result_vec.capacity(), original_vec_capacity);
+        assert!(std::ptr::eq(result_vec.as_ptr(), original_vec_ptr));
```
edgarriba, line 361: @decahedron1 this test seems to fail, which means that we enter in the Err(buf) case

edgarriba: maybe there's something wrong with the `TensorStorage::from_vec` which underneath uses [`Buffer::from_custom_allocation`](https://docs.rs/arrow/latest/arrow/buffer/struct.Buffer.html#method.from_custom_allocation)

edgarriba: UPDATE: i think i fixed ... but I'm missing something around how to tie with the provided allocator. My idea here was that the user can provider custom allocators. eg. CudaAllocator to hold CudaImages to interop easily with other frameworks like ort, vpi, etc  to not move memory forth and back

decahedron1: Hmm, what was causing the test failure? The fix in https://github.com/kornia/kornia-rs/pull/125/commits/1642803ab4c8113b772b0ef8c839cfe3d02ae622 doesn't seem obvious to me, but again, I know nothing about `arrow_buffer` 🤔

Integration with CUDA buffers would have to go through a different API, [`TensorRefMut::from_raw`](https://docs.rs/ort/2.0.0-rc.5/ort/type.TensorRefMut.html#method.from_raw). It just needs a pointer and shape, so Kornia's existing APIs suffice. `TensorRefMut`s can be used zero-copy in session inputs just like a `Value`, but since the memory is user-managed, it does have a lifetime bound. See ort's `cudarc` example: https://github.com/pykeio/ort/blob/60f6ecae9064a75a474678998af42ea87067b1b1/examples/cudarc/src/main.rs#L35-L43

edgarriba: > Hmm, what was causing the test failure?
I might need to investigate a bit more the `Buffer::from_custom_allocation`, maybe is doing copy because later the test checking ptrs is failing.

> Integration with CUDA buffers would have to go through a different API, 
I'll play with it , thanks for the pointers.

PR #129, opened 2024-09-08
implement kornia-dnn with RTDETR detector

edgarriba:


crates/kornia-dnn/src/rtdetr.rs
```diff
@@ -0,0 +1,174 @@
...
+}
+
+/// RT-DETR object detector.
+///
+/// This struct represents an instance of the RT-DETR object detection model.
+pub struct RTDETRDetector {
+    session: Session,
+}
+
+impl RTDETRDetector {
```
edgarriba, line 77: @decahedron1 we got RTDETR working and seems pretty smooth. Woulk like to hear you opinion about this api or any considerations we should have for future models

crates/kornia-dnn/Cargo.toml
```diff
@@ -0,0 +1,22 @@
...
+readme.workspace = true
+repository.workspace = true
+rust-version.workspace = true
+version.workspace = true
+
+[dependencies]
+kornia-core = { workspace = true }
+kornia-image = { workspace = true }
+ort = { version = "2.0.0-rc.4", features = [
+    "load-dynamic",
```
decahedron1, line 20: I'm generally against libraries setting features that affect linking like `load-dynamic` because Cargo features are additive workspace-wide; if an application's dependency sets `load-dynamic` for ort, there's no way for the developer to unset it. If a user of `kornia-dnn` wants to ship with CPU-only ort they'd be forced to compile their own dynamic library (since only CUDA doesn't ship with static libraries).

While I'd imagine most users *would* be using CUDA and thus probably the `load-dynamic` feature, I think it would be better to split this out into a passthrough feature like
```toml
[features]
ort-load-dynamic = [ "ort/load-dynamic" ]
ort-cuda = [ "ort/cuda" ]
```

and document accordingly, with a link to https://ort.pyke.io/setup/linking for more info. This way, end users are granted flexibility, but it isn't too much of a hassle for the majority of users looking to use the most common configuration.

crates/kornia-dnn/src/rtdetr.rs
```diff
@@ -0,0 +1,174 @@
...
+/// Builder for the RT-DETR detector.
+///
+/// This struct provides a convenient way to configure and create an `RTDETRDetector` instance.
+pub struct RTDETRDetectorBuilder {
+    /// Path to the RT-DETR model file.
+    pub model_path: PathBuf,
+    /// Path to the ONNX Runtime dynamic library.
+    pub ort_dylib_path: PathBuf,
+    /// Number of threads to use for inference.
+    pub num_threads: usize,
```
decahedron1, line 24: Execution providers can be registered through the environment so this is super minor but WDYT about an `execution_providers: Vec<ExecutionProviderDispatch>` field to configure EPs specifically for the RTDETR session?

crates/kornia-dnn/src/rtdetr.rs
```diff
@@ -0,0 +1,174 @@
...
+use ort::{GraphOptimizationLevel, Session};
+
+/// Builder for the RT-DETR detector.
+///
+/// This struct provides a convenient way to configure and create an `RTDETRDetector` instance.
+pub struct RTDETRDetectorBuilder {
+    /// Path to the RT-DETR model file.
+    pub model_path: PathBuf,
+    /// Path to the ONNX Runtime dynamic library.
+    pub ort_dylib_path: PathBuf,
```
decahedron1, line 22: Should be removed along with default `load-dynamic`, instead add docs to instruct the user to use [`ort::init_from`](https://docs.rs/ort/2.0.0-rc.4/ort/fn.init_from.html)

decahedron1 (changes requested):
Outside of the `load-dynamic` issue, everything looks great! 🔥

crates/kornia-dnn/src/rtdetr.rs
edgarriba: i wanted to somehow make ort transparent to the user and avoid them to explicitly pass the `ort::ExecutionProvider` and have something custom 

```rust
enum ExecutionProvided {
    Cpu,
    Cuda,
    TensorRT,
}

let execution_providers = match execution_provider {
    ExecutionProvider::Cpu => vec![CpuExecutionProvider::default()],
    ExecutionProvider::Cuda => vec![CUDAExecutionProvider::default()],
    ExecutionProvider::TensorRT => vec![TensorRTExecutionProvider::default()],
};

let session = Session::builder()?
     .with_optimization_level(GraphOptimizationLevel::Level3)?
     .with_intra_threads(num_threads)?
     .with_execution_providers([execution_providers])?
     .commit_from_file(model_path)?;
```

I'm still experimenting with the execution providers. What is the point of defining multiple as a Vec, just because of a fallback provider ?

I've played a bit with it and i noticed that cuda/tensorrt takes few seconds to run the first frames.

A couple of questions:
- Is there any way to leave this to the constructor of the session before i fetch the session ? (might suffer some queuying issues with cameras)
- For tensorrt, means that's is compiling the model at runtime ? could we somehow passe a compiled model ?
- As for the `commit_from_file` -- the idea is that we will have a bunch of operators/models in our [Kornia HF hub ](https://huggingface.co/kornia/ONNX_models) which can use the `commit_from_url` but somehow also let the user also to give a local onnx file. Any tips here ?

crates/kornia-dnn/src/rtdetr.rs
edgarriba: you mean force the user to `export ORT_DYLIB_PATH=/path/to/onnxruntime/build/Linux/RelWithDebInfo/libonnxruntime.so` -- as I'm seeing that ` ort::init_from` takes the dyn lib path

crates/kornia-dnn/src/rtdetr.rs
```diff
@@ -0,0 +1,174 @@
...
+            ],
+            image_chw.into_vec(),
+            CpuAllocator,
+        )?;
+
+        // make the ort tensor
+        let ort_tensor = ort::Tensor::from_array((image_nchw.shape, image_nchw.into_vec()))?;
+
+        // run the model
+        let outputs = self.session.run(ort::inputs!["input" => ort_tensor]?)?;
```
edgarriba, line 141: @decahedron1 how could we pre-allocate the output tensor here ?

crates/kornia-dnn/src/rtdetr.rs
decahedron1: What about re-exporting EPs like `pub use ort::{CPUExecutionProvider, CUDAExecutionProvider, TensorRTExecutionProvider}` so users can still configure each EP's options instead of being limited to defaults? That should still keep things neat.

> What is the point of defining multiple as a Vec, just because of a fallback provider ?

Yes.

> Is there any way to leave this to the constructor of the session before i fetch the session ? (might suffer some queuying issues with cameras)

You could run it on 1 dummy frame inside the constructor, that should get the graph warmed up.

> For tensorrt, means that's is compiling the model at runtime ? could we somehow passe a compiled model ?

Yes, and for CUDA it is determining the most optimal cuDNN convolution kernels. By default it performs an exhaustive search which gets the best performance at the cost of significant warmup time - this can be configured with [`CUDAExecutionProvider::with_conv_algorithm_search`](https://docs.rs/ort/2.0.0-rc.4/ort/struct.CUDAExecutionProvider.html#method.with_conv_algorithm_search).

TensorRT graphs can *theoretically* be cached with [`TensorRTExecutionProvider::with_engine_cache`](https://docs.rs/ort/2.0.0-rc.4/ort/struct.TensorRTExecutionProvider.html#method.with_engine_cache) but some users in the pyke Discord have reported that ONNX Runtime sometimes doesn't respect this option, and session creation can still take a few seconds despite using a cached engine. Your mileage may vary, though; I personally haven't been able to reproduce the issue, but it's just something to keep in mind.

> As for the `commit_from_file` -- the idea is that we will have a bunch of operators/models in our [Kornia HF hub ](https://huggingface.co/kornia/ONNX_models)which can use the `commit_from_url` but somehow also let the user also to give a local onnx file. Any tips here ?

How about something like this? (very roughly)

```rs
pub trait ModelSource {
    fn commit_session(&self, builder: SessionBuilder) -> ort::Result<Session>;
}

pub trait SessionBuilderExt {
    fn commit_from_source<S: ModelSource>(self, source: S) -> ort::Result<Session>;
}

impl SessionBuilderExt for SessionBuilder {
    fn commit_from_source<S: ModelSource>(self, source: S) -> ort::Result<Session> {
        source.commit_session(self)
    }
}

impl<P: AsRef<Path>> ModelSource for P {
    fn commit_session(&self, builder: SessionBuilder) -> ort::Result<Session> {
        builder.commit_from_file(self.as_ref())
    }
}

pub mod models {
    pub struct ExampleRTDETR;

    impl super::ModelSource for ExampleRTDETR {
        fn commit_session(&self, builder: SessionBuilder) -> ort::Result<Session> {
            builder.commit_from_url("https://kornia.rs/model/rtdetr.onnx")
        }
    }
}

// rtdetr.rs
pub struct RTDETRDetectorBuilder {
    pub source: Box<dyn ModelSource>
}

impl RTDETRDetectorBuilder {
    pub fn with_source<S: ModelSource>(mut self, source: S) -> Self {
        self.source = Box::new(source);
        self
    }
}

impl RTDETRDetector {
    pub fn new(...) -> Self {
        let session = Session::builder()?
            ...
            .commit_from_source(source)?;
        Ok(Self { session })
    }
}

// user code
let rtdetr = RTDETRDetectorBuilder::with_source(kornia_dnn::models::ExampleRTDETR).build()?;
// or from file
let rtdetr = RTDETRDetectorBuilder::with_source("./my-local-rtdetr.onnx").build()?;
```

crates/kornia-dnn/src/rtdetr.rs
decahedron1: Either works - `ort::init_from` functions the same as setting `ORT_DYLIB_PATH` and then running `ort::init`, but I have ran into a few instances where it, for whatever reason (likely my fault 🙃), ignored the value set from `std::env::set_var`. If the user wants to set the path programmatically I'd advise them to use `ort::init_from` rather than `std::env::set_var`.

crates/kornia-dnn/src/rtdetr.rs
decahedron1: This seems like the perfect use case for [`OutputSelector`](https://docs.rs/ort/2.0.0-rc.4/ort/struct.OutputSelector.html)!

crates/kornia-dnn/src/rtdetr.rs
edgarriba: > i wanted to somehow make ort transparent to the user and avoid them to explicitly pass the ort::ExecutionProvider and have something custom

actually one more idea would be to kinda automatically set based on feature flags?

crates/kornia-dnn/Cargo.toml
edgarriba: done

crates/kornia-dnn/src/rtdetr.rs
edgarriba: done

Issue #136, opened 2024-09-18
Add cosine similarity metric


edgarriba:
Follow the implementation below

![image](https://github.com/user-attachments/assets/c1998843-0430-465b-a583-ba91f5c36216)

*https://www.linkedin.com/posts/ashvardanian_100x-10000x-numerical-error-reduction-activity-7242002088803663872-kVsx?utm_source=share&utm_medium=member_desktop

PR #138, opened 2024-09-18
Add cosine dist

jandremarais:
First attempt to add cosine distance metric as mentioned in #136

### DIFF
--- a/crates/kornia-core/src/tensor.rs
+++ b/crates/kornia-core/src/tensor.rs
@@ -866,10 +866,46 @@ where
     }
 }
 
+/// Compute the cosine similarity between `input1` and `input2` along dimension `dim`.
+///
+/// todo: add latex
+///
+/// # Arguments
+///
+/// * `tensor1` - todo.
+/// * `tensor2` - todo.
+/// * `dim` - todo.
+///
+/// # Returns
+///
+/// A new `Tensor` instance of the cosine similarities.
+///
+/// # Example
+///
+/// ```
+/// todo
+/// ```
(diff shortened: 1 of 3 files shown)


edgarriba:
@jandremarais thanks for the implementation. I'm considering wether this should be implemented more in terms of Tensor instead of Image. Which brings the question about how we set the axis direction where to compute the metric. What do you think ?

jandremarais:
I see. That makes more sense. Then it probably needs a "dim" parameter in the function too. Should it support broadcasting?

I found these implementation notes in the pytorch source interesting and leaving it here for reference:

https://github.com/pytorch/pytorch/blob/main/aten/src/ATen/native/Distance.cpp#L274-L303

edgarriba:
i see, maybe we need even first a proper `norm` operator. Which I've been recently thinking about having a separated `kornia-core-ops` crate to disentagle the tensor memory and layout management to the rest of all the operators (even the existing ones). That was actually one of multiple reasons why decided to drop ndarray support to have a clean separated Tensor struct for the mere purpose of moving data round.

jandremarais:
Closing reasons:
- Cosine sim should operate on tensors not images. 
- Cosine sim needs vector norm operation
- Unsure where tensor operations should live

jandremarais:
Closing reasons:

- Cosine sim should operate on tensors not images.
- Cosine sim needs vector norm operation
- Unsure where tensor operations should live

edgarriba:
> Unsure where tensor operations should live

feel free to open a PR creating a new subcrate `kornia-core-ops` in the workspace

PR #142, opened 2024-09-20
create new crate for core-ops

jandremarais:
Separate the core tensor operations from the core crate to a core-ops crate.

### DIFF
--- a/crates/kornia-core-ops/src/error.rs
+++ b/crates/kornia-core-ops/src/error.rs
@@ -1,9 +1,14 @@
+use kornia_core::TensorError;
 use thiserror::Error;
 
 /// An error type for tensor operations.
 #[derive(Error, Debug, PartialEq)]
 pub enum TensorOpsError {
-    /// Dimension out of bounds.
+    /// The dimension to perform the operation over is greater than the number of dimsions of the tensor.
     #[error("Dimension out of bounds. The dimension {0} is out of bounds ({1}).")]
     DimOutOfBounds(usize, usize),
+
+    /// Tensor error
+    #[error("Error with the tensor: {0}")]
+    TensorError(#[from] TensorError),
 }
--- a/crates/kornia-core-ops/src/error.rs
+++ b/crates/kornia-core-ops/src/error.rs
@@ -1,0 +1,9 @@
+use thiserror::Error;
+
+/// An error type for tensor operations.
+#[derive(Error, Debug, PartialEq)]
+pub enum TensorOpsError {
+    /// Dimension out of bounds.
+    #[error("Dimension out of bounds. The dimension {0} is out of bounds ({1}).")]
+    DimOutOfBounds(usize, usize),
+}
--- a/crates/kornia-core-ops/src/ops.rs
+++ b/crates/kornia-core-ops/src/ops.rs
@@ -79,14 +79,22 @@ mod tests {
     fn test_sum_2d() -> Result<(), TensorError> {
         let data: [u8; 6] = [1, 1, 1, 1, 1, 1];
         let t = Tensor::<u8, 2>::from_shape_slice([2, 3], &data, CpuAllocator)?;
+        let t_f32 = t.cast::<f32>();
+        let t_i32 = t.cast::<i32>();
 
         let agg = sum_elements(&t, 1).unwrap();
-        assert_eq!(agg.as_slice(), [3, 3]);
         assert_eq!(agg.shape, [2, 1]);
+        assert_eq!(agg.as_slice(), [3, 3]);
+
+        assert_eq!(sum_elements(&t_f32, 1).unwrap().as_slice(), [3., 3.]);
+        assert_eq!(sum_elements(&t_i32, 1).unwrap().as_slice(), [3, 3]);
 
         let agg = sum_elements(&t, 0).unwrap();
-        assert_eq!(agg.as_slice(), [2, 2, 2]);
         assert_eq!(agg.shape, [1, 3]);
+        assert_eq!(agg.as_slice(), [2, 2, 2]);
+
+        assert_eq!(sum_elements(&t_f32, 0).unwrap().as_slice(), [2., 2., 2.]);
+        assert_eq!(sum_elements(&t_i32, 0).unwrap().as_slice(), [2, 2, 2]);
         Ok(())
     }
 
@@ -94,18 +102,26 @@ mod tests {
     fn test_sum_3d() -> Result<(), TensorError> {
         let data: [u8; 24] = [1; 24];
         let t = Tensor::<u8, 3>::from_shape_slice([2, 3, 4], &data, CpuAllocator)?;
+        let t_f32 = t.cast::<f32>();
+        let t_i32 = t.cast::<i32>();
 
         let agg = sum_elements(&t, 0).unwrap();
-        assert_eq!(agg.as_slice(), [2; 12]);
         assert_eq!(agg.shape, [1, 3, 4]);
+        assert_eq!(agg.as_slice(), [2; 12]);
+        assert_eq!(sum_elements(&t_f32, 0).unwrap().as_slice(), [2.; 12]);
+        assert_eq!(sum_elements(&t_i32, 0).unwrap().as_slice(), [2; 12]);
 
         let agg = sum_elements(&t, 1).unwrap();
-        assert_eq!(agg.as_slice(), [3; 8]);
         assert_eq!(agg.shape, [2, 1, 4]);
+        assert_eq!(agg.as_slice(), [3; 8]);
+        assert_eq!(sum_elements(&t_f32, 1).unwrap().as_slice(), [3.; 8]);
+        assert_eq!(sum_elements(&t_i32, 1).unwrap().as_slice(), [3; 8]);
 
         let agg = sum_elements(&t, 2).unwrap();
-        assert_eq!(agg.as_slice(), [4; 6]);
         assert_eq!(agg.shape, [2, 3, 1]);
+        assert_eq!(agg.as_slice(), [4; 6]);
+        assert_eq!(sum_elements(&t_f32, 2).unwrap().as_slice(), [4.; 6]);
+        assert_eq!(sum_elements(&t_i32, 2).unwrap().as_slice(), [4; 6]);
 
         Ok(())
     }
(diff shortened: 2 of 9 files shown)


jandremarais:
@edgarriba do you want the operations currently defined in `kornia-core/src/tensor.rs` like `add`/`mul`/`powi`/etc to live in `kornia-core-ops/src/ops.rs`? 
If so, how should we keep the ability to call for example `tensor.powi(2)` since we can't define the impl for a foreign type? Do we need a wrapper struct?

edgarriba:
possibly wiht a TensorOps trait but let's explore this later. Put a first an initial crate with something simple and we keep iterating

jandremarais:
Ok. Should I start with something simple that is not yet implemented or something that is already implement for Tensor?

edgarriba:
maybe the norm operator needed by cosine distance

jandremarais:
I started with the norm operator but then realised it needed the `sum` reduce operator. So I did its implementation first.

crates/kornia-core-ops/src/ops.rs
```diff
@@ -0,0 +1,111 @@
+use kornia_core::{storage::TensorStorage, SafeTensorType, Tensor, TensorAllocator};
+
+use crate::error::TensorOpsError;
+
+/// Compute the sum of the elements in the tensor along dimension `dim`
+///
+pub fn sum<T, const N: usize, A>(
```
edgarriba, line 7: Maybe call it `sum_elements` to be more specific

jandremarais:
Marked this as ready for review so that the new core-ops crate can be merged without being blocked by all the ops that need to be implemented. I you would rather want this to be a bigger PR with more ops added first, I can change the status to draft and work on it further.

crates/kornia-core-ops/src/ops.rs
```diff
@@ -0,0 +1,111 @@
...
+        let data: [u8; 4] = [1, 1, 1, 1];
+        let t = Tensor::<u8, 1>::from_shape_slice([4], &data, CpuAllocator)?;
+        let res = sum_elements(&t, 0);
+
+        assert!(res.is_ok_and(|v| v.as_slice() == [4]));
+        Ok(())
+    }
+
+    #[test]
+    fn test_sum_2d() -> Result<(), TensorError> {
```
edgarriba, line 78: for sanity check can we test with a different types e.g. here `f32`?

crates/kornia-core/src/tensor.rs
```diff
@@ -381,6 +381,26 @@ where
...
+    /// Get the index of the element at the given offset. The reverse of `Self::get_iter_offset`.
+    ///
+    /// # Arguments
+    ///
+    /// * `offset` - The offset of the element at the given index.
+    ///
+    /// # Returns
+    ///
+    /// The list of indices to get the element from.
+    pub fn get_index(&self, offset: usize) -> [usize; N] {
```
edgarriba, line 393: if guess this should return an error if `offset` is larger than the number of elements

crates/kornia-core-ops/src/error.rs
```diff
@@ -0,0 +1,9 @@
+use thiserror::Error;
+
+/// An error type for tensor operations.
+#[derive(Error, Debug, PartialEq)]
+pub enum TensorOpsError {
+    /// Dimension out of bounds.
+    #[error("Dimension out of bounds. The dimension {0} is out of bounds ({1}).")]
+    DimOutOfBounds(usize, usize),
```
edgarriba, line 8: can we reuse or unify somehow with https://github.com/kornia/kornia-rs/blob/main/crates/kornia-core/src/tensor.rs#L23 ?

crates/kornia-core-ops/src/ops.rs
```diff
@@ -0,0 +1,111 @@
+use kornia_core::{storage::TensorStorage, SafeTensorType, Tensor, TensorAllocator};
+
+use crate::error::TensorOpsError;
+
+/// Compute the sum of the elements in the tensor along dimension `dim`
+///
```
edgarriba, line 6: add args in the docs and a small snippet example

crates/kornia-core-ops/src/ops.rs
```diff
@@ -0,0 +1,111 @@
...
+pub fn sum_elements<T, const N: usize, A>(
```
edgarriba, line 7: for future PR -- worth finding an abstraction like here https://github.com/kornia/kornia-rs/blob/main/crates/kornia-core/src/tensor.rs#L710

one more thought is how to support with the same interface tensors with contiguous memory that do not require computing the offset for the given index.

edgarriba: and something I coudn't figure out yet is how to add a `keepdim` e.g to reduce and return the tensor rank to N -1. @emilmgeorge maybe you have some thoughts here

crates/kornia-core/src/tensor.rs
jandremarais: Yes, I was wondering if this should follow the same pattern as the `get_iter_offset` method with a checked and unchecked version where the checked version returns an `Option`. What do you think?

crates/kornia-core-ops/src/error.rs
jandremarais: I can add `DimOutOfBounds` to `TensorError`. Feels like it means something different to `IndexOutOfBounds`. The latter is when the index is greater than the size of the array and the former when the queried dimension/axis does not exist.

jandremarais: I couldn't think of a scenario where this error is applicable outside of tensor operations (tensor indexing is already validated by the const generics) and therefore I created it in the core-ops crate initially.

crates/kornia-core-ops/src/ops.rs
emilmgeorge: @edgarriba 
I think we can simply remove the `dim` dimension from `shape` and `strides` when `keepdim == false` since the num elements stay the same. I hope I'm not misunderstanding what you meant.

edgarriba: this is actually the only limitation we have with generic tensor ranks. Hopefully we con figure out a workaround as this comes useful to convert tensor to onnxruntime or any other dl engine :)

see: https://github.com/kornia/kornia-rs/blob/8cd3394ab7aad40f0d1ef503366eca7a98756b94/crates/kornia-dnn/src/rtdetr.rs#L158

edgarriba: where BTW, we should be able to bindo directly to cuda at some point

crates/kornia-core/src/tensor.rs
edgarriba: good idea, add test to check forth and back

crates/kornia-core-ops/src/error.rs
edgarriba: i see, yes makes sense. Make sure the description and error message is clear for both then

crates/kornia-core-ops/src/ops.rs
emilmgeorge: Would it be feasible to make dimensions dynamic? Like `Tensor<T, A>` instead of `Tensor<T, N, A>` and then manage shape/stride at runtime using Vec. I don't know the whole code base well enough yet to know the possibility and consequences of this. Just sharing a thought. If feasible, it would solve this problem and give more flexibility. But I guess this would also cause some performance hit.

jandremarais: I tried to experiment with this idea of keeping const generic ranks and using it to inform the sum operation on what rank tensor to return. I've added these traits heavily borrowed (simplified and probably worsen) from [dfdx](https://github.com/coreylowman/dfdx/blob/main/dfdx-core/src/shapes/shape.rs) with this [commit](https://github.com/kornia/kornia-rs/pull/142/commits/4b36377d7e122d3a6a9beb79bb7431ef6e1ce79b). Example usage can be seen in the test from this [commit](https://github.com/kornia/kornia-rs/pull/142/commits/d4fa9512851aa36bb230e47d619e7c60faadeb3d).

I might have messed up this PR now. Let me know what's best to do next.

edgarriba: @jandremarais that's a nice feature ! Maybe as a separated PR we can introduce the reshape stuff, and later integrate into the ops?

crates/kornia-core-ops/src/ops.rs
```diff
@@ -1,7 +1,76 @@
...
+    where
+        Self::Shape: ReduceShapeTo<Dst>;
+}
+
+impl<T: SafeTensorType + std::ops::Add<Output = T>, A: TensorAllocator>
+    ReduceShapeTo<Tensor<T, 2, A>> for Tensor<T, 3, A>
+{
+    fn reduce(self, dim: usize) -> Tensor<T, 2, A> {
+        let shape = match dim {
+            0 => [self.shape[1], self.shape[2]],
```
edgarriba, line 30: what if dim 0 is greater than 1 ? return error as it's not possible to perform the reduction ?

edgarriba:
@jandremarais any progress here ? we can land step by step and keep improving

jandremarais:
I will work on this tonight (UTC +2). I will remove the dynamic shape stuff, add docs to sum_elements and put the DimOutOfBounds error in TensorError.

edgarriba:
@jandremarais sound good -- consider this too https://github.com/apache/arrow-rs/issues/6492#issuecomment-2387214140. I might be working on a PR for it during the week. Arrow seems pretty unflexible, so maybe we should go for our full implementation

jandremarais:
@edgarriba I addressed all of your comments in the latest commits. Open to further feedback if I interpreted anything incorrectly or if there are any new suggestions.

crates/kornia-core-ops/src/ops.rs
```diff
@@ -0,0 +1,152 @@
...
+        let t = Tensor::<u8, 1>::from_shape_slice([4], &data, CpuAllocator)?;
+        let res = sum_elements(&t, 0);
+
+        assert!(res.is_ok_and(|v| v.as_slice() == [4]));
+        Ok(())
+    }
+
+    #[test]
+    fn test_sum_2d() -> Result<(), TensorOpsError> {
+        let data: [u8; 6] = [1, 1, 1, 1, 1, 1];
```
edgarriba, line 104: if one of this test can with non-equals numbers for sanity check, like
```suggestion
        let data: [u8; 6] = [1, 2, 3, 4, 5, 6];
```

edgarriba (approved):
LGTM. Make a rebase and all good !

Issue #134, opened 2024-09-16
Memory leak in `remap`


liamkinne:
There is some amount of memory being allocated during the `remap` function that isn't getting cleaned up.

This simple program will slowly grow in memory usage:

```rust
fn main() {
    let size = ImageSize {
        width: 1440,
        height: 1440,
    };

    let correction_map = correction_map(&size);

    loop {
        let input = Image::<u8, 3>::from_size_val(size, 0).unwrap();
        let mut output = Image::<f32, 3>::from_size_val(size, 0.0).unwrap();

        remap(
            &input.cast().unwrap(),
            &mut output,
            &correction_map.0,
            &correction_map.1,
            InterpolationMode::Bilinear,
        )
        .unwrap();
    }
}
```

edgarriba:
That’s weird, however the idea with the new function signature allowing mutable outputs is that you allocate once outside of the for loop. If you check in the benches is how I’m doing.

If you’re up to, would be great to provide a bench for remap using also `[flamegraph](https://github.com/flamegraph-rs/flamegraph)` or similar to understand better the memory usage pattern.

BTW, remap is not the fastest implementation for now, it needs some optimisations as eg. the interpolation function it´s computing everytime the offset indices of the tensor from where to interpolate which could be done as a pre-step by iterating row by row with step 1. A second one, would be using simd to compute the weighted interpolation.

emilmgeorge:
I couldn't reproduce the growing memory usage issue while running for about 5 minutes. But, when testing with `valgrind`, an unrelated leak is found in the `correction_map` I used (included below). (No confirmed leaks within the original snippet). 
<details>
  <summary>This is the correction_map that I used.</summary>

```rust
use kornia::image::{Image, ImageSize};
use kornia::core::Tensor2;
use kornia::imgproc::interpolation::{remap, InterpolationMode, grid::meshgrid_from_fn};

fn correction_map(size: &ImageSize) -> (Tensor2<f32>, Tensor2<f32>) {
    // Identity mapping
    let (map_x, map_y) = meshgrid_from_fn(size.width, size.height, |x, y| {
        Ok((x as f32, y as f32))
    }).unwrap();
    (map_x, map_y)
}
```

</details>

<details>
  <summary>Valgrind log</summary>

```valgrind
==419002== Memcheck, a memory error detector
==419002== Copyright (C) 2002-2024, and GNU GPL'd, by Julian Seward et al.
[...]

PR #145, opened 2024-09-23
Fix TensorStorage memory deallocation

emilmgeorge:
**Changes in this PR**
- Adds checks to make sure the TensorAllocator is invoked properly.
- Fixes memory leaks/issues observed in https://github.com/kornia/kornia-rs/issues/134#issuecomment-2363726170 and https://github.com/kornia/kornia-rs/pull/117#issuecomment-2343983888.

**Details**
A new struct `TensorCustomAllocationOwner` is added to represent a custom memory allocations made by TensorAllocators. This struct implements the `Drop` trait to deallocate the associated memory. An instance of this struct is then passed as owner to `arrow_buffer::Buffer::from_custom_allocation`, so that when the Arrow buffer is dropped, this memory will also be deallocated using the correct TensorAllocator.

**Future changes**
When the `allocator_api` feature comes to stable rust, it may be possible to use a standard container such as `Vec<T, A=..>` instead of the above mentioned struct to handle custom deallocation.

### DIFF
--- a/crates/kornia-core/src/storage.rs
+++ b/crates/kornia-core/src/storage.rs
@@ -18,6 +18,30 @@ impl SafeTensorType for i64 {}
 impl SafeTensorType for f32 {}
 impl SafeTensorType for f64 {}
 
+/// Represents the owner of custom Arrow Buffer memory allocations.
+///
+/// This struct is used to facilitate the automatic deallocation of the memory it owns,
+/// using the `Drop` trait.
+pub struct TensorCustomAllocationOwner<A: TensorAllocator> {
+    /// The allocator used to allocate the tensor storage.
+    alloc: A,
+    /// The layout used for the allocation.
+    layout: Layout,
+    /// The pointer to the allocated memory
+    ptr: *const u8,
+}
+
+// SAFETY: TensorCustomAllocationOwner is never modifed from multiple threads.
+impl<A: TensorAllocator> std::panic::RefUnwindSafe for TensorCustomAllocationOwner<A> {}
+unsafe impl<A: TensorAllocator> Sync for TensorCustomAllocationOwner<A> {}
+unsafe impl<A: TensorAllocator> Send for TensorCustomAllocationOwner<A> {}
+
+impl<A: TensorAllocator> Drop for TensorCustomAllocationOwner<A> {
+    fn drop(&mut self) {
+        self.alloc.dealloc(self.ptr as *mut u8, self.layout);
+    }
+}
+
 /// Represents a contiguous memory region that can be shared with other buffers and across thread boundaries.
 ///
 /// This struct provides methods to create, access, and manage tensor storage using a custom allocator.
@@ -35,9 +59,10 @@ where
     alloc: A,
 }
 
-impl<T, A: TensorAllocator> TensorStorage<T, A>
+impl<T, A> TensorStorage<T, A>
 where
     T: SafeTensorType + Clone,
+    A: TensorAllocator + 'static,
 {
     /// Creates a new tensor storage with the given length and allocator.
     ///
@@ -51,8 +76,13 @@ where
     /// A new tensor storage if successful, otherwise an error.
     pub fn new(len: usize, alloc: A) -> Result<Self, TensorAllocatorError> {
         // allocate memory for tensor storage
-        let ptr =
-            alloc.alloc(Layout::array::<T>(len).map_err(TensorAllocatorError::LayoutError)?)?;
+        let layout = Layout::array::<T>(len).map_err(TensorAllocatorError::LayoutError)?;
+        let ptr = alloc.alloc(layout)?;
+        let owner = TensorCustomAllocationOwner {
+            alloc: alloc.clone(),
+            layout,
+            ptr,
+        };
 
         // create the buffer
         let buffer = unsafe {
@@ -60,7 +90,7 @@ where
             Buffer::from_custom_allocation(
                 NonNull::new_unchecked(ptr),
                 len * std::mem::size_of::<T>(),
-                Arc::new(Vec::<T>::with_capacity(len)),
+                Arc::new(owner),
             )
         };
 
@@ -223,7 +253,7 @@ where
         let buffer = Buffer::from_custom_allocation(
             NonNull::new_unchecked(ptr as *mut u8),
             len * std::mem::size_of::<T>(),
-            Arc::new(Vec::<T>::with_capacity(len)),
+            Arc::new(()),
         );
 
         // create tensor storage
@@ -238,7 +268,7 @@ where
 impl<T, A> Clone for TensorStorage<T, A>
 where
     T: SafeTensorType + Clone,
-    A: TensorAllocator + Clone,
+    A: TensorAllocator + Clone + 'static,
 {
     fn clone(&self) -> Self {
         let mut new_storage = Self::new(self.len(), self.alloc.clone())
--- a/crates/kornia-core/src/storage.rs
+++ b/crates/kornia-core/src/storage.rs
@@ -253,6 +253,8 @@ mod tests {
     use super::*;
     use crate::allocator::CpuAllocator;
     use std::alloc::Layout;
+    use std::cell::RefCell;
+    use std::rc::Rc;
 
     #[test]
     fn test_tensor_storage() -> Result<(), TensorAllocatorError> {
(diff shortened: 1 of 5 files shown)


crates/kornia-core/src/storage.rs
```diff
@@ -365,4 +397,76 @@ mod tests {
...
+        // TensorStorage::new()
+        // Deallocation should happen when `storage` goes out of scope.
+        {
+            let _storage = TensorStorage::<u8, _>::new(len, allocator.clone()).unwrap();
+            assert_eq!(*allocator.bytes_allocated.borrow(), len as i32);
+        }
+        assert_eq!(*allocator.bytes_allocated.borrow(), 0);
+
+        // TensorStorage::new() -> TensorStorage::into_vec()
+        // TensorStorage::into_vec() consumes the storage and creates a copy (in this case).
```
edgarriba, line 434: is it possible to achieve the same behaviour without the copy?

crates/kornia-core/src/storage.rs
```diff
@@ -365,4 +397,76 @@ mod tests {
...
+            assert_eq!(*allocator.bytes_allocated.borrow(), 0);
+        }
+        assert_eq!(*allocator.bytes_allocated.borrow(), 0);
+
+        // TensorStorage::from_vec()  -> TensorStorage::into_vec()
+        // TensorStorage::from_vec() currently does not use the custom allocator, so the
+        // bytes_allocated value should not change.
+        {
+            let vec = Vec::<u8>::with_capacity(len);
+            let storage = TensorStorage::<u8, _>::from_vec(vec, allocator.clone());
```
edgarriba, line 450: i think we need to resolve well here when we create a storage from a Vec how's the allocator involved

crates/kornia-core/src/storage.rs
```diff
@@ -365,4 +397,76 @@ mod tests {
...
+
+        // TensorStorage::from_ptr()
+        // TensorStorage::from_ptr() does not take ownership of buffer. So the memory should not be
+        // deallocated when the TensorStorage goes out of scope.
+        // In this case, the memory will be deallocated when the vector goes out of scope.
+        {
+            let mut vec = Vec::<u8>::with_capacity(len);
+            {
+                let _storage =
+                    unsafe { TensorStorage::<u8, _>::from_ptr(vec.as_mut_ptr(), len, &allocator) };
```
edgarriba, line 466: can we add just for sanity in all this tests that the ptr from the original data is the same as in the storage

edgarriba (commented):
LGTM. Added some suggestions to make testing more robust

crates/kornia-core/src/storage.rs
edgarriba: worth to check this https://github.com/apache/arrow-rs/issues/6362

emilmgeorge: Yes that would be good. This is my related understanding:
- Std Vector and `Buffer::from_vec` do not support custom allocators until `allocator_api` comes to stable rust. So, for our future custom allocators like `CudaAllocator`, `from_vec` will have to involve copying and use `Buffer::from_custom_allocator`.
- For `CpuAllocator`, we can have zero-copy using `Buffer::from_vec` (as it is currently). 
But to be fully safe, I think we should also change `CpuAllocator` to use `std::alloc::{alloc,dealloc}` (Global allocator) instead of `std::alloc::System.{alloc,dealloc}`. This is because vector uses the `Global` allocator. This is usually the same as `std::alloc::System` but the user can change it using the [global_allocator](https://doc.rust-lang.org/std/alloc/index.html) attribute. By changing `CpuAllocator` to use `std::alloc::{alloc,dealloc}`, it always matches the allocator used by the vector (even when user changes it).

I'm not sure how to switch the implementation of `TensorStorage::from_vec` to one of the above based on whether `A` is `CpuAllocator` or `CudaAllocator` though. (maybe different functions? Ideas welcome!)

I haven't done any of the above in this PR though. Please let me know your thoughts and I can change accordingly.

crates/kornia-core/src/storage.rs
```diff
@@ -18,6 +18,30 @@ impl SafeTensorType for i64 {}
...
+///
+/// This struct is used to facilitate the automatic deallocation of the memory it owns,
+/// using the `Drop` trait.
+pub struct TensorCustomAllocationOwner<A: TensorAllocator> {
+    /// The allocator used to allocate the tensor storage.
+    alloc: A,
+    /// The layout used for the allocation.
+    layout: Layout,
+    /// The pointer to the allocated memory
+    ptr: *const u8,
```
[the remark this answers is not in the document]
emilmgeorge: Yea, that makes more sense. Will change, thanks!

crates/kornia-core/src/storage.rs
```diff
@@ -51,16 +76,21 @@ where
     /// A new tensor storage if successful, otherwise an error.
     pub fn new(len: usize, alloc: A) -> Result<Self, TensorAllocatorError> {
         // allocate memory for tensor storage
-        let ptr =
-            alloc.alloc(Layout::array::<T>(len).map_err(TensorAllocatorError::LayoutError)?)?;
+        let layout = Layout::array::<T>(len).map_err(TensorAllocatorError::LayoutError)?;
+        let ptr = alloc.alloc(layout)?;
+        let owner = TensorCustomAllocationOwner {
+            alloc: alloc.clone(),
```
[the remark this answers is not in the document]
emilmgeorge: I cloned because I needed alloc below also, to store in TensorStorage struct. (I know it's not ideal to have two copies of it but I couldn't easily find a way around it.). Do you mean wrap A in Arc everywhere or just here?

crates/kornia-core/src/storage.rs
emilmgeorge: My comment in the below thread may be relevant.

crates/kornia-core/src/storage.rs
emilmgeorge: Will add.

crates/kornia-core/src/storage.rs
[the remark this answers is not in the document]
edgarriba: yeap, wrap `alloc: Arc<A>` in both `TensorStorage` and `TensorCustomAllocationOwner` maybe ?

crates/kornia-core/src/storage.rs
edgarriba: > But to be fully safe, I think we should also change CpuAllocator to use std::alloc::{alloc,dealloc} (Global allocator) instead of std::alloc::System.{alloc,dealloc}

please, do :+1: 

maybe the behaviour for cuda should be that when a cuda storage is created via vec, the data is consumed, cuda allocated and copied to device, and deallocate the original cpu vector ? Haven't faced yet the full use case. Probably we should use the `kornia::dnn` module to try this workflows and prototype from there. Found a similar c++ implementation maybe to have a reference: https://gist.github.com/CommitThis/1666517de32893e5dc4c441269f1029a

edgarriba: one more request in this direction, is the ability to easily create Image views. As Image is tuple struct out of Tensor: https://github.com/kornia/kornia-rs/blob/main/crates/kornia-image/src/image.rs#L59

In some workflows i have different types of images out of Tensor which i need to convert to Image::new (`as_slice().to_vec() everytime which involves copies) in order to use any kornia function. Unless we decide e.g to adapt the whole api to accept Tensor, and Image in the end it's just a trait in order to give some semantics and define specific types of images with formats e.g Rgb8U, Mono8U.

emilmgeorge: I didn't quite understand the image view part. But for converting Tensor3 to Image without copy, we could implement the `TryFrom` trait or a function `from_tensor` that uses the passed tensor if channel dimension matches.

edgarriba: Image in the end is a Tensor3 — would be a bit overkill to have a from_tensor method. I think ideally for this case we might want to have a method that somehow transfers the ownership of the storage, shape and strides ?

crates/kornia-core/src/storage.rs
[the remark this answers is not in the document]
emilmgeorge: Done

crates/kornia-core/src/storage.rs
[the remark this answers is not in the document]
emilmgeorge: Done

crates/kornia-core/src/storage.rs
```diff
@@ -365,4 +397,76 @@ mod tests {
...
+
+        let allocator = TestAllocator {
+            bytes_allocated: Rc::new(RefCell::new(0)),
+        };
+        let len = 1024;
+
+        // TensorStorage::new()
+        // Deallocation should happen when `storage` goes out of scope.
+        {
+            let _storage = TensorStorage::<u8, _>::new(len, allocator.clone()).unwrap();
```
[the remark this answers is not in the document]
emilmgeorge: Done

crates/kornia-core/src/storage.rs
emilmgeorge: Done

crates/kornia-core/src/storage.rs
emilmgeorge: Just to make myself clear, this is what I had in mind:
```rust
impl<T, const C: usize> TryFrom<Tensor<T, 3, CpuAllocator>> for Image<T, C>
where
    T: SafeTensorType,
{
    type Error = ImageError;

    fn try_from(value: Tensor<T, 3, CpuAllocator>) -> Result<Self, Self::Error> {
        if value.shape[2] == C {
            Ok(Self(value))
        } else {
            Err(ImageError::InvalidTensorShape)
        }
    }
}

```
Used like:
```rust
let image: Image<_, 3> = tensor.try_into().unwrap(); 
// OR
let image = Image::<_, 3>::try_from(tensor).unwrap();

```

edgarriba: Oh, I see, sounds good! I’ll try myself in a separated PR. I have also some potential improvements for the image struct, like adding a third ImageColorSpace in order to define more specific types like
`type ImageRgb8 = Image<u8, 3, ColorSpace::Rgb>` which color will have associated values too

crates/kornia-core/src/storage.rs
```diff
@@ -365,4 +394,88 @@ mod tests {
...
+        let allocator = TestAllocator {
+            bytes_allocated: Rc::new(RefCell::new(0)),
+        };
+        let len = 1024;
+
+        // TensorStorage::new()
+        // Deallocation should happen when `storage` goes out of scope.
+        {
+            let _storage = TensorStorage::<u8, _>::new(len, allocator.clone())?;
+            assert_eq!(*allocator.bytes_allocated.borrow(), len as i32);
```
edgarriba, line 426: No pointer test here ?

edgarriba (approved):
LGTM

crates/kornia-core/src/storage.rs
emilmgeorge: In this case the storage is not created from another buffer so there is no original pointer to compare to. Are there any other checks I should have added?

crates/kornia-core/src/storage.rs
emilmgeorge: Regarding this PR, I just want to note that it only fixes the deallocation issues noted in the first comment. It does not currently include the changes related to `into_vec` for non-CpuAllocators as discussed above in this review thread. I had started it, but it is not ready yet. I can send a separate PR when it's ready (hope that's ok).

PR #146, opened 2024-09-24
Tensor display

jandremarais:
implement Display for Tensor

### DIFF
--- a/crates/kornia-core/src/tensor.rs
+++ b/crates/kornia-core/src/tensor.rs
@@ -1,3 +1,5 @@
+use std::cmp::Ordering;
+
 use num_traits::Float;
 use thiserror::Error;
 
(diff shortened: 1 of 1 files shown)


crates/kornia-core/src/tensor.rs
```diff
@@ -1304,4 +1380,103 @@ mod tests {
...
+        let lines = disp.lines().collect::<Vec<_>>();
+
+        #[rustfmt::skip]
+        assert_eq!(lines.as_slice(),
+        ["[[[ 1, 2],",
+         "  [ 3, 4],",
+         "  [ 5, 6]],",
+         "",
+         " [[ 7, 8],",
+         "  [ 9,10],",
```
edgarriba, line 1412: maybe with a space too here ?

crates/kornia-core/src/tensor.rs
```diff
@@ -1304,4 +1380,103 @@ mod tests {
...
+    #[test]
+    fn display_big_tensor() -> Result<(), TensorError> {
+        let data: [u8; 1000] = [0; 1000];
+        let t = Tensor::<u8, 3>::from_shape_slice([10, 10, 10], &data, CpuAllocator)?;
+        let disp = t.to_string();
+        let lines = disp.lines().collect::<Vec<_>>();
+
+        #[rustfmt::skip]
+        assert_eq!(lines.as_slice(),
+        ["[[[0,0,0,...,0],",
```
edgarriba, line 1455: can be done later, but playing around with numpy/torch i've noticed that the magic number to truncate output is 10

image
image

edgarriba (approved):
LGTM. Improvements can be done later

PR #151, opened 2024-09-28
improve image tensor conversions

edgarriba:
- use builtin `Buffer::from_slice_ref` in TensorStorage creation
- add method `from_size_slice` in `Image`
- implement `TryFrom` for Image/Tensor conversion
- added tests

### DIFF
--- a/crates/kornia-core/src/storage.rs
+++ b/crates/kornia-core/src/storage.rs
@@ -160,6 +160,54 @@ where
         }
     }
 
+    /// Creates a new `TensorStorage` from a slice of data.
+    ///
+    /// # Arguments
+    ///
+    /// * `data` - A slice containing the data to be stored.
+    /// * `alloc` - The allocator to use for creating the storage.
+    ///
+    /// # Returns
+    ///
+    /// A new `TensorStorage` instance containing a copy of the input data.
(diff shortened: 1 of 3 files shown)


crates/kornia-core/src/storage.rs
```diff
@@ -160,6 +160,54 @@ where
...
+    ///
+    /// # Returns
+    ///
+    /// A new `TensorStorage` instance containing a copy of the input data.
+    ///
+    /// # Errors
+    ///
+    /// Returns a `TensorAllocatorError` if the allocation fails.
+    pub fn from_slice(data: &[T], alloc: A) -> Self {
+        let buffer = Buffer::from_slice_ref(data);
```
edgarriba, line 178: @emilmgeorge discovered this function, i thinks it still makes copy but we can explore later more safer ways to achieve zero copies. Another alternative could be to explore `from_bytes` https://docs.rs/arrow-buffer/latest/arrow_buffer/buffer/struct.Buffer.html#method.from_bytes

emilmgeorge: 👍🏼  (also: need to check how to handle custom allocations for non-CPU allocators?).

PR #155, opened 2024-09-29
implement a crop function

edgarriba:
- implement an image crop function in `kornia_improc::crop` 
- [x] TODO: benchmark against image-rs

image


edgarriba:
@emilmgeorge @jandremarais

PR #156, opened 2024-10-02
Implement Tensor Buffer

edgarriba:
- remove arrow-rs Buffer and go for custom
- simplify api for now

PR #157, opened 2024-10-07
[wip] Add cosine similarity

jandremarais:
#136 

[x] add norm functions
[ ] add cosine similarity function

### DIFF
--- a/crates/kornia-core-ops/src/ops.rs
+++ b/crates/kornia-core-ops/src/ops.rs
@@ -49,7 +49,7 @@ where
     ///
     /// ```
     /// use kornia_core::{Tensor, CpuAllocator};
-    /// use kornia_core_ops::ops::TensorOps;
+    /// use kornia_core_ops::TensorOps;
     ///
     /// let data: [u8; 6] = [1, 1, 1, 1, 1, 1];
     /// let t = Tensor::<u8, 2>::from_shape_slice([2, 3], &data, CpuAllocator).unwrap();
@@ -80,7 +80,7 @@ where
     ///
     /// ```
     /// use kornia_core::{Tensor, CpuAllocator};
-    /// use kornia_core_ops::ops::TensorOps;
+    /// use kornia_core_ops::TensorOps;
     ///
     /// let data: [f32; 5] = [3., 3., 2., 1., 1.];
     /// let t = Tensor::<f32, 1>::from_shape_slice([5], &data, CpuAllocator).unwrap();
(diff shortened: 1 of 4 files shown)


jandremarais:
@edgarriba I want to give the cosine similarity another shot. I started with the norm functions and thought I'll ask for an interim review before I go further.

crates/kornia-core/src/tensor.rs
```diff
@@ -653,6 +653,36 @@ where
...
+    /// Apply the power function to the tensor data.
+    ///
+    /// # Arguments
+    ///
+    /// * `n` - The power to raise the tensor data to.
+    ///
+    /// # Returns
+    ///
+    /// A new tensor with the data raised to the power.
+    pub fn powf(&self, n: T) -> Tensor<T, N, A>
```
edgarriba, line 665: the point of having the kornia-core-ops was to start reducing operators in the tensor definition. Does it make sense to have `TensorOps` trait that implements all this `map` based operators ?

jandremarais: I understand. I was unsure how would you like to go about moving the old operators over and the powf function was so close to the existing operators, that I thought I would rather put it with them. I can move them over to a TensorOps trait if we are ready for that.

edgarriba: yes, can do step by step and iterate on the api

jandremarais:
This is what it looks like wrapped in a TensorOps trait. Open to feedback before moving the older ops over.

crates/kornia-core-ops/src/ops.rs
```diff
@@ -1,78 +1,203 @@
...
+    ///
+    /// # Errors
+    ///
+    /// If the requested dimension is greater than the number of axes of the tensor, an error is returned.
+    ///
+    /// # Example
+    ///
+    /// ```
+    /// use kornia_core::{Tensor, CpuAllocator};
+    /// use kornia_core_ops::ops::TensorOps;
```
edgarriba, line 52:
```suggestion
    /// use kornia_core_ops::TensorOps;
```
Maybe we can expose to root ?

crates/kornia-core-ops/src/ops.rs
```diff
@@ -1,78 +1,203 @@
...
+        })
     }
 
-    let storage = TensorStorage::from_vec(data, tensor.storage.alloc().clone());
+    fn pnorm(&self, p: T, dim: usize) -> Result<Tensor<T, N, A>, TensorOpsError>
+    where
+        T: std::ops::Add<Output = T> + Float,
+    {
+        let p_inv = T::one() / p;
+        Ok(self.powf(p).sum_elements(dim)?.powf(p_inv))
```
edgarriba, line 192: Should we provide a reduction argument?

edgarriba (commented):
Looking good !

crates/kornia-core-ops/src/ops.rs
jandremarais: Do you mean to drop the dimension that the operation was run over?

edgarriba:
@jandremarais consdier some changes from https://github.com/kornia/kornia-rs/pull/156

edgarriba:
@jandremarais any updates here ?

edgarriba:
@jandremarais are you planning to continue working on this ?

PR #239, opened 2025-02-24
[wip] Cubecl API and grayscale to rgb u8 kernel

edgarriba:


omar-abdelgawad:
Hey! I have a question since the [free github actions can't support a GPU](https://betatim.github.io/posts/github-action-with-gpu/) how would we be able to test the GPU backend for future implementations?

edgarriba:
for now assume that we have to check with our own computers with a gpu

crates/kornia-imgproc/src/cubecl.rs
```diff
@@ -0,0 +1,139 @@
...
+
+    if x < cols && y < rows {
+        let idx = y * cols + x;
+        let r = u16::cast_from(src[3 * idx]);
+        let g = u16::cast_from(src[3 * idx + 1]);
+        let b = u16::cast_from(src[3 * idx + 2]);
+        let gray = u8::cast_from(((r * 77 + g * 150 + b * 29) + 128) >> 8);
+        dst[idx] = Line::new(gray);
+    }
+}
```
omar-abdelgawad, line 94: I ran the same test successfully with the following implementation instead:
```rust
#[cube(launch_unchecked)]
fn gray_from_rgb_u8_cl_kernel(
    src: &Array<Line<u8>>,
    dst: &mut Array<Line<u8>>,
) {
    let idx = ABSOLUTE_POS;
    if idx < src.len() {
        let r = u16::cast_from(src[3 * idx]);
        let g = u16::cast_from(src[3 * idx + 1]);
        let b = u16::cast_from(src[3 * idx + 2]);
        let gray = u8::cast_from(((r * 77 + g * 150 + b * 29) + 128) >> 8);
        dst[idx] = Line::new(gray);
    }
}
```

omar-abdelgawad: Should I make a PR to this branch or to the main branch? I am also revising the ImageCL abstraction layer for #251 from burn [CubeTensor](https://github.com/tracel-ai/burn/blob/960a635a55a15e8aacf84f67bb38f099506ac8a7/crates/burn-cubecl/src/tensor/base.rs#L15).

edgarriba: hi @omar-abdelgawad thanks for trying! Have you found a huge difference ? I tried at some point and was minimal. Happy to adopt your version in any case. I think for now I'll refactor this work into a separated experimental kornia-cubecl crate to isolate deps until it matures. This will happen between today and tomorrow

omar-abdelgawad: > Have you found a huge difference ?

No The difference is a few microseconds in the mean time of 256x224 benchmark only since all the measurements are quite fast anyway. But I was just trying to write the kernel to be as concise as possible and also figure out what are the optimal cube count, cube dim, and vectorization to use. btw, I tried to write about my results in the [burn Grayscale thread](https://discord.com/channels/1038839012602941528/1341456695414624257) that you opened before. Is your current benchmark still slow?

edgarriba: with the recent changes of having the imagecl struct holding the runtime data everything got more clear, as the cubecl guys mentioned the bottleneck was in the host/device copy

omar-abdelgawad: > I think for now I'll refactor this work into a separated experimental kornia-cubecl crate to isolate deps until it matures. This will happen between today and tomorrow

Maybe you can tell me the layout of what you want to be done and I will handle it if possible. I submitted my proposal for anyway so I would be more than happy to start making a PR. Also should I skip the gpu tests for CI passing?

edgarriba: hi, sorry for late reply here. Not sure if you want to keep contributing on this but possibly the best to move forward is to have it for now as separated experimental `kornia-cubecl` crate. As we are still figuring out the final api for the current Tensor/Image to be agnostic to the memory allocator and backend, we might want the cubecl Image as decoupled implementation

omar-abdelgawad: Sounds good. As for my contribution, I honestly don't have much free time lately but I don't mind contributing to this branch and keeping it updated with the main branch. for now, can I make PRs to this branch (cubecl-grayscale)?

Issue #220, opened 2025-01-09
Implement native SVD for 3x3 matrices


edgarriba:
Port the following code to be used in kornia::linalg
https://github.com/wi-re/tbtSVD/blob/master/source/SVD.h

PR #241, opened 2025-03-04
Implement SVD for Mat3x3

prikshitsingh24:
fixes: #220 
This is a port of : https://github.com/wi-re/tbtSVD/blob/master/source/SVD.h
It calculates V matrix correctly but the remaining S and U matrix are still not accurate.
Correct answer:
chrome_d60ehatq9n
Algorithm answer:
`
SVDSet { U: Mat3 { x_axis: Vec3(1.7881393e-7, 0.0, 0.9999999), y_axis: Vec3(0.0, -1.0, 0.0), z_axis: Vec3(0.9999999, 0.0, -1.7881393e-7) }, S: Mat3 { x_axis: Vec3(2.9999995, -0.0, 1.7881393e-7), y_axis: Vec3(0.0, 2.0, -0.0), z_axis: Vec3(-5.364418e-7, 0.0, 0.9999999) }, V: Mat3 { x_axis: Vec3(1.0, 0.0, 0.0), y_axis: Vec3(0.0, 1.0, -0.0), z_axis: Vec3(-0.0, 0.0, 1.0) } }
`

### DIFF
--- a/crates/kornia-linalg/src/linalg.rs
+++ b/crates/kornia-linalg/src/linalg.rs
@@ -5,7 +5,7 @@ const GAMMA: f32 = 5.828427124;
 const CSTAR: f32 = 0.923879532;
 const SSTAR: f32 = 0.3826834323;
 const SVD_EPSILON: f32 = 1e-6;
-const JACOBI_STEPS: u8 = 12;
+const JACOBI_STEPS: u8 = 6;
 const RSQRT_STEPS: u8 = 4;
 const rsqrt_mut_STEPS: u8 = 6;
 
@@ -153,9 +153,10 @@ fn approximate_givens_quaternion(A: &Symmetric3x3) -> Givens {
         ch: 2.0 * (A.m_00 - A.m_11),
         sh: A.m_10,
     };
-
-    let mut b = GAMMA * g.sh * g.sh < g.ch * g.ch;
-    let w = rsqrt(g.ch * g.ch + g.sh * g.sh);
+    let ch2 = g.ch * g.ch;
+    let sh2 = g.sh * g.sh;
+    let mut b = GAMMA * sh2 < ch2;
+    let w = rsqrt(ch2 + sh2);
 
     if w != w {
         // Checking for NaN
@@ -217,8 +218,10 @@ fn jacobi_conjugation(x: usize, y: usize, z: usize, S: &mut Symmetric3x3, q: &mu
     // Compute the Givens rotation (approximated)
     let mut g = approximate_givens_quaternion(S);
     // Scale and calculate intermediate values
-    let scale = 1.0 / (g.ch * g.ch + g.sh * g.sh);
-    let a = (g.ch * g.ch - g.sh * g.sh) * scale;
+    let ch2 = g.ch * g.ch;
+    let sh2 = g.sh * g.sh;
+    let scale = 1.0 / (ch2 + sh2);
+    let a = (ch2 - sh2) * scale;
     let b = 2.0 * g.sh * g.ch * scale;
 
     // Create a copy of the matrix to avoid modifying the original during calculations
--- a/crates/kornia-linalg/src/linalg.rs
+++ b/crates/kornia-linalg/src/linalg.rs
@@ -1,22 +1,23 @@
-use std::ops::{IndexMut, Index};
+use std::ops::{Index, IndexMut};
 
-const GEMMA:f32 = 5.828427124; 
-const CSTAR:f32 = 0.923879532;
-const SSTAR:f32 = 0.3826834323;
+use glam::{Mat3, Quat, Vec3};
+
+const GEMMA: f32 = 5.828427124;
+const CSTAR: f32 = 0.923879532;
+const SSTAR: f32 = 0.3826834323;
 const SVD_EPSILON: f32 = 1e-6;
 const JACOBI_STEPS: u32 = 12;
 const RSQRT_STEPS: u32 = 4;
 const RSQRT1_STEPS: u32 = 6;
 
-
 /// Calculates the result of x / y. Required as the accurate square root function otherwise uses a reciprocal approximation when using optimizations on a GPU which can lead to slightly different results. If non exact matching results are acceptable a simple division can be used.
-pub fn fdiv(x:f32, y:f32) -> f32 {
+pub fn fdiv(x: f32, y: f32) -> f32 {
     return x / y;
 }
 
 /// Calculates the reciprocal square root of x using a fast approximation.
 pub fn rsqrt(x: f32) -> f32 {
-    let mut xhalf = -0.5 * x;
+    let xhalf = -0.5 * x;
     let mut i = x.to_bits() as i32; // Convert float to raw bits
     i = 0x5f375a82 - (i >> 1); // Magic constant and bit manipulation
     let mut x = f32::from_bits(i as u32); // Convert bits back to float
@@ -43,7 +44,7 @@ pub fn rsqrt1(mut x: f32) -> f32 {
 }
 
 /// Calculates the square root of x using 1.f/rsqrt1(x) to give a square root with controllable and consistent precision.
-pub fn accurate_sqrt(mut x:f32) -> f32 {
+pub fn accurate_sqrt(x: f32) -> f32 {
     return fdiv(1.0, rsqrt(x));
 }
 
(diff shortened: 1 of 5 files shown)


crates/kornia-linalg/src/linalg.rs
```diff
@@ -0,0 +1,509 @@
...
+
+    Q.z_axis.x = 2.0 * g2.ch * g2.sh;
+    Q.z_axis.y = -2.0 * g3.ch * sh22 * g3.sh;
+    Q.z_axis.z = sh22 * sh32;
+
+    QR { Q, R }
+}
+
+/// Wrapping function used to contain all of the required sub calls
+pub fn svd(A: Mat3) -> SVDSet {
```
edgarriba, line 451: maybe call it `svd3` ?

crates/kornia-linalg/src/linalg.rs
```diff
@@ -0,0 +1,509 @@
...
+
+    let w = rsqrt(g.ch * g.ch + g.sh * g.sh);
+    g.ch *= w;
+    g.sh *= w;
+
+    return g;
+}
+
+/// Implements a QR decomposition of a Matrix
+pub fn qr_decomposition(B: &mut Mat3) -> QR {
```
edgarriba, line 377:
```suggestion
fn qr_decomposition(B: &mut Mat3) -> QR {
```
make all the "utils" as private functions

edgarriba (commented):
I'll review the math soon -- but good start !

crates/kornia-linalg/src/linalg.rs
```diff
@@ -0,0 +1,509 @@
...
+}
+
+#[cfg(test)]
+mod tests {
+    use glam::{Mat3, Vec3};
+
+    use super::*;
+
+    #[test]
+    fn test_svd() {
```
edgarriba, line 481: can you add a numerical test ? to check that we do not have any regression ? You could als try to reconstuct the input `A' = U @ (S @ VT)`

crates/kornia-linalg/src/linalg.rs
```diff
@@ -0,0 +1,488 @@
...
+    let z = -(*x);
+    if c {
+        *x = *y;
+        *y = z;
+    }
+}
+
+#[derive(Debug, Clone, Copy)]
+/// A simple symmetric 3x3 Matrix class (contains no storage for (0, 1) (0, 2) and (1, 2)
+pub struct Symmetric3x3 {
```
edgarriba, line 71:
```suggestion
struct Symmetric3x3 {
```

crates/kornia-linalg/src/linalg.rs
```diff
@@ -0,0 +1,488 @@
...
+            m_20: mat.z_axis.x,
+            m_21: mat.z_axis.y,
+            m_22: mat.z_axis.z,
+        }
+    }
+}
+
+#[derive(Debug, Clone, Copy)]
+/// Helper struct to store 2 floats to avoid OUT parameters on functions
+pub struct Givens {
```
edgarriba, line 119:
```suggestion
struct Givens {
```

crates/kornia-linalg/src/linalg.rs
```diff
@@ -0,0 +1,488 @@
...
+// Helper function to swap X and Y and swap Y with -X if c is true
+fn cond_neg_swap(c: bool, x: &mut f32, y: &mut f32) {
+    let z = -(*x);
+    if c {
+        *x = *y;
+        *y = z;
+    }
+}
+
+#[derive(Debug, Clone, Copy)]
```
edgarriba, line 69: Double check that clone and copy traits are really needed. Same for the others

crates/kornia-linalg/src/linalg.rs
```diff
@@ -0,0 +1,488 @@
...
+            3 => &mut self.0.w,
+            _ => panic!("Index out of bounds for Quaternion"),
+        }
+    }
+}
+
+/// Function used to apply a givens rotation S. Calculates the weights and updates the quaternion to contain the cumultative rotation
+fn jacobi_conjugation(x: usize, y: usize, z: usize, S: &mut Symmetric3x3, q: &mut IndexedQuat) {
+    // Compute the Givens rotation (approximated)
+    let mut g = approximate_givens_quaternion(S);
```
edgarriba, line 240:
```suggestion
    let mut g = approximate_givens_quaternion(S);

```

edgarriba (changes requested):

crates/kornia-linalg/src/linalg.rs
```diff
@@ -0,0 +1,488 @@
...
+            x_axis: Vec3::new(1.0, 0.0, 0.0),
+            y_axis: Vec3::new(0.0, 2.0, 0.0),
+            z_axis: Vec3::new(0.0, 0.0, 3.0),
+        };
+
+        // Perform SVD on matrix A
+        let A_clone = A.clone();
+        let svd_result = svd3(A_clone);
+        // Check matrix V
+        assert!(
```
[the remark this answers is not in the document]
prikshitsingh24: glam::Mat3 type doesn't implement the RelativeEq and AbsDiffEq

crates/kornia-linalg/src/linalg.rs
```diff
@@ -0,0 +1,488 @@
...
+}
+
+#[cfg(test)]
+mod tests {
+    use glam::{Mat3, Vec3};
+
+    use super::*;
+
+    #[test]
+    fn test_svd3() {
```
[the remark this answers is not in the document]
prikshitsingh24: @edgarriba I wrote a bench file and these are the benchmarks:

svd3/svd3/ - time:   [1.3185 µs 1.3202 µs 1.3221 µs]
Found 6 outliers among 100 measurements (6.00%)
  1 (1.00%) low mild
  1 (1.00%) high mild
  4 (4.00%) high severe

svd3/svd3_faer/ - time:   [614.62 ns 618.62 ns 623.58 ns]
Found 19 outliers among 100 measurements (19.00%)
 7 (7.00%) high mild
 12 (12.00%) high severe

crates/kornia-linalg/src/linalg.rs
```diff
@@ -0,0 +1,487 @@
...
+const RSQRT_STEPS: u8 = 4;
+const rsqrt_mut_STEPS: u8 = 6;
+
+/// Standard CPU division.
+fn fdiv(x: f32, y: f32) -> f32 {
+    return x / y;
+}
+
+/// Calculates the reciprocal square root of x using a fast approximation.
+fn rsqrt(x: f32) -> f32 {
```
edgarriba, line 18: We might want to try https://betterexplained.com/articles/understanding-quakes-fast-inverse-square-root/ which doesn’t have the loops. That could be the cause difference between faer ?

crates/kornia-linalg/src/linalg.rs
```diff
@@ -0,0 +1,487 @@
...
+
+    Q.z_axis.x = 2.0 * g2.ch * g2.sh;
+    Q.z_axis.y = -2.0 * g3.ch * sh22 * g3.sh;
+    Q.z_axis.z = sh22 * sh32;
+
+    QR3 { Q, R }
+}
+
+/// Wrapping function used to contain all of the required sub calls
+pub fn svd3(A: Mat3) -> SVD3Set {
```
edgarriba, line 428:
```suggestion
pub fn svd3(A: &Mat3) -> SVD3Set {
```

crates/kornia-linalg/src/linalg.rs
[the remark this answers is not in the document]
edgarriba: The only places where I could see the improvements are in the rsqrt, and in some places with micro optimisation to precompute some variables

crates/kornia-linalg/src/linalg.rs
prikshitsingh24: Ok

prikshitsingh24: Current benchmark after changing the rsqrt:
Code_Og4rurVpvO

crates/kornia-linalg/src/linalg.rs
```diff
@@ -0,0 +1,466 @@
...
+    S.m_20 = _S.m_20;
+    S.m_21 = _S.m_21;
+    S.m_22 = _S.m_22;
+}
+
+/// Function used to contain the givens permutations and the loop of the jacobi steps controlled by JACOBI_STEPS
+/// Returns the quaternion q containing the cumultative result used to reconstruct S
+fn jacobi_eigenanalysis(mut S: Symmetric3x3) -> Mat3 {
+    let mut q = IndexedQuat::new(Quat::from_xyzw(0.0, 0.0, 0.0, 1.0));
+    for _i in 0..JACOBI_STEPS {
```
edgarriba, line 268: Maybe play with the num steps to reduce and find balance between speed and results quality

crates/kornia-linalg/src/linalg.rs
edgarriba: We are getting closer ! Let’s try to find out how to improve a little bit more

edgarriba: Also one suggestion while running benchmarks, try to have as less things as possible running in the background like web browser or slack, etc

prikshitsingh24: Ok got it

crates/kornia-linalg/src/linalg.rs
prikshitsingh24: That worked changing JACOBI_STEPS from 12 to 6 is accurate and more speed than faer:
Code_itWQQcKALL

edgarriba: amazing! can we make sure with tests that the function produce proper results under some corner cases ? test with a couple different input matrices. Good job!

crates/kornia-linalg/src/linalg.rs
```diff
@@ -0,0 +1,561 @@
+// Reference: https://github.com/wi-re/tbtSVD/blob/master/source/SVD.h
+use glam::{Mat3, Quat, Vec3};
```
edgarriba, line 2:
```suggestion
use glam::{Mat3, Quat};
```

crates/kornia-linalg/src/linalg.rs
```diff
@@ -0,0 +1,561 @@
...
+
+    /// The sine of the angle in the Givens rotation.
+    sh: f32,
+}
+
+#[derive(Debug)]
+/// Helper struct to store 2 Matrices to avoid OUT parameters on functions
+struct QR3 {
+    /// The orthogonal matrix Q from the QR decomposition.
+    Q: Mat3,
```
edgarriba, line 124:
```suggestion
    q: Mat3,
```

edgarriba: let's use the lowecase notation. double check all the rest of the code

crates/kornia-linalg/src/linalg.rs
```diff
@@ -0,0 +1,561 @@
...
+
+    /// The upper triangular matrix R from the QR decomposition.
+    R: Mat3,
```
edgarriba, line 127:
```suggestion
    r: Mat3,
```

crates/kornia-linalg/src/linalg.rs
```diff
@@ -0,0 +1,561 @@
...
+}
+
+#[derive(Debug)]
+/// Helper struct to store 3 Matrices to avoid OUT parameters on functions
+pub struct SVD3Set {
+    /// The matrix of left singular vectors.
+    U: Mat3,
```
edgarriba, line 134:
```suggestion
    u: Mat3,
```

crates/kornia-linalg/src/linalg.rs
```diff
@@ -0,0 +1,561 @@
...
+
+    /// The diagonal matrix of singular values.
+    S: Mat3,
```
edgarriba, line 137:
```suggestion
    s: Mat3,
```

crates/kornia-linalg/src/linalg.rs
```diff
@@ -0,0 +1,561 @@
...
+
+    /// The matrix of right singular vectors.
+    V: Mat3,
```
edgarriba, line 140:
```suggestion
    v: Mat3,
```

crates/kornia-linalg/src/linalg.rs
```diff
@@ -0,0 +1,561 @@
...
+
+/// Implementation of Algorithm 4
+fn qr_givens_quaternion(a1: f32, a2: f32) -> Givens {
+    // a1 = pivot point on diagonal
+    // a2 = lower triangular entry we want to annihilate
+    let epsilon = SVD_EPSILON; // Assuming _SVD_EPSILON is defined elsewhere
+    let rho = accurate_sqrt(a1 * a1 + a2 * a2);
+
+    let mut g = Givens {
+        ch: (a1.abs() + (f32::max(rho, epsilon))),
```
edgarriba, line 321:
```suggestion
        ch: a1.abs() + f32::max(rho, epsilon),
```

edgarriba (changes requested):

edgarriba (commented):
@prikshitsingh24 some test do not seems to success

prikshitsingh24:
> @prikshitsingh24 some test do not seems to success

@edgarriba When i ported this in rust it gives this result for first test when I recompute the original matrix, is it how the algo should work like the answer from the algo should match the matrix A or there is some error in the code.
Since only the zero matrix test passes.
Left is the original A and right is the A computed by USV<sup>t</sup>
Code_fQnMxsJXTg

edgarriba:
> > @prikshitsingh24 some test do not seems to success
> 
> @edgarriba When i ported this in rust it gives this result for first test when I recompute the original matrix, is it how the algo should work like the answer from the algo should match the matrix A or there is some error in the code. Since only the zero matrix test passes. Left is the original A and right is the A computed by USVt Code_fQnMxsJXTg

for sanity check, make sure that the improvements we made for example when computing the rsqrt (removing loops) didn't break the algorithm. Make also sure that glam matrices all well constructed and accessed. Maybe worth to add some other test from intermediate function to make sure everything makes sense.

prikshitsingh24:
Hey @edgarriba, I found the issue. I reverted to my old code, and it worked fine. The problem was that the code I referred to used a matrix that stored data in row-major order, while I didn't realize that glam stores matrices in column-major order. Will Try to fix the glam part

edgarriba:
@prikshitsingh24  rebase from main as I fixed the ci issues

edgarriba (approved):

Issue #244, opened 2025-03-09
[`gftt_response`](https://github.com/kornia/kornia/blob/main/kornia/feature/responses.py#L109)


edgarriba:

Puellaquae:
Hi, I plan to work on this issue. I think I'll start by implementing the `gaussian_blur2d` and `spatial_gradient` functions, as they are used in multiple tensor computations.

edgarriba:
@Puellaquae sounds good ! Just notice that `gaussian_blur` is already available in https://github.com/kornia/kornia-rs/blob/e706461e26e390da5d6e0067ee2494cbdf9a8341/crates/kornia-imgproc/src/filter/ops.rs#L35

Puellaquae:
Thanks for pointing that out! I’ll focus on implementing the `spatial_gradient` function first.

AdarshGuptaa:
Hello @edgarriba . I was looking into the OpenCV implementation of [gftt in corner.cpp](https://github.com/opencv/opencv/blob/4.x/modules/imgproc/src/corner.cpp) and I think I could try to implement the core of this in rust.

If this is a good approach could I give it a go?

Issue #243, opened 2025-03-09
implement `harris_response`


edgarriba:
use kornia pytorch implementation as a reference

andrew-shc:
Hi, I'd like to work on this issue. I suppose I can branch off @aayushhzz 's branch.

edgarriba:
i believe this can be started as independent branch

PR #264, opened 2025-03-12
Added additional functions for kornia tensors

aayushhzz:
- `mul_scalar`:- multiplies the tensor by any scalar number
- `powf`:- raises a tensor to a float power
- `min`:- implementation of `torch.minimum` function of pytorch

these functions are used in the implementation of #243, #244

### DIFF
--- a/crates/kornia-tensor-ops/src/ops.rs
+++ b/crates/kornia-tensor-ops/src/ops.rs
@@ -122,13 +122,13 @@ where
 pub fn min<T, const N: usize>(
     tensor: &Tensor<T, N, CpuAllocator>,
     other: &Tensor<T, N, CpuAllocator>,
-) -> Tensor<T, N, CpuAllocator>
+) -> Result<Tensor<T, N, CpuAllocator>, TensorOpsError>
 where
     T: PartialOrd + Clone,
 {
     tensor
         .element_wise_op(other, |a, b| if a < b { a.clone() } else { b.clone() })
-        .expect("Tensor dimension mismatch")
+        .map_err(|_| TensorOpsError::ShapeMismatch(tensor.shape.to_vec(), other.shape.to_vec()))
 }
 
 /// Compute the dot product between two 1D tensors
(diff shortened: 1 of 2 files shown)


edgarriba (changes requested):
ideally this should be part of `tensor-ops` crate

aayushhzz:
> ideally this should be part of `tensor-ops` crate

there are many other functions like these in `tensor` should move them also to `tensor-ops`?

edgarriba:
> > ideally this should be part of `tensor-ops` crate
> 
> there are many other functions like these in `tensor` should move them also to `tensor-ops`?

yep, as much as possible so that the tensor crate keeps light-weighted. Don't do it all at once, try to make it in relatively small PR chunks

crates/kornia-tensor-ops/src/ops.rs
```diff
@@ -77,6 +78,64 @@ where
     })
 }
 
+/// Multiply the pixel data by a scalar.
```
edgarriba, line 81: format is wrong

crates/kornia-tensor-ops/src/ops.rs
```diff
@@ -77,6 +78,64 @@ where
...
+    pub fn powf<T, const N: usize, A>(
+        tensor:&Tensor<T, N, A>, 
+        n: T
+    ) -> Tensor<T, N, A>
+    where
+        T: Float + Clone,
+        A: TensorAllocator + Clone + 'static,
+    {
+        tensor.map(|x| x.powf(n))
+    }
```
edgarriba, line 120:
```suggestion
    }

```

crates/kornia-tensor-ops/src/ops.rs
```diff
@@ -77,6 +78,64 @@ where
...
+    /// # Returns
+    /// 
+    /// A new `Tensor` instance.
+    pub fn min<T, const N: usize, A>(tensor:&Tensor<T, N, CpuAllocator>,other:&Tensor<T, N, CpuAllocator>) -> Tensor<T, N, CpuAllocator>
+    where
+        T: PartialOrd + Clone,
+        A: TensorAllocator + Clone + 'static,
+    {
+        tensor.element_wise_op(other, |a, b| if a < b { a.clone() } else { b.clone() })
+            .expect("Tensor dimension mismatch")
```
edgarriba, line 136: maybe we should return handle the error properly instead of panicking

edgarriba (changes requested):
please, add tests

edgarriba (approved):
LGTM. Thanks so much !

PR #266, opened 2025-03-14
Implemented pyrup

### LINKED ISSUE #249: implement `pyrup`
references:
- https://docs.opencv.org/3.4/d4/d86/group__imgproc__filter.html#gada75b59bdaaca411ed6fee10085eb784
- https://github.com/kornia/kornia/blob/0f25d2ad2a128d010d6e5a5517a14fa901510439/kornia/geometry/transform/pyramid.py#L291

Saga690:
This PR implements pyrup function in crates/kornia-imgproc/src/pyramid.rs and provides a benchmark in benches/bench_pyramid.rs. 

This PR aims to resolve #249.

Please let me know if further changes are to be made!

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -3,10 +3,15 @@ resolver = "2"
 members = [
     "crates/*",
     "examples/*",
+    # NOTE: need to add dora examples to make rust analyzer to work
+    "examples/dora/video-capture",
+    "examples/dora/imgproc",
+    "examples/dora/video-sink",
+    "examples/dora/image-utils",
     "kornia-viz",
     # "kornia-py",
 ]
-exclude = ["kornia-py"]
+exclude = ["kornia-py", "examples/dora"]
 
 [workspace.package]
 authors = ["kornia.org <edgar@kornia.org>"]
--- a/examples/dora/.gitignore
+++ b/examples/dora/.gitignore
@@ -1,0 +1,2 @@
+out
+target
--- a/examples/dora/Cargo.toml
+++ b/examples/dora/Cargo.toml
@@ -1,0 +1,3 @@
+[workspace]
+members = ["video-capture", "video-sink", "imgproc", "image-utils"]
+resolver = "2"
--- a/examples/dora/README.md
+++ b/examples/dora/README.md
@@ -1,0 +1,28 @@
+Example using kornia-rs and dora-rs to build a simple pipeline to process images from multiple sources.
+
+Instructions:
+
+1. Install dependencies:
+
+```bash
+cargo install dora-cli
+```
+
+2. Build the pipeline:
+
+```bash
+dora build dataflow.yml
+```
+
+3. Run the pipeline:
+
+```bash
+dora run dataflow.yml
+```
+
+4. Open the rerun viewer:
+
+```bash
+rerun --bind 127.0.0.1 --port 9876
+```
+![Screenshot from 2025-03-18 16-37-20](https://github.com/user-attachments/assets/e1372851-7ba3-4890-a98e-703855cfe304)
--- a/examples/dora/dataflow.yml
+++ b/examples/dora/dataflow.yml
@@ -1,0 +1,41 @@
+nodes:
+  - id: web-camera
+    build: cargo build -p dora-video-capture
+    path: ./target/debug/dora-video-capture
+    inputs:
+      tick: dora/timer/millis/10
+    outputs:
+      - frame
+    env:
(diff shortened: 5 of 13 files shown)


crates/kornia-imgproc/src/pyramid.rs
```diff
@@ -0,0 +1,120 @@
...
+            expected_width,
+            expected_height,
+            dst.width(),
+            dst.height(),
+        ));
+    }
+
+    let mut upsampled = Image::<f32, C>::from_size_val(dst.size(), 0.0)?;
+    
+    resize_native(src, &mut upsampled, InterpolationMode::Bilinear)?;
```
edgarriba, line 78: would be interesting to benchmark another variant using resize_fast

edgarriba (commented):
can you report a screenshot with the benchmark numbers ?

crates/kornia-imgproc/src/pyramid.rs
Saga690: The benchmark numbers are as seen here:
Screenshot_20250314_160758

Saga690: Screenshot_20250314_160758

Saga690: I will create another variant using resize_fast and get back to you :)

Saga690: I used resize_fast and below are the benchmarking results for 3 channel images (marked by Pyramid Operations/pyrup_fast):
Screenshot_20250314_162419
Screenshot_20250314_162445

edgarriba: I would keep then with fast. The function is simple enough that anyone can replicate using native for large image. Can specify that in the docs ?

Saga690: We can keep both in form of pyrup and pyrup_fast? because resize_fast supports only 3 channel images so the application would be limited.
I'll also make sure to update the API documentation and mention the same.

Saga690: So should i keep only resize_fast finally or both?

edgarriba: let's keep both

crates/kornia-imgproc/src/pyramid.rs
```diff
@@ -0,0 +1,203 @@
...
+        .collect();
+    let kernel_y = vec![1.0, 4.0, 6.0, 4.0, 1.0]
+        .iter()
+        .map(|&x| x / 16.0)
+        .collect();
+
+    (kernel_x, kernel_y)
+}
+
+/// Upsample an image and then blur it using resize_native.
```
edgarriba, line 28:
```suggestion
/// Upsample an image and then blur it.
```

crates/kornia-imgproc/src/pyramid.rs
```diff
@@ -0,0 +1,203 @@
...
+    }
+
+    let mut upsampled = Image::<u8, 3>::from_size_val(dst.size(), 0)?;
+
+    resize_fast(src, &mut upsampled, InterpolationMode::Bilinear)?;
+
+    let mut upsampled_f32 = Image::<f32, 3>::from_size_val(dst.size(), 0.0)?;
+    let mut dst_f32 = Image::<f32, 3>::from_size_val(dst.size(), 0.0)?;
+
+    for (i, &val) in upsampled.as_slice().iter().enumerate() {
```
edgarriba, line 126: you can use `as_slice_mut` and zip with the other tensor

crates/kornia-imgproc/src/pyramid.rs
```diff
@@ -0,0 +1,203 @@
...
+        upsampled_f32.as_slice_mut()[i] = val as f32;
+    }
+
+    let (kernel_x, kernel_y) = get_pyramid_gaussian_kernel();
+    separable_filter(&upsampled_f32, &mut dst_f32, &kernel_x, &kernel_y)?;
```
edgarriba, line 131: can you add a TODO here saying that this should be improved by https://github.com/kornia/kornia-rs/issues/278
this double casting it's just overkilling

edgarriba (commented):
BTW, I just launched the benches in my computer and the results between using resize / resize_fast here are really small. Let's consider only supporting the non-fast one (I believe the reason might the casts needed around it)

PR #267, opened 2025-03-15
implemented `harris_response`

### LINKED ISSUE #243: implement `harris_response`

andrew-shc:
Closes #243

If you check the tests, I borrowed it from the Python example they had in their docs. It seems like there's some differences in Gaussian blurring and kernel normalization, but overall the high points are at the corners and low points at the edges and the center.


crates/kornia-imgproc/src/features.rs
```diff
@@ -1,6 +1,22 @@
 use kornia_image::{Image, ImageError};
 
 use rayon::prelude::*;
+use crate::filter::gaussian_blur;
+
+/// Method to calculate gradient for feature response
+pub enum GradsMode {
+    /// Sobel operators
+    SOBEL,
```
edgarriba, line 9:
```suggestion
    Sobel,
```
Enums go in CamelCase in rust

edgarriba (commented):
Can you add a bench and compare to vip ?https://docs.nvidia.com/vpi/algo_harris_corners.html

crates/kornia-imgproc/src/features.rs
```diff
@@ -68,6 +84,124 @@ pub fn hessian_response(src: &Image<f32, 1>, dst: &mut Image<f32, 1>) -> Result<
     Ok(())
 }
 
+/// Computes the harris response
+pub fn harris_response(src: &Image<f32, 1>,
+                       dst: &mut Image<f32, 1>,
```
edgarriba, line 89: Why optional ?

edgarriba (commented):
Idk what is completely required but worth thinking about a raw u8 implementation depending on the benchmark results

andrew-shc:
Using `gaussian_blur`:
screenshot_20250315-112226
Using manual summation (ignore the change):
screenshot_20250315-124642

It seems like implementing u8 would not change that much. Also, manual summation was able to produce the zeroes along the center and the axes (similiar to the python example). Looking to see if we can force SIMD (maybe its already using it) and other low-level optimization.

andrew-shc:
Tried it with longer warmup (10s) using manual summation:
screenshot_20250316-203341
Using `filter::separable_filter` (supposedly can also be parallelized with rayon):
screenshot_20250316-215122
I tried SIMD autovectorization on a specific part and seems it doesn't want to cooperate (checked via `cargo asm kornia_imgproc::features::harris_response | grep vmov` and it returned nothing). I'm considering a raw u8 implementation would need to be converted to i8 first (to at least support unary negation). Up to you if you think 6ms (with rayon) is good enough compared to their 3.9ms.

edgarriba:
if you think it's feasible to implement for u8 then please go ahead. The time you are reporting I guess are from your personal laptop ? VPI reports from a jetson which usually the ratio is 3 slower than a "normal" cpu

crates/kornia-imgproc/src/features.rs
```diff
@@ -68,6 +87,167 @@ pub fn hessian_response(src: &Image<f32, 1>, dst: &mut Image<f32, 1>) -> Result<
...
+
+    dst.as_slice_mut().par_chunks_exact_mut(src.cols())
+        // .zip(dx2_data.as_slice().par_chunks_exact(src.cols()))
+        // .zip(dx2_data.as_slice().par_chunks_exact(src.cols()))
+        // .zip(dx2_data.as_slice().par_chunks_exact(src.cols()))
+        // .zip(gx.as_slice().par_chunks_exact(src.cols()))
+        // .zip(gy.as_slice().par_chunks_exact(src.cols()))
+        .enumerate()
+        .for_each(|(row_idx, dst_chunk)| {
+            if row_idx == 0 || row_idx == src.rows() - 1 {
```
edgarriba, line 204: you could also use iterators and skip (3) till (rows-3) to avoid this ifelse statement. I've seen sometime some improvements in this situations.

crates/kornia-imgproc/Cargo.toml
```diff
@@ -25,6 +25,7 @@ image = "0.25"
 imageproc = "0.25"
 kornia-io = { workspace = true }
 ndarray = { version = "0.15", features = ["rayon"] }
+rand = "0.8.5"
```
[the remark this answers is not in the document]
andrew-shc: it should already be in dev dependencies

crates/kornia-imgproc/src/features.rs
andrew-shc: can you elaborate

andrew-shc:
Benchmark results for various SIMD implementations of `harris_response`. Tested on Intel i7-13700H.
| Method    | Time |
| -------- | ------- |
| (Original) Manual Summation  | 6.0ms   |
| Using `wide::f32x8` for `dx2`,`dy2`,`dxy` | 10.0ms     |
| Using `wide::f32x8` for `dx2`,`dy2`,`dxy` with `RUSTFLAGS="-C target-feature=+avx2"` | 11.1ms    |
| Using `wide::f32x8` for `dx2`,`dy2`,`dxy`, & `m11`,`m22`,`m12` with `RUSTFLAGS="-C target-feature=+avx2"` | 20.9ms    |

I guess SIMD/`wide` does not work well. I also tried `u8`/`i8` but realized there will be many numerical issues (even with `i16`). I think they meant it as the input image format itself, not as in they compute mostly in `u8`. Looking if there's anyway to bring 6ms down to maybe ~1s.

crates/kornia-imgproc/src/features.rs
edgarriba: I see, my bad

crates/kornia-imgproc/Cargo.toml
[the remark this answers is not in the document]
edgarriba: I see, my bad

crates/kornia-imgproc/src/features.rs
edgarriba: check if something in this direction could work avoid the if else in the inner loops

```rust
dx_data.as_mut_slice().par_chunks_exact_mut(src.cols()).skip(1).take(src.rows()-1)
```

andrew-shc:
Current progress:
| Method (1920x1080)   | Time |
| -------- | ------- |
| Manual Summation (retested) | ~6.1ms   |
| Manual Summation + slicing + w/o bounds checking (tested more) | ~4.4ms   |
| Manual Summation + w/o bounds checking | ~5.2ms   |
| Manual Summation + skip/take iter | ~5.7ms   |
| Manual Summation + slicing | ~5.7ms   |
| Manual Summation + prefix sum (of 3x3 window) + w/o bounds checking | ~12ms   |
| Manual Summation + prefix sum (of 3x3 window) | ~15ms   |
| Using `wide::f32x8` (`avx2`) + slicing | ~20ms    |
| Using `wide::f32x8` (`avx2`) w/ parallel chunk size of 4096 + slicing | ~14ms    |
| Using `wide::f32x8` (`avx2`) w/ parallel chunk size of 1024 + slicing | ~14ms    |
| Using `wide::f32x8` (`avx2`) w/ parallel chunk size of 256 + slicing | ~15ms    |
| Using `wide::f32x8` (`avx2`) w/ parallel chunk size of 16384 + slicing | ~14ms    |
| Using `wide::f32x4` (`avx`) w/ parallel chunk size of 16384 + slicing | ~14ms    |
| Using `wide::f32x4` (`avx`) w/ parallel chunk size of 4096 + slicing | ~13ms    |
| Using `wide::f32x4` (`avx`) w/ parallel chunk size of 1024 + slicing | ~14ms    |
| Using `wide::f32x4` (`sse`) w/ parallel chunk size of 1024 + slicing | ~14ms    |
| Using `wide::f32x4` (`sse`) w/ parallel chunk size of 4096 + slicing | ~14ms    |

andrew-shc:
Final comment:
The only (code) comments I've left are the `gaussian_blur` and `separable_filter`, which can be used later once they are fast enough. With the new builder pattern `HarrisResponse`, I was able to get it to run ~4.0ms (well within the range in VPI benchmark between u8 and s16).

This is ready to merge.

crates/kornia-imgproc/src/features.rs
```diff
@@ -82,6 +94,158 @@ pub fn hessian_response(src: &Image<f32, 1>, dst: &mut Image<f32, 1>) -> Result<
...
+                src.rows(),
+                dst.cols(),
+                dst.rows(),
+            ));
+        }
+
+        let src_data = src.as_slice();
+        let col_slice = src.cols()..src_data.len() - src.cols();
+        let row_slice = 1..src.rows() - 1;
+        let mut dx2_data = vec![0.0; src_data.len()];
```
edgarriba, line 149: for this strategy you might want to keep and allocate this buffers in the constructor (will require knowing the size)

andrew-shc: lmao thanks, never thought of this. It's now reaching ~1020µs (~1.02ms), and replacing slicing with `get_unchecked` slicing yields ~1000µs. It now complains 5-10 outliers presumably from the initial allocation. However, this does disallow parallelization per call if it ever becomes important.

crates/kornia-imgproc/src/features.rs
```diff
@@ -108,6 +110,9 @@ impl HarrisResponse {
             k: 0.04,
             grads_mode: GradsMode::default(),
             sigmas: 0.0,
+            dx2_data: Vec::new(),
```
edgarriba, line 113: probably the right thing is to pass ImageSize in the constructor to avoid a resize

andrew-shc:
Does this look good now?

crates/kornia-imgproc/src/features.rs
```diff
@@ -82,6 +93,179 @@ pub fn hessian_response(src: &Image<f32, 1>, dst: &mut Image<f32, 1>) -> Result<
...
+            grads_mode: GradsMode::default(),
+            sigmas: 0.0,
+            dx2_data: vec![0.0; image_size.width*image_size.height],
+            dy2_data: vec![0.0; image_size.width*image_size.height],
+            dxy_data: vec![0.0; image_size.width*image_size.height],
+        }
+    }
+
+    /// Sets the `k` value (usually between 0.04 to 0.06)
+    pub fn with_k(self, k: f32) -> Self {
```
edgarriba, line 122: can you add a small test to validate that you can concatenate calls in the builder pattern ? I'm curious to see how it works as I had implemented slightly different in here https://github.com/kornia/kornia-rs/blob/main/crates/kornia-io/src/stream/v4l2.rs#L47 maybe we should stick a common style

crates/kornia-imgproc/src/features.rs
```diff
@@ -163,6 +347,42 @@ mod tests {
         Ok(())
     }
 
+    #[test]
+    fn test_harris_response() -> Result<(), ImageError> {
+        #[rustfmt::skip]
+        let src = Image::from_size_slice(
+            [9, 9].into(),
```
edgarriba, line 354: it's a good practice to test with a non-squared image

crates/kornia-imgproc/src/features.rs
```diff
@@ -82,6 +93,179 @@ pub fn hessian_response(src: &Image<f32, 1>, dst: &mut Image<f32, 1>) -> Result<
...
+                dst.size().height,
+                self.image_size.width,
+                self.image_size.height,
+            ));
+        }
+
+        let src_data = src.as_slice();
+        let col_slice = src.cols()..src_data.len() - src.cols();
+        let row_slice = 1..src.rows() - 1;
+        // self.dx2_data.resize(src_data.len(), 0.0);
```
edgarriba, line 164: remove commented code

crates/kornia-imgproc/src/features.rs
```diff
@@ -82,6 +93,179 @@ pub fn hessian_response(src: &Image<f32, 1>, dst: &mut Image<f32, 1>) -> Result<
...
+                            // filter normalization
+                            *dx2_pixel = dx * dx;
+                            *dy2_pixel = dy * dy;
+                            *dxy_pixel = dx * dy;
+                        });
+                });
+        }
+
+
+        // gaussian_blur(&Image::from_size_slice(src.size(), &dx2_data)?, &mut dx2_blurred,
```
edgarriba, line 220: remove commented code

edgarriba (commented):
LGTM. Please, cleanup code and ready to go

andrew-shc:
Could you check again? For the builder pattern, it's just a variation (usually you have a BuilderStruct and Struct itself like the one you sent), but we can also say in the doc like in `features` all response computation follows this variation of builder pattern for extra clarity.

edgarriba (approved):
LGTM. Try to run the check in local with the `just` commands to pass tests in ci

crates/kornia-imgproc/src/features.rs
```diff
@@ -82,6 +91,183 @@ pub fn hessian_response(src: &Image<f32, 1>, dst: &mut Image<f32, 1>) -> Result<
     Ok(())
 }
 
+/// A builder object to initialize Harris response
+pub struct HarrisResponse {
+    image_size: ImageSize,
+    k: f32,
+    grads_mode: GradsMode, // TODO
```
edgarriba, line 98: can we just remove it for now untill we don't implement it. Thows some warnings, same for sigmas

edgarriba:
thanks! we can merge as soon as I find a fix for https://github.com/kornia/kornia-rs/pull/308

edgarriba:
@andrew-shc can you rebase from main as ci was fixed?

andrew-shc:
I think I accidentally set my working code for harris response on the main of my fork, so rebasing might be more compliacted.

Issue #171, opened 2024-11-05
Implement `adjust_brightness`


edgarriba:
UPDATE:

- implement from kornia pytorch: https://github.com/kornia/kornia/blob/main/kornia/enhance/adjust.py#L440

~~https://github.com/powei-lin/aprilgrid-rs/blob/353b9c4f08beb784758c36e51c92d5f9e93f72b1/src/quad.rs#L102~~

Rahuldrabit:
can i work on this issue ?

edgarriba:
sure

AnilKumarSingh9856:
Hi! I was directed here from Discord. The maintainers confirmed this feature is still wanted. I'd like to work on implementing adjust_brightness. Can you please assign this to me?

Issue #124, opened 2024-09-04
Add apriltag detector support


edgarriba:
Integrate ~~[`apriltag`](https://docs.rs/apriltag/latest/apriltag/)~~ [aprilgrid-rs](https://github.com/powei-lin/aprilgrid-rs) crate which is useful for robotics and camera calibration tasks.

Probably put into a new `kornia-calib` crate

powei-lin:
Hi, I removed those three functions in my crate. It's' more robust now.

powei-lin:
I added the support for kornia in aprilgrid in 0.5.0.
https://github.com/powei-lin/aprilgrid-rs/blob/cb4e56ee4bd1a24a074ada3d97571e21e351c6b3/tests/test_detector.rs#L37

MuhammadBilalKhan267:
Hello, I would like to contribute to this.

Aren't most already implemented and just need to be ported? More information on what exactly to implement and port would be appreciated.

edgarriba:
Supported natively

PR #269, opened 2025-03-15
added brightness adjustment feature `adjust_brightness`.

TomatoChocolate12:
solves issue #171  which is a subissue of #124 .

i added a function `adjust_brightness` that takes inspiration from the `aprilgrid` implementation, but is implemented with only `kornia-rs` functions. added tests as well for the same (both with a uniform and non-uniform image). 

i ran `just test` and it seemed to pass without failure. i also formatted the code to suit rust guidelines.


edgarriba (changes requested):
better to implement this  https://github.com/kornia/kornia/blob/main/kornia/enhance/adjust.py#L440

I updated the description on the original issue

PR #272, opened 2025-03-16
Moved Arithmetic functions to `tensor-ops`

aayushhzz:
This is the first step in 
_making `tensor` crate lightweight #271_

PR #280, opened 2025-03-18
tensops: fix self reference for internal methods

edgarriba:
fixes from https://github.com/kornia/kornia-rs/pull/272


edgarriba:
@aayushhzz some things detected from cargo clippy

Issue #137, opened 2024-09-18
Add support for `nvcodec` in the `StreamCapture`


edgarriba:
Implement hardware acceleration for the `StreamCapture` via `nvcodec`
*https://gstreamer.freedesktop.org/documentation/nvcodec/index.html?gi-language=c

This means extending the `Gstreamer` pipeline for capture as gstreamer has support already for it.

k3rnel-pan1c-a:
I added nvenc h264 encoding to the VideoWriter, should I open a PR?

edgarriba:
sure, go ahead

PR #281, opened 2025-03-18
Add NVIDIA hardware acceleration support for StreamCapture (#137)

Harshiths1:
- Implement NVCameraConfig with NV12 format support
- Create hardware-accelerated GStreamer pipeline using nvcodec
- Add NV12 to RGB conversion for efficient real-time streaming
- Update error handling for format detection
- Integrate with existing camera capture architecture
- Maintain backward compatibility with V4L2 capture

This implementation enables GPU-accelerated video decoding through the nvv4l2decoder element, providing better performance for high-resolution streams and reducing CPU usage. Both RGB and more lightweight NV12 formats are supported for different use cases.

### DIFF
--- a/crates/kornia-io/src/stream/capture.rs
+++ b/crates/kornia-io/src/stream/capture.rs
@@ -7,6 +7,7 @@ use kornia_image::Image;
 // utility struct to store the frame buffer
 struct FrameBuffer {
     buffer: gst::MappedBuffer<gst::buffer::Readable>,
+    sample: gst::Sample,
     width: usize,
     height: usize,
 }
@@ -94,16 +95,31 @@ impl StreamCapture {
             .last_frame
             .lock()
             .map_err(|_| StreamCaptureError::LockError)?;
-
+    
         last_frame.take().map_or(Ok(None), |frame_buffer| {
-            // TODO: solve the zero copy issue
-            // https://discourse.gstreamer.org/t/zero-copy-video-frames/3856/2
-            let img = Image::<u8, 3>::new(
-                [frame_buffer.width, frame_buffer.height].into(),
-                frame_buffer.buffer.to_owned(),
-            )
-            .map_err(|_| StreamCaptureError::CreateImageFrameError)?;
-
+            // Get the format from the caps
+            let caps = frame_buffer.sample.caps().ok_or_else(|| StreamCaptureError::GetCapsError)?;
+            let structure = caps.structure(0).ok_or_else(|| StreamCaptureError::GetStructureError)?;
+            
+            // Check if the format is NV12 or RGB
+            let format = structure.get::<String>("format")
+                .map_err(|_| StreamCaptureError::GetFormatError)?;
+            
+            let img = if format == "NV12" {
+                // Convert NV12 to RGB - note the Self:: prefix
+                Self::nv12_to_rgb(
+                    &frame_buffer.buffer,
+                    frame_buffer.width,
+                    frame_buffer.height,
+                )?
+            } else {
+                // Assume RGB format
+                Image::<u8, 3>::new(
+                    [frame_buffer.width, frame_buffer.height].into(),
+                    frame_buffer.buffer.to_owned(),
+                ).map_err(|_| StreamCaptureError::CreateImageFrameError)?
+            };
+    
             Ok(Some(img))
         })
     }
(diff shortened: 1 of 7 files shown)


crates/kornia-io/src/stream/capture.rs
```diff
@@ -7,6 +7,7 @@ use kornia_image::Image;
 // utility struct to store the frame buffer
 struct FrameBuffer {
     buffer: gst::MappedBuffer<gst::buffer::Readable>,
+    sample: gst::Sample,
```
edgarriba, line 10: why you need to keep the sample here ?

crates/kornia-io/src/stream/capture.rs
```diff
@@ -156,12 +172,61 @@ impl StreamCapture {
...
+            sample: sample.clone(),
             width,
             height,
         };
 
         Ok(frame_buffer)
     }
+
+    // Helper function to convert NV12 to RGB
+    fn nv12_to_rgb(
```
edgarriba, line 184: this should be part of kornia_improc::color

crates/kornia-io/src/stream/capture.rs
```diff
@@ -156,12 +172,61 @@ impl StreamCapture {
...
+        buffer: &[u8],
+        width: usize,
```
edgarriba, line 186:
```suggestion
        cols: usize,
```
to follow same convention as in the rest of library for rows/cols

crates/kornia-io/src/stream/capture.rs
```diff
@@ -156,12 +172,61 @@ impl StreamCapture {
...
+        height: usize,
+    ) -> Result<Image<u8, 3>, StreamCaptureError> {
+        // Allocate memory for RGB output
+        let mut rgb_data = vec![0u8; width * height * 3];
```
edgarriba, line 190: this should be part of the function signature

crates/kornia-io/src/stream/capture.rs
```diff
@@ -94,16 +95,31 @@ impl StreamCapture {
...
-
+            // Get the format from the caps
+            let caps = frame_buffer.sample.caps().ok_or_else(|| StreamCaptureError::GetCapsError)?;
+            let structure = caps.structure(0).ok_or_else(|| StreamCaptureError::GetStructureError)?;
+            
+            // Check if the format is NV12 or RGB
+            let format = structure.get::<String>("format")
+                .map_err(|_| StreamCaptureError::GetFormatError)?;
+            
+            let img = if format == "NV12" {
```
edgarriba, line 108: what's the benefit to decode in nv12 if the convert right away to rgb8 again ? I think this api should we designed in two modes (specified somehow in the constructor) so that the `grab` method can return either `ImageRgb8` or `ImageNv12` (the last one should be defined). Then consumers of this api can really share the encoded image. What do you think ?

edgarriba: maybe we need to rename the grab method to `grab_gray8` | `grab_rgb8` | `grab_nv12`
open to suggestions though

crates/kornia-io/src/stream/nvcamera.rs
```diff
@@ -0,0 +1,106 @@
...
+    // If NV12 format is requested, we'll use a different output format
+    let format_conversion = if use_nv12 {
+        // NV12 format - more lightweight for real-time applications
+        "! nvvidconv ! video/x-raw(memory:NVMM),format=NV12 ! appsink name=sink"
+    } else {
+        // RGB format - convert to standard RGB format
+        "! nvvidconv ! video/x-raw,format=RGB ! appsink name=sink"
+    };
+
+    format!(
```
edgarriba, line 102: can we give a thought about just extending the extending the existing v4lsrc pipeline ? In the end you are replacing the `nvdecoder` | `nvideoconver` only. I believe also that the output format should be somehow tied to the `StreamCapture` constructor params

edgarriba: Consider also that the ideal case to cover (not in this PR) is that when we use the nvidia plugins we can get the image in device to avoid copies from/to host/device

kornia-viz/src/bin/app.rs
```diff
@@ -27,7 +27,8 @@ struct KorniaApp {
 
 impl Default for KorniaApp {
     fn default() -> Self {
-        let capture = kornia::io::stream::V4L2CameraConfig::new()
+        // Use NVCameraConfig instead of V4L2CameraConfig for GPU acceleration
+        let capture = kornia::io::stream::NVCameraConfig::new()
```
edgarriba, line 31: I would make a new example instead using rerun for now. The kornia-viz was a dead end experiment which probably I'll remove soon as it need to much maintainership

crates/kornia-io/src/stream/nvcamera.rs
```diff
@@ -0,0 +1,106 @@
...
+        Self::new()
+    }
+}
+
+/// Returns a GStreamer pipeline string for capturing frames using NVIDIA hardware acceleration.
+pub fn nv_camera_pipeline_description(
+    device: &str, 
+    size: Option<ImageSize>, 
+    fps: u32,
+    use_nv12: bool
```
edgarriba, line 85: use enums intead

crates/kornia-io/src/stream/nvcamera.rs
```diff
@@ -0,0 +1,106 @@
+
+use std::any::Any;
+use crate::stream::{
+    camera::{CameraCapture, CameraCaptureConfig},
+    error::StreamCaptureError,
+};
+use kornia_image::ImageSize;
+
+/// A configuration object for capturing frames using NVIDIA hardware acceleration.
+pub struct NVCameraConfig {
```
edgarriba, line 10: i'd suggest to reuse as much as possible from v4lsrc struct

crates/kornia-io/src/stream/capture.rs
Harshiths1: I have included the sample here so as to access the caps (which describe the format, dimensions, frame-rate, and other metadata about the media) also I read that  GStreamer uses reference counting for memory management and by keeping the original sample we ensure that any references the buffer may have to the sample's memory are not invalidated.
But there is overhead to this approach so we could just extract info like "format: String" .
This was my thinking so how would you like me to proceed here let me know .

crates/kornia-io/src/stream/capture.rs
Harshiths1: I have added this function in a new file yuv.rs (based on color model ) in kornia-imgproc/src/color 
and then imported and used in the capture.rs

crates/kornia-io/src/stream/capture.rs
Harshiths1: For now should I update the grab method to grab_rgb8 as the consumers for now will only get rgb8 image

Harshiths1: > what's the benefit to decode in nv12 if the convert right away to rgb8 again ? I think this api should we designed in two modes (specified somehow in the constructor) so that the `grab` method can return either `ImageRgb8` or `ImageNv12` (the last one should be defined). Then consumers of this api can really share the encoded image. What do you think ?

I thought that this was the current requirement and most GPU's natively output nv12 hence the need to convert .

edgarriba:
@AS1100K can you take a look at this one ?

PR #287, opened 2025-03-19
implement copper example with multi cam

edgarriba:


examples/copper/Cargo.toml
```diff
@@ -0,0 +1,21 @@
+[package]
+name = "copper"
```
gbin, line 2: If you publish this crate it might be colliding with an existing one

examples/copper/Cargo.toml
```diff
@@ -0,0 +1,21 @@
...
+homepage.workspace = true
+include.workspace = true
+license.workspace = true
+readme.workspace = true
+repository.workspace = true
+rust-version.workspace = true
+version.workspace = true
+
+[dependencies]
+bincode = "2.0.0"
```
gbin, line 16: workspace = true ?

examples/copper/kornia_app.ron
```diff
@@ -0,0 +1,41 @@
+(
+    tasks: [
+        (
+            id: "cam0",
+            type: "copper::tasks::video_capture::VideoCapture",
```
gbin, line 5: We have a v4l src in copper, maybe reuse it and if we miss stuff add it back into it?
We also have a gstreamer source now if this is easier too

examples/copper/Cargo.toml
edgarriba: not publishing the examples

examples/copper/kornia_app.ron
edgarriba: I'd prefer to use the kornia apis, that's the main purpose of the example to show libraries integration

edgarriba: if there's any performance issue that would help also from our side to iterate on bug fixing

examples/copper/src/tasks/rerun_viz.rs
```diff
@@ -0,0 +1,99 @@
...
+                None,
+            )
+            .map_err(|e| CuError::new_with_cause("Failed to spawn rerun stream", e))?;
+
+        Ok(Self(rec))
+    }
+
+    fn process(&mut self, clock: &RobotClock, input: Self::Input) -> Result<(), CuError> {
+        let (img1, img2, img3) = input;
+        let timestamp_ns = clock.now().as_nanos();
```
edgarriba, line 53: @gbin is there any particular way to get the adquisition/received time from each of the tuple elements?

gbin: img1.metadata.tov
if the source put it correctly :)

PR #296, opened 2025-03-19
fix separable filter overflows

edgarriba:
image


edgarriba:
@andrew-shc it definitely crashed by an overflow. One hack could be to use saturated addition but won't correct at all. Any other suggestions besides reverting and having a separated kernel per type ?

edgarriba:
@andrew-shc added a trait to implement u8 and f32. If you see the benchmark there's no much difference. Wondering if in case of u8 we just cast to f32

andrew-shc:
We should just cast u8 and other types T to f32 for now. I'm 70% sure opencv does this internally and pretty much all DL libraries out there just use f32 explicitly. But this does bring a problem when T is f64 and we're casting down to f32. I'm just looking at the `burn ` crate right now and it seems like they define their int, float, and bool primitives (through tensor) by each backends (eg cuda, wgpu, etc). https://docs.rs/burn/latest/burn/prelude/trait.Backend.html https://github.com/tracel-ai/burn/blob/main/crates/burn-cuda/src/lib.rs

Issue #132, opened 2024-09-10
Add sccache support in the CI


edgarriba:
Use `sccache` in the CI in order to bootstrap compilation time

- https://crates.io/crates/sccache

TempestBirds729804:
I'd like to try to do this issue,  it seems that reducing repetitive compilation time is really useful. 🙂

edgarriba:
sure, go ahead

TempestBirds729804:
I've set up the YAML workflow file following the sccache docs. What do you think is a reasonable cache size for kornia-rs? I feel like 6GB should be good enough. (I used 2GB when running tests locally) 🤔

`    env:
      RUSTC_WRAPPER: sccache
      SCCACHE_GHA_ENABLED: "true"
      SCCACHE_CACHE_SIZE: 2G
      SCCACHE_ERROR_LOG: /tmp/sccache_log.txt
      SCCACHE_LOG: info
`

PR #298, opened 2025-03-20
Add sccache to ci

TempestBirds729804:
This PR introduces sccache to the Rust workflow with some basic configurations.

-       RUSTC_WRAPPER: sccache
        SCCACHE_GHA_ENABLED: "true"
        SCCACHE_CACHE_SIZE: 6G
        SCCACHE_ERROR_LOG: /tmp/sccache_log.txt
        SCCACHE_LOG: info

After getting some CI runtime data, we can consider tuning things like cache size, log settings, and sccache’s max frame length.
I also noticed sccache has a distributed mode, but I don't think it's necessary for this project.

PS: I'm still learning, so feel free to point out anything I missed! Related issue is #132


edgarriba:
I think something went wrong possibly with the integration of cross ?

TempestBirds729804:
Wow that is Unexpected.🤔 I took a look at the error message, and it seems like the error happened in `arrow-arith`, version 53.4.0, where there’s some ambiguity in a method call in the `temporal.rs` file. Specifically, `DatePart::Quarter => |d| d.quarter() as i32`.

The `quarter()` method has two possible implementations: one defined in the `ChronoDateExt` trait and another in the `Datelike` trait. E0034 seems to mean the compiler can't figure out which implementation to use. 

It could be that after I enabled `sccache`, the CI picked up the latest version of the dependencies, and this version has a conflict. Maybe arrow-arith 53.4.0 ` is the most recent update that introduced this conflict? But, it could also be that enabling `sccache` changed how dependencies are resolved, since `sccache`, being a compiler wrapper (RUSTC_WRAPPER), can alter the build behavior. I’ll need to investigate further.

TempestBirds729804:
屏幕截图 2025-03-22 192331

Crates.io page right here, this thing did update 19 hours ago, but it doesn't match the version number(arrow-arith v53.4.0) in the error message...🤔

edgarriba:
that's weird, something proably something in rerun or dora-rs changed as they are the only one using arrow packages (we compile them only for the examples)

https://github.com/rerun-io/rerun/issues/9159

TempestBirds729804:
I believe sccache could work now

edgarriba:
@ngoldbaum in case you want to help with review this CI work

ngoldbaum:
I think if you rebase the test failure will go away. Also why not also set up sccache in the python test workflow as well?

TempestBirds729804:
@ngoldbaum Thank you for reply! I didn't add it for Python because I think, Python is an interpreted language, doesn't have an explicit compilation process like Rust does. Maybe there are Python projects depend on some extensions written in Rust? but in that case, the use of Rust is only indirect. So for now, I decided to add sccache only for Rust. This is my personal understanding.

ngoldbaum:
The python tests build the rust bindings to kornia-rs as well as a large number of rust packages that the python bindings depend on. Adding sccache will definitely help.

TempestBirds729804:
Good catch🤔, I'll work on it.

TempestBirds729804:
Hi! Is this ok to be accepted?

edgarriba (approved):

Issue #275, opened 2025-03-17
fix the `distance_transform` function within `distance_transform.rs`


TomatoChocolate12:
there is some issue with the `distance_transform` function, and this could cause more problems down the line as this function is essential for implementing solutions for issues #169 and #170 .

edgarriba:
the distance transform function it's not working right now. It needs to be rewritten following this repository: https://github.com/giorgiomarcias/distance_transform

TomatoChocolate12:
thanks for the reference! i am working on it.

aviralgarg05:
I would like to solve the issue, please assign me

zeel991:
Hi, I would like to work on this issue!

I have reviewed the current code and the reference repository. I plan to implement the linear-time Euclidean distance transform using the Felzenszwalb & Huttenlocher (2012) algorithm. I already have a working prototype of the 1D parabolic lower envelope logic ready to integrate.

Quick question on the output format: Does kornia-rs strictly require the output to be the Euclidean distance ($L_2$) (i.e., taking the square root at the end), or would you prefer Squared Euclidean ($L_2^2$) to save on computational cost?I can implement either based on your preference. Please assign this to me!

Issue #170, opened 2024-11-05
Implement `mophology::dilate`


edgarriba:
https://github.com/powei-lin/aprilgrid-rs/blob/353b9c4f08beb784758c36e51c92d5f9e93f72b1/src/quad.rs#L113

edgarriba:
implement via separable convolution

Mars-204:
Hi, As required, I would like to implement the dilate function using separable convolution.

edgarriba:
please go ahead, reminder to provide benchmarks and tests

Mars-204:
Issue already solved here #124 .

dhruv6677:
Hi, I’d like to work on this issue as my first contribution. Could you please assign it to me?

edgarriba:
@dhruv6677 any thoughts about how to implement it ? consider joining discord before coding to discuss the implementation

dhruv6677:
@edgarriba 
My plan is to implement the dilate() function inside the imgproc crate, similar to other morphology operations.
The function will apply morphological dilation using a fixed or user-defined kernel, operating over grayscale images (DynamicImage), and return the resulting image.
I’ll also add tests and examples for validation.

I’ll join the Discord to confirm the implementation details before I start coding.

edgarriba:
please, understand first the codebase, e.g kornia-rs doesn't have any `DynamicImage` or morphological operators yet

dhruv6677:
Got it, thanks for the clarification!
I explore the codebase first in detail

Issue #169, opened 2024-11-05
`find_countours`


edgarriba:

Udit-git-acc:
Is this issue resolved ?

edgarriba:
@Udit-git-acc not yet, but please focus on finishing first other opened PRs first

mathurojus:
Hi @edgarriba ,
I'd like to take this up. Familiar with the codebase and have reviewed the references. Ready to start whenever

edgarriba:
which reference code are you planning to use?

mathurojus:
Hi @edgarriba ,
I plan to use the aprilgrid-rs quad.rs implementation as the main code reference, and follow OpenCV’s findContours documentation for mode/method semantics and API behavior.
Should I start with the same approach ?

edgarriba:
i updated the issue, it should mimic opencv at first. Read carefullt the development guidelines, and local tests should pass before submitting a PR

mathurojus:
Hi @edgarriba, I've submitted a PR (#611) with the find_contours implementation. Is the implementation looking good? Any feedback or changes needed?

VyasGuru:
Hi, I'd like to work on this if the current pr is inactive

edgarriba:
@VyasGuru what plans you have or ref implementation you have in mind ?

VyasGuru:
> [@VyasGuru](https://github.com/VyasGuru) what plans you have or ref implementation you have in mind ?

the opencv version as it is. The one you linked to, treats any non zero value as 1, and doesn't threshold using any other method so i suppose ill do the same. also the different modes it has ie external contours, 2 level contours etc. do lmk if there are any other requests.

zeel991:
Hi @sidd-27 , I'd like to work on this issue, I've gone through the OpenCV docs and the Suzuki-Abe border following algorithm and all other functions they have used. Please assign this to me

I will raise the PR with all the functions and their tests and benchmarks

mathurojus:
Hey @sidd-27 , I’ve opened PR #797 to add find_contours with OpenCV‑style hierarchy modes to kornia-imgproc. When you get a chance, could you please review it and let me know if anything needs to be adjusted, or approve/merge it if it looks good?

zeel991:
hey @sidd-27 I had already taken this issue on your confirmation on discord and have already raised a PR for it already approved by you

PR #300, opened 2025-03-21
added `find_contour`, `dilate` features, along with support for `distance_transform` and `draw_polygon`

TomatoChocolate12:
### this PR solves issues:
- #169 , #170 which are sub-issues of #124 . (`find_contour` and `dilate`)
- #275 `distance_transform` fix.

### it also adds the features:
- `draw_polygon` : to draw a polygon given vertices.

### how this should be extended?
- the tests for the `find_contours` function is not done with `draw_polygon` because the algorithm interprets the inner side of the polygon edges to be holes.
- so, a `draw_filled_polygon` should be implemented which can be used in these tests. i tried doing this with the `scanline fill` algorithm, but it was not very successful and so i did not use it.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -3,10 +3,15 @@ resolver = "2"
 members = [
     "crates/*",
     "examples/*",
+    # NOTE: need to add dora examples to make rust analyzer to work
+    "examples/dora/video-capture",
+    "examples/dora/imgproc",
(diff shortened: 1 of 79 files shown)


edgarriba (changes requested):
please submit one method at a time in separated PRs

TomatoChocolate12:
done :+1:

PR #301, opened 2025-03-21
added `dilate` functionality and fixed `distance_transform`

TomatoChocolate12:
### this PR solves issues:
- #170 which is a sub issue of #124 
- #275 

### it implements:
- a dilation functionality.
- fixes distance transform which is used in several other algorithms.


edgarriba:
please, split each functionality in a separated PR

PR #303, opened 2025-03-21
Create `kornia-lie` crate on top of glam-rs

### LINKED ISSUE #268: Create kornia-liealg crate with basic implementation of lie algebra on top of glam-rs
this probably could be the starting point for a kornia-liealg crate (i'll open a separated ticket for this)
ref: https://kornia.readthedocs.io/en/latest/geometry.liegroup.html

_Originally posted by @edgarriba in https://github.com/kornia/kornia-rs/pull/263#discussion_r1993150074_

andrew-shc:
The initial PR is not complete, but I'm opening because I want to check if the general crate design is consistent to what you want. I've only implemented `SO3`, but once it looks good I can just quickly copy the design to `SO2`, `SE3`, and `SE2`.

Closes #268


crates/kornia-liealg/src/se3.rs
```diff
@@ -0,0 +1,5 @@
+pub struct SE3 {
+    pub quaternion: [f32; 4],
```
edgarriba, line 2: use glam Quat ?

edgarriba (commented):
structure looks good. Check here: https://github.com/kornia/kornia/tree/main/kornia/geometry/liegroup there are some tests that could be reusable. /cc @cjpurackal if you want to review this piece

crates/kornia-liealg/src/se3.rs
```diff
@@ -0,0 +1,5 @@
...
+    pub translation: [f32; 3],
```
edgarriba, line 3: Vec3

crates/kornia-liealg/src/so3.rs
```diff
@@ -0,0 +1,135 @@
+use std::fmt::Debug;
+use std::ops::Mul;
+use glam::{Affine3A, Mat3, Mat3A, Mat4, Quat, Vec3};
```
edgarriba, line 3: we might consider having also the D (f64) variant

cjpurackal:
I think the fitting name for the crate would be kornia-lie as it abstracts away the group and algebra aspects.

crates/kornia-liealg/src/so3.rs
andrew-shc: just use the D variant or like a generic over or two separate implementation?

andrew-shc:
@edgarriba I see that SE3 returns a concatenated [1x6](https://github.com/kornia/kornia/blob/57296883b1500e87c2a5c92ad76d3708f65dac8b/kornia/geometry/liegroup/se3.py#L220C8-L220C13) vector for rotation and translation when calling `.log()`. Glam does not have this. Do you think we should implement our own or have it return two 1x3 vectors?

I think this also relates to whether we want it to be batched or not. In my opinion, batching would be much easier/cleaner on a single vector, but would mean we have to re-implement Vec6.

edgarriba:
if you see how it's used later the vec6 it's just split again. So yes, maybe we return a tuple of Vec3. Not sure if there's any other other use per se of the 6d vector

andrew-shc:
@edgarriba Adjoint matrices most likely needs to be 6x6. Your python impl is in that and an online [book (pg. 226 7.1.4)](http://asrl.utias.utoronto.ca/~tdb/bib/barfoot_ser17.pdf) says it is in 6x6. [`glam`](https://docs.rs/glam/latest/glam/) doc states its (only) for games and graphics. You sure you still want this to be in `glam`? (It has other issues too like hard to use with generics).

edgarriba:
@andrew-shc how are this 6x6 used later ? never had to implement by myself. Maybe it's fine just to provide a simple interface for Mat6 for now out of native arrays? The main idea to use glam is that all the basics are there compacted in a tiny light weighted library and quite robust for prod (I validated that in other projects) and really for simd

cjpurackal:
what are current blockers to merging this other than the adjoint issue? I can take it up if you're not working on this @andrew-shc @edgarriba.

andrew-shc:
Was working on this originally to familiarize with the kornia libraries. I think feature-wise it should be done and it’s just that one aforementioned issue. I’m not sure if @edgarriba wants benchmarking on this too. We also might need to implement the double (f64) variant too (currently it’s all f32).

edgarriba:
it's totally fine to me to merge. We can iteratively improve with small PR to make it feature complete

crates/kornia-lie/src/se2.rs
```diff
@@ -0,0 +1,203 @@
...
+        ])
+    }
+
+    pub fn inverse(&self) -> Self {
+        Self { r: self.r.inverse(), t: self.r.inverse() * (-self.t) }
+    }
+
+    pub fn adjoint(&self) -> Mat3A {
+        let mut mat = self.matrix();
+        mat.col_mut(0)[2] = self.t[1];
```
edgarriba, line 47: can we access directly to the mat public members, might be slighlty faster

crates/kornia-lie/src/so3.rs
```diff
@@ -0,0 +1,232 @@
...
+    pub fn from_random() -> Self {
+        let mut rng = rand::rng();
+
+        let r1: f32 = rng.random();
+        let r2: f32 = rng.random();
+        let r3: f32 = rng.random();
+
+        let w = (1.0 - r1).sqrt() * (2.0 * std::f32::consts::PI * r2).sin();
+        let x = (1.0 - r1).sqrt() * (2.0 * std::f32::consts::PI * r2).cos();
+        let y = r1.sqrt() * (2.0 * std::f32::consts::PI * r3).sin();
```
edgarriba, line 36: make a reusable `r1_sqrt` variable

crates/kornia-lie/src/so3.rs
```diff
@@ -0,0 +1,232 @@
...
+        Mat3A::from_cols_array(&[
+             0.0, -v.z, v.y,
+             v.z,  0.0,-v.x,
+            -v.y,  v.x, 0.0,
+        ])
+    }
+
+    /// Lie algebra -> vector space
+    pub fn vee(omega: Mat3A) -> Vec3A {
+        Vec3A::from_array([omega.col(2)[1], omega.col(0)[2], omega.col(1)[0]])
```
edgarriba, line 99: access directly to the Mat3A public members

crates/kornia-lie/Cargo.toml
```diff
@@ -0,0 +1,19 @@
...
+publish = true
+repository.workspace = true
+rust-version.workspace = true
+version.workspace = true
+
+[dependencies]
+glam = "0.30.1"
+rand = "0.9.0"
+
+[dev-dependencies]
```
edgarriba, line 18: remove

crates/kornia-lie/src/se2.rs
```diff
@@ -0,0 +1,203 @@
...
+        let r = self.r.matrix();
+        Mat3A::from_cols_array(&[
+            r.x_axis.x, r.y_axis.x, self.t.x,
+            r.x_axis.y, r.y_axis.y, self.t.y,
+            0.0, 0.0, 1.0,
+        ])
+    }
+
+    pub fn inverse(&self) -> Self {
+        Self { r: self.r.inverse(), t: self.r.inverse() * (-self.t) }
```
edgarriba, line 42: compute inverse only once

cjpurackal:
Cool, i think we can merge it once we resolve what @edgarriba has raised and I'll address the other issues/ feature complete on a different PR.

andrew-shc:
@cjpurackal I'm going to open and merge this pr once I finish the 6x6 Mat implementation for those special methods by tomorrow and you can pull it directly from the main instead of going through my fork

crates/kornia-lie/Cargo.toml
andrew-shc: well just added an `approx` crate for floating point comparison

crates/kornia-lie/src/so3.rs
```diff
@@ -0,0 +1,238 @@
...
+        ident - ((1.0 - theta.cos()) / theta.powi(2)) * skew
+            + ((theta - theta.sin()) / theta.powi(3)) * (skew * skew)
+    }
+}
+
+impl Mul for SO3 {
+    type Output = SO3;
+
+    fn mul(self, rhs: Self) -> Self::Output {
+        todo!()
```
edgarriba, line 132: Maybe just implement it, actually the other structs miss the Mul trait too

edgarriba (approved):
LGTM add minor changes. Make sure that tests cover the same or very similar to the Kornia PyTorch implementations to cover some corner cases

edgarriba:
It’s also missing some nice docs with formulation. That can be done in a next PR

PR #305, opened 2025-03-22
Updating onnx example to resize image to model input shape.

kylekam:
ONNX example currently doesn't work unless the image already matches the model input shape. Resizing allows any 3 channel image to be used.

### DIFF
--- a/examples/onnx/src/main.rs
+++ b/examples/onnx/src/main.rs
@@ -3,11 +3,13 @@ use kornia_tensor::{CpuAllocator, Tensor};
 use std::path::PathBuf;
 use std::time::Instant;
 
-use kornia::image::Image;
+use kornia::image::{Image, ImageSize};
 use kornia::io::functional as F;
+use kornia::imgproc;
 
 use ort::session::builder::GraphOptimizationLevel;
 use ort::session::Session;
+use ort::value::ValueType;
 
 /// Represents a detected object in an image.
 #[derive(Debug)]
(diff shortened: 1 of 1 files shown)


examples/onnx/src/main.rs
```diff
@@ -52,18 +54,38 @@ fn main() -> Result<(), Box<dyn std::error::Error>> {
...
-
     let model = Session::builder()?
         .with_optimization_level(GraphOptimizationLevel::Level3)?
         .with_intra_threads(4)?
         .commit_from_file(&args.onnx_model_path)?;
 
+    // get the model input dimensions, assumes (B, C, H, W) format
+    let mut width = 0;
+    let mut height = 0;
+    if let ValueType::Tensor{dimensions,..} = &model.inputs[0].input_type {
```
edgarriba, line 65: let (width, height) = match &model.inputs[0].input_type {
      ValueType::Tensor{dimensions,..}  {}
      Err(_) => { return Err(some_error)
}

or something similar

edgarriba (changes requested):
Have you actually tried ? you can use one of these models: https://huggingface.co/kornia/ONNX_models/tree/main/models

can you also update the readme in the example exaplaining how and where to download the model ot use

kylekam:
> Have you actually tried ? you can use one of these models: https://huggingface.co/kornia/ONNX_models/tree/main/models
> 
> can you also update the readme in the example exaplaining how and where to download the model ot use

Ah okay, I didn't realize there was a huggingface location for the models. I confirmed this issue using RT-DETR from their official repo, but if the intended use case is to only use kornia models, then I can just update the readme and remove the other changes.

edgarriba:
did it work at all with theirs ? our version provides some pre/post processings

kylekam:
> did it work at all with theirs ? our version provides some pre/post processings

Initially there was a shape mismatch, but adding the resizing seemed to fix the input issue. As for post processing, I think stuff like NMS needs to be implemented as well, so maybe it's best to recommend that people use the kornia models.

kylekam:
Reverted the changes, linked kornia ONNX, and updated the README for clarification.

johnnv1 (approved):

edgarriba:
can you rebase from `main` ?

PR #307, opened 2025-03-22
fix: prevent overflow in separable_filter for u8 max values

### LINKED ISSUE #293: Add test for `separable_filter` with `u8` using max values 255
possibly there's an overflow right now with the current implementation

SanjayUG:
# Fix Overflow in Separable Filter for u8 Max Values

## Description
This PR addresses an issue where the `separable_filter` function could experience overflow when processing `u8` images with maximum values (255). The fix ensures proper handling of maximum values while maintaining the expected behavior for all other cases.

## Changes
- Added trait bounds `Into<u16>` and `From<u16>` to the generic type `T` to support intermediate calculations
- Implemented intermediate conversion to `u16` to prevent overflow during calculations
- Added result clamping to ensure values stay within the valid `u8` range (0-255)
- Added a test case `test_separable_filter_u8_overflow` to verify the fix

## Testing
Added a new test case that:
- Creates an image filled with maximum `u8` values (255)
- Applies a separable filter with a kernel that would cause overflow
- Verifies that the output remains at 255 without any overflow issues

## Impact
This fix ensures that:
- The separable filter works correctly with `u8` images containing maximum values
- No overflow occurs during intermediate calculations
- The original functionality is preserved for non-overflow cases
- The solution is robust for any kernel size and values

## Example
```rust
let img = Image::<u8, 1>::new(
    ImageSize { width: 3, height: 3 },
    vec![255; 9], // All pixels set to 255
)?;
let mut dst = Image::<u8, 1>::from_size_val(img.size(), 0)?;
let kernel = vec![1, 1, 1];
separable_filter(&img, &mut dst, &kernel, &kernel)?;
// Result will be 255 for all pixels without overflow
```

## Related Issue
Fixes #293

### DIFF
--- a/crates/kornia-imgproc/src/filter/separable_filter.rs
+++ b/crates/kornia-imgproc/src/filter/separable_filter.rs
@@ -16,7 +16,7 @@ pub fn separable_filter<T, const C: usize>(
     kernel_y: &[T],
 ) -> Result<(), ImageError>
 where
-    T: Zero + std::ops::Mul<Output = T> + std::ops::AddAssign + Copy,
+    T: Zero + std::ops::Mul<Output = T> + std::ops::AddAssign + Copy + Into<u16> + From<u16>,
 {
     if kernel_x.is_empty() || kernel_y.is_empty() {
         return Err(ImageError::InvalidKernelLength(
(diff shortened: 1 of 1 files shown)


edgarriba:
Fixed here https://github.com/kornia/kornia-rs/pull/296

edgarriba:
fixed in #296

Issue #185, opened 2024-12-18
Implement Voxel-grid


edgarriba:
https://pointclouds.org/documentation/classpcl_1_1_voxel_grid.html

HelloDuoLA:
Hi, I'm a first time contributor and I would like to participate in GSoC2025. The VoxelGrid class in the PCL library has a lot of functions, can I implement some basic ones first?

edgarriba:
hi! yep, this one can be a bit tricky. I suggest to start somehow with a simple struct and then improve it over time

ajit2704:
Hey @HelloDuoLA , this seems like a good amount of work. I'd like to help.

edgarriba:
If anyone has interest please, reopen

PR #309, opened 2025-03-23
implemented voxel grid filter based on voxel size against issue #185

Udit-git-acc:
Voxel grid filter using voxel size and method in python binding


edgarriba:
Please, read the code in the reference library. This current implementation doesn’t make any sense

PR #311, opened 2025-03-25
fix: resolve conflict between arrow-arith and chrono

TempestBirds729804:
Perhaps the previous changes didn’t take effect? This time, I declared the dependencies under [workspace.dependencies], using version pinning (=53.3.0 and =0.4.39). I also disabled the default features of chrono to reduce conflicts, and tried to ensure that all subprojects correctly inherit these versions by using workspace = true as you added (the effectiveness still needs to be verified through CI testing).
#308

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -40,6 +40,9 @@ kornia = { path = "crates/kornia", version = "0.1.9-rc.2" }
 kornia-linalg = { path = "crates/kornia-linalg", version = "0.1.9-rc.2" }
 kernels = { path = "crates/kernels", version = "0.1.9-rc.2" }
 
+arrow-arith = { version = "=53.3.0", features = [] }
+chrono = { version = "=0.4.39", default-features = false }
+
 # dev dependencies for workspace
 argh = "0.1"
 approx = "0.5"


edgarriba:
I haven’t tried yet, but maybe downgrading rerun version could work

viktorho:
Hi @TempestBirds729804 and @edgarriba,
I tried to tackle this issue by using the [patch.crates-io] section to override the default source for the chrono crate from crates.io with the chronotope GitHub repository
this seems to work in my case
`
[patch.crates-io]
chrono = { git = "https://github.com/chronotope/chrono", tag = "v0.4.39" }
`

TempestBirds729804:
@viktorho  
Has Rust cross-compiling been set up and working as well?

edgarriba:
I've been trying in parallel here https://github.com/kornia/kornia-rs/pull/308 and for what i've found is that dora 0.3.10 requires arrow 54.2.1 -- but then rerun tries always to pull a recent version of arrow which take priority over dora then. @viktorho try to compile with cross build to double check if you managed to fix

TempestBirds729804:
Hey bro, I found the conflicted versions in Cargo.lock was never updated after I checked it! 
屏幕截图 2025-03-26 210243
屏幕截图 2025-03-26 210731
 So, I want to try updating it manually.

Issue #226, opened 2025-01-22
Ship free-threaded (cp313t) wheels


ngoldbaum:
It looks like you've already updated to PyO3 0.23, so all that's left is adding CI on the free-threaded interpreter, adding testing for thread safety, and fixing any issues that are discovered in the process. 

There is some multithreaded testing and functionality on the rust side, but it looks like there isn't any testing of using `kornia-py` from the Python `threading` module.

I'm a PyO3 maintainer and helped lead the effort to ship support for free-threaded python in PyO3 0.23. I'm also actively working on ecosystem support for free-threaded Python as part of my job at quansight labs.

All that to say that I'm happy to do all the heavy lifting here, I'd just appreciate being able to ask questions and get code review.

edgarriba:
hi @ngoldbaum thanks for reaching out ! Would be awesome if you could lead that part. I've been recently focusing more on the Rust side of the project so any help to improve the Python bindings is very welcomed. For awareness, recently @johnnv1 have been helping a bit too here. Feel also free to join our discord to discuss more: https://discord.com/invite/HfnywwpBnD

edgarriba:
I believe this is done right ‘

ngoldbaum:
No this isn't done yet. I'm waiting on a PyO3 and rust-numpy 0.24 release before I can start working on kornia-rs.

ngoldbaum:
I updated the pyo3 and rust-numpy dependencies to 0.24.0 and it looks like the existing tests pass:

```
(kornia-py) goldbaum@goldbaum-QEMU-Virtual-Machine:~/kornia-rs/kornia-py$ PYTHON_GIL=0 pytest
============================================================= test session starts =============================================================
platform linux -- Python 3.13.2, pytest-8.3.5, pluggy-1.5.0
rootdir: /home/goldbaum/kornia-rs/kornia-py
configfile: pyproject.toml
collected 16 items                                                                                                                            

tests/test_color.py ...                                                                                                                 [ 18%]
tests/test_enhance.py .                                                                                                                 [ 25%]
tests/test_histogram.py .                                                                                                               [ 31%]
tests/test_icp.py ..                                                                                                                    [ 43%]
tests/test_io.py .....                                                                                                                  [ 75%]
tests/test_package.py .                                                                                                                 [ 81%]
tests/test_resize.py .                                                                                                                  [ 87%]
tests/test_warp.py ..                                                                                                                   [100%]

[...]

PR #315, opened 2025-03-25
Update python bindings to be compatible with free-threading, add CI

ngoldbaum:
See https://github.com/kornia/kornia-rs/issues/226#issuecomment-2749385053 for more context.

### DIFF
--- a/.github/workflows/python_test.yml
+++ b/.github/workflows/python_test.yml
@@ -42,6 +42,39 @@ jobs:
     - name: Run tests
       run: cd kornia-py/ && just test
 
+  # TODO: integrate with linux tests once uv bug is fixed
+  # https://github.com/astral-sh/uv/issues/12445
+  test-python-linux-free-threaded:
+    name: py3.13t-linux
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v4
+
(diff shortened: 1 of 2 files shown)


edgarriba:
@ngoldbaum can you try to rebase on main. We added some fixes for the ci

ngoldbaum:
Hmm, I guess that revealed deeper issues. I’ll try to take  a closer look at this soon.

edgarriba (approved):

edgarriba:
bless copilot :)

ngoldbaum:
It looks like setting up two different uv environments led to issues. Fixed by removing the `python-version` from the `setup-uv` use.

edgarriba:
Thanks so much ! One question, do you have any experience with the release pipelines using maturin ? I’ve had some trouble in the past e.g. to include the gstreamer dependency which blocks from adding cool features like video stuff from python side

ngoldbaum:
>  do you have any experience with the release pipelines using maturin

A little bit on other projects in the past few months, mostly to add free-threaded wheels.

Is there an issue somewhere?

edgarriba:
I don’t have any open issue for this (I’ll create one), maybe first would be interesting to see how we can support from ci adding the gstreamer feature. Possibly the rest will come together

Issue #179, opened 2024-11-29
Implement TSDF


edgarriba:
Reference: https://www.open3d.org/docs/0.14.1/python_api/open3d.pipelines.integration.TSDFVolume.html?highlight=tsdf#open3d.pipelines.integration.TSDFVolume

k3rnel-pan1c-a:
Hi, is it alright if I start working on this?

edgarriba:
Totally!

PR #323, opened 2025-03-29
Add thread-safe TSDF volume integration using Mutex for parallel processing

GautamSharmaaa:
Fixes https://github.com/kornia/kornia-rs/issues/179

# Title
Add thread-safe TSDF volume integration using Mutex for parallel processing

# Description
## Overview
This PR implements thread-safe Truncated Signed Distance Function (TSDF) volume integration by using a thread-safe approach with a Mutex to synchronize updates from parallel processing. The implementation leverages Rayon's parallel iterators for performance gains while ensuring data consistency.

## Changes
- Modified the `integrate` method in the TSDF volume to allow concurrent updates
- Added thread-safe update mechanism using a Mutex for coordinating parallel voxel processing
- Fixed test cases for proper validation of TSDF functionality
- Improved floating-point comparisons for better numerical stability 
- Added proper documentation and cleaned up unused code

## Implementation Details
The core of the thread-safe implementation uses a Mutex to protect shared data during parallel updates:

```rust
// Create a thread-safe vector for updates
let updates = Mutex::new(Vec::new());

// Process all voxels and collect updates in parallel
(0..self.dims[2]).into_par_iter().for_each(|z| {
    let mut local_updates = Vec::new();
    for y in 0..self.dims[1] {
        for x in 0..self.dims[0] {
            // Process voxel and collect updates locally
            // ...
            local_updates.push((idx, tsdf_val, weight));
        }
    }
    
    // ThreadLocal updates are pushed to global updates
    if !local_updates.is_empty() {
        let mut global_updates = updates.lock().unwrap();
        global_updates.extend(local_updates);
    }
});

// Apply all updates at once
let updates = updates.into_inner().unwrap();
for (idx, tsdf_val, weight) in updates {
    // Update TSDF and weight data
    // ...
}
```

## Why This Approach
1. **Thread safety**: Using a Mutex ensures that updates to the global collection are coordinated between threads, preventing data races.

2. **Reduced lock contention**: By using thread-local collections and then merging them with a single lock acquisition per thread, we minimize lock contention compared to locking for each voxel update.

3. **Improved performance**: Parallel processing of z-layers allows for better CPU utilization on multi-core systems while maintaining data consistency.

4. **Memory efficiency**: The approach avoids unnecessary memory allocation by using a single collection for all updates.

## Testing
All tests have been verified to pass, including:
- Basic TSDF creation and property verification
- Depth integration from multiple viewpoints
- Mesh extraction
- Error handling for invalid inputs

## Example Usage
An example application has been created to demonstrate the usage, integrating a synthetic depth image from multiple viewpoints to create a TSDF volume.


edgarriba (changes requested):
This looks exactly the same as another open PR in this repo. No copies are allowed, please provide reference implementation

PR #262, opened 2025-03-12
Add support for writing images in PNG and JPEG formats. Fixes #122

suhailnadaf509:
- Added write_image_png_gray8 for PNG grayscale images.
- Added write_image_png_rgb8 for PNG RGB images.
- Added write_image_jpg_gray8 for JPEG grayscale images.
- Added write_image_jpg_rgb8 for JPEG RGB images.
- Utilizes the image crate for encoding, with JPEG quality set to 90.
- Integrated with existing Image struct from kornia-image.
- Tested with sample images (e.g., tests/data/dog_gray.png, dog.jpg).
-This adds functions to the kornia-io crate for saving both grayscale (gray8) and RGB (rgb8) 8-bit images in PNG and JPEG formats. Each function takes an Image<u8, N> (N=1 or 3) and returns a Result<(), ImageError> for robust error handling.

Closes #122
Closes #252
Closes #253
Closes #254
Closes #255

PR #327, opened 2025-04-01
Add JPEG image encoding and decoding functions

### LINKED ISSUE #254: implement `write_image_jpg_rgb8`

### LINKED ISSUE #255: implement `write_image_jpg_gray8`

AS1100K:
Fixes #254
Fixes #255

I have added encoding and decoding functions for JPEG file format.

In the issue #254 and #255, it says `jpg` in the function name, but I went with `jpeg` because they are interchangeable extensions and there is already jpegturbo module. For the sake of simplicity, I went with `jpeg`, If you want I can change it to `jpg`.

### DIFF
--- a/crates/kornia-io/src/jpeg.rs
+++ b/crates/kornia-io/src/jpeg.rs
@@ -68,30 +68,8 @@ pub fn read_image_jpeg_rgb8(file_path: impl AsRef<Path>) -> Result<Image<u8, 3>,
 /// # Returns
 ///
 /// A grayscale image with a single channel _(mono8)_.
-pub fn read_image_mono8(file_path: impl AsRef<Path>) -> Result<Image<u8, 1>, IoError> {
+pub fn read_image_jpeg_mono8(file_path: impl AsRef<Path>) -> Result<Image<u8, 1>, IoError> {
     read_image_jpeg_internal(file_path)
-}
-
-/// Reads a JPEG file with a single channel _(mono16)_.
-///
-/// # Arguments
-///
-/// - `file_path` - The path to the JPEG file.
-///
-/// # Returns
-///
-/// A grayscale image with a single channel _(mono16)_.
-pub fn read_image_mono16(file_path: impl AsRef<Path>) -> Result<Image<u16, 1>, IoError> {
-    let image: Image<u8, 1> = read_image_jpeg_internal(file_path)?;
-    let image_size = image.size();
-    let image_buf = image.as_slice();
-
-    let mut buf = Vec::with_capacity(image_buf.len() / 2);
-    for img in image_buf {
-        buf.push(*img as u16);
-    }
-
-    Ok(Image::new(image_size, buf)?)
 }
 
 fn read_image_jpeg_internal<const N: usize>(


crates/kornia-io/src/jpeg.rs
```diff
@@ -0,0 +1,202 @@
+use crate::error::IoError;
+use image::codecs::jpeg::{JpegDecoder, JpegEncoder};
```
edgarriba, line 2: can you use the raw `jpeg-decoder` crate ? https://github.com/image-rs/jpeg-decoder
the idea is to get rid of the top level image-rs crate

edgarriba (changes requested):
Good one!

crates/kornia-io/src/jpeg.rs
AS1100K: I moved the decoding functionality to use `jpeg-decoder` crate. Also, for the encoding functionality should I use something like [`jpeg-encoder`](https://crates.io/crates/jpeg-encoder) or just leave it to use `image` crate

crates/kornia-io/src/jpeg.rs
```diff
@@ -0,0 +1,201 @@
+use crate::error::IoError;
+use image::codecs::jpeg::JpegEncoder;
+use image::{ExtendedColorType, ImageEncoder};
```
edgarriba, line 3: what about the encoder? maybe this one https://github.com/vstroebel/jpeg-encoder

crates/kornia-io/src/jpeg.rs
```diff
@@ -0,0 +1,201 @@
...
+///
+/// # Arguments
+///
+/// - `file_path` - The path to the JPEG image.
+/// - `image` - The tensor containing the JPEG image data
+pub fn write_image_jpeg_rgb8(
+    file_path: impl AsRef<Path>,
+    image: &Image<u8, 3>,
+) -> Result<(), IoError> {
+    write_image_jpeg_internal(file_path, image, ExtendedColorType::Rgb8)
```
edgarriba, line 19: maybe we could have our simple `ColorType` or just `Encoding` (most robotics libraries use the latest as convention

crates/kornia-io/src/jpeg.rs
```diff
@@ -0,0 +1,201 @@
...
+
+    let image_size = ImageSize {
+        width: image_info.width as usize,
+        height: image_info.height as usize,
+    };
+
+    let img_data = decoder.decode()?;
+
+    // Convert the image color type, if necessary
+    if color_type != pixel_format {
```
edgarriba, line 135: I'm not sure about this. I would make the assumption for now that the user knows or has enough documentation to properly the function by passing the right image. At least we raise an error. In the worst case I believe the user can perform themselves a re-enconding with proper error handling. Let's keep the function simple

crates/kornia-io/src/jpeg.rs
AS1100K: Since, now we are only using color types in encoding, we are using `jpeg_encoder::ColorType` only and therefore, I believe that now have our color type wrapper won't be necessary

crates/kornia-io/Cargo.toml
```diff
@@ -17,6 +17,8 @@ all-features = true
 image = "0.25"
 kornia-image = { workspace = true }
 png = "0.17"
+jpeg-encoder = "0.6"
```
edgarriba, line 20: I was double checking the library docs and I just realized that there's no functionality to encode an image to raw bytes, only into files. Which is another missing feature required. Oh, maybe we should use zune-jpeg https://github.com/image-rs/image/blob/main/Cargo.toml#L79

AS1100K: `jpeg-encoder` supports writing the image to raw bytes.

```rust
// Example
let image_size = image.size();
let mut w = Vec::with_capacity(image_size.width * image_size.height);
let encoder = Encoder::new(&mut w, 100);
encoder.encode(
    image.as_slice(),
     image_size.width as u16,
     image_size.height as u16,
     color_type,
 )?;
```

Here the value of `w` will be updated, and it can be any type that implements [`JfifWrite`](https://docs.rs/jpeg-encoder/latest/jpeg_encoder/trait.JfifWrite.html)

Also, the `zune-jpeg` crate doesn't provide encoding support only decoding support and an high-level crate `zune-image` also uses `jpeg-encoder` behind the scenes for encoding https://docs.rs/zune-image/0.4.15/src/zune_image/codecs/jpeg.rs.html#152

edgarriba: alright ! good findings

edgarriba (approved):

edgarriba:
let's try to find out what's makig the ci to crash before merging

edgarriba:
@AS1100K take a look at this one too ;)
https://github.com/kornia/kornia-rs/pull/262

PR #330, opened 2025-04-01
Fix CI

AS1100K:
This PR attempts to fix the `python_test` workflow caused due to incorrect version of cmake available on venv

I have tested this on a [separate fork](https://github.com/as1100k-forks/kornia-rs-ci) and all the CI tests now passes

### DIFF
--- a/.github/workflows/python_test.yml
+++ b/.github/workflows/python_test.yml
@@ -34,10 +34,7 @@ jobs:
         profile: minimal
 
     - name: Install system dependencies
-      run: |
-        sudo apt-get install -y nasm libunwind-dev
-        sudo apt remove cmake -y
-        pip install cmake
+      run: sudo apt-get install -y nasm libunwind-dev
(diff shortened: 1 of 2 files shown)


.github/workflows/python_test.yml
```diff
@@ -34,7 +34,10 @@ jobs:
         profile: minimal
 
     - name: Install system dependencies
-      run: sudo apt-get install -y cmake nasm libunwind-dev
+      run: |
+        sudo apt-get install -y nasm libunwind-dev
+        sudo apt remove cmake -y
+        pip install cmake
```
ngoldbaum, line 40: You’re going to need to do this again for the free-threaded CI, which is duplicated below

AS1100K: Thanks! I forgot to change it in the other job

edgarriba (approved):

PR #332, opened 2025-04-02
Extend PNG encoding and decoding support

### LINKED ISSUE #252: implement `write_image_png_gray8`

### LINKED ISSUE #253: implement `write_image_png_rgb8`

AS1100K:
Fixes #252 
Fixes #253 

Also, I have added rgb16 and rgba16 support in both read and write functions. With this PR `png` module now supports every conversion.

I was thinking that `kornia_image::Image` should have a function for converting e.b. `Image<u16, 1>` to `Image<u8, 1>` and vice-versa, while maintain the same data using functions like [`u16::from_be_bytes`](https://doc.rust-lang.org/stable/std/primitive.u16.html#method.from_be_bytes)


crates/kornia-io/src/png.rs
```diff
@@ -98,10 +140,147 @@ fn read_png_impl(file_path: impl AsRef<Path>) -> Result<(Vec<u8>, [usize; 2]), I
...
+///
+/// # Arguments
+///
+/// - `file_path` - The path to the JPEG image.
+/// - `image` - The tensor containing the JPEG image data.
+pub fn write_image_png_gray8(
+    file_path: impl AsRef<Path>,
+    image: &Image<u8, 1>,
+) -> Result<(), IoError> {
+    write_png_impl(file_path, image, BitDepth::Eight, ColorType::Rgb)
```
edgarriba, line 179: Rgb ? :)

crates/kornia-io/src/png.rs
```diff
@@ -47,6 +47,48 @@ pub fn read_image_png_rgba8(file_path: impl AsRef<Path>) -> Result<Image<u8, 4>,
...
+/// Read a PNG image with a four channels (rgba16).
+///
+/// # Arguments
+///
+/// * `file_path` - The path to the PNG file.
+///
+/// # Returns
+///
+/// A RGB image with three channels (rgb8).
+pub fn read_image_png_rgba16(file_path: impl AsRef<Path>) -> Result<Image<u16, 4>, IoError> {
```
edgarriba, line 80: would be nice to test this one. Maybe load in test the rgb, convert to u16, save and reload ?

crates/kornia-io/src/png.rs
```diff
@@ -98,10 +140,147 @@ fn read_png_impl(file_path: impl AsRef<Path>) -> Result<(Vec<u8>, [usize; 2]), I
...
+        file_path,
+        &new_image,
+        BitDepth::Sixteen,
+        ColorType::Grayscale,
+    )
+}
+
+fn write_png_impl<const N: usize>(
+    file_path: impl AsRef<Path>,
+    image: &Image<u8, N>,
```
edgarriba, line 258: Do we have to pass u8 always ?

AS1100K: Yes, the `png` only supports `u8`. For using `u16`, we need break `u16` into `[u8; 2]` https://docs.rs/png/latest/png/struct.Writer.html#method.write_image_data

crates/kornia-io/src/png.rs
AS1100K: I added the test for rgb16. Instead of converting the u8 image to u16, I added a u16 image for the test. I did this because I thought that could be better for testing.

edgarriba (approved):

PR #270, opened 2025-03-15
Implemented decoding images directly from raw bytes . Fixes #121 

suhailnadaf509:
Here's a summary of what we've added:
Rust implementation:
Added decode_image_jpegturbo_rgb8 and decode_image_jpegturbo_gray8 functions to decode JPEG images from raw bytes using the turbojpeg backend
Added decode_image_bytes_rgb8 and decode_image_bytes_gray8 functions to decode any image format from raw bytes
Added unit tests for the new functions
Python bindings:
Added decode_image_bytes and decode_image_bytes_gray functions to decode any image format from raw bytes
Added decode_jpeg_bytes and decode_jpeg_bytes_gray functions to decode JPEG images from raw bytes using the turbojpeg backend
Registered these functions in the Python module
Examples:
Created a Rust example (decode_from_bytes.rs) demonstrating how to use the new functions
Created a Python example (decode_from_bytes.py) demonstrating how to use the new functions
Updated the README with documentation on the new functionality

Closes  #121

PR #336, opened 2025-04-03
Decode Images directly from raw bytes

### LINKED ISSUE #121: Decode images directly from raw bytes
Support both rust and python to decode directly the images coming in raw bytes.

Possibly specifying the encoder format

AS1100K:
Fixes #121 

This PR also improves error message of `IoError`

I created this PR even #270 exists because of [Discord Message](https://discord.com/channels/1295102166763049061/1319987694377635840/1355800511466700960), and also that PR uses some types of jpegturbo that doesn't exists

## Merge Conflicts
This PR will be conflicted once #334 merges, I will rebase and fix the conflicts once it is merged. _**Rebased**_

### DIFF
--- a/crates/kornia-image/src/error.rs
+++ b/crates/kornia-image/src/error.rs
@@ -14,7 +14,7 @@ pub enum ImageError {
     ImageDataNotContiguous,
 
     /// Error when the image shape is not valid.
-    #[error("Invalid image shape")]
+    #[error("Invalid image shape. {0}")]
     InvalidImageShape(#[from] kornia_tensor::TensorError),
 
     /// Error when the image size is not valid.
--- a/crates/kornia-image/src/error.rs
+++ b/crates/kornia-image/src/error.rs
@@ -14,7 +14,7 @@ pub enum ImageError {
     ImageDataNotContiguous,
 
     /// Error when the image shape is not valid.
-    #[error("Invalid image shape. {0}")]
+    #[error(transparent)]
     InvalidImageShape(#[from] kornia_tensor::TensorError),
 
     /// Error when the image size is not valid.
--- a/crates/kornia-io/Cargo.toml
+++ b/crates/kornia-io/Cargo.toml
@@ -22,7 +22,7 @@ kornia-image = { workspace = true }
 kornia-tensor = { workspace = true }
 png = "0.17"
 jpeg-encoder = "0.6"
-jpeg-decoder = "0.3"
+zune-jpeg = "0.4"
 log = { workspace = true }
 thiserror = { workspace = true }
 
--- a/crates/kornia-io/Cargo.toml
+++ b/crates/kornia-io/Cargo.toml
@@ -19,7 +19,6 @@ rustdoc-args = ["--cfg", "docsrs"]
 image = "0.25"
 circular-buffer = "1.1.0"
 kornia-image = { workspace = true }
-kornia-tensor = { workspace = true }
 png = "0.17"
 jpeg-encoder = "0.6"
 zune-jpeg = "0.4"
--- a/crates/kornia-io/Cargo.toml
+++ b/crates/kornia-io/Cargo.toml
@@ -19,6 +19,7 @@ rustdoc-args = ["--cfg", "docsrs"]
 image = "0.25"
 circular-buffer = "1.1.0"
 kornia-image = { workspace = true }
+kornia-tensor = { workspace = true }
 png = "0.17"
 jpeg-encoder = "0.6"
 jpeg-decoder = "0.3"
@@ -41,3 +42,4 @@ turbojpeg = ["dep:turbojpeg"]
 [[bench]]
 name = "bench_io"
 harness = false
+required-features = ["turbojpeg"]
--- a/crates/kornia-io/src/error.rs
+++ b/crates/kornia-io/src/error.rs
@@ -10,28 +10,28 @@ pub enum IoError {
     InvalidFileExtension(std::path::PathBuf),
 
     /// Error to open the file.
-    #[error("Failed to manipulate the file. {0}")]
+    #[error(transparent)]
     FileError(#[from] std::io::Error),
 
     /// Error to map the file to memory.
     #[cfg(feature = "turbojpeg")]
-    #[error("Error with Jpeg encoding/decoding. {0}")]
+    #[error(transparent)]
     JpegTurboError(#[from] crate::jpegturbo::JpegTurboError),
 
     /// Error to decode the JPEG image.
-    #[error("Error with Jpeg decoding. {0}")]
+    #[error(transparent)]
     JpegDecodingError(#[from] zune_jpeg::errors::DecodeErrors),
 
     /// Error to encode the JPEG image.
-    #[error("Error with Jpeg encoding. {0}")]
+    #[error(transparent)]
     JpegEncodingError(#[from] jpeg_encoder::EncodingError),
 
     /// Error to create the image.
-    #[error("Failed to create image. {0}")]
+    #[error(transparent)]
     ImageCreationError(#[from] kornia_image::ImageError),
 
     /// Error to decode the image.
-    #[error("Failed to decode the image. {0}")]
+    #[error(transparent)]
     ImageDecodeError(#[from] image::ImageError),
 
     /// Error to encode the PNG image.
--- a/crates/kornia-io/src/error.rs
+++ b/crates/kornia-io/src/error.rs
@@ -13,6 +13,16 @@ pub enum IoError {
     #[error(transparent)]
     FileError(#[from] std::io::Error),
 
+    /// Error to validate image resolution.
+    #[error(
+        "The Image Resolution didn't matched. Expected H: {0}, W: {1}, but found H: {2}, W: {3}"
+    )]
+    DecodeMismatchResolution(usize, usize, usize, usize),
+
+    /// Error when invalid buffer size slice is provided.
+    #[error("Expected buffer size to be {1}, but found {0}")]
+    InvalidBufferSize(usize, usize),
+
     /// Error to map the file to memory.
     #[cfg(feature = "turbojpeg")]
     #[error(transparent)]
--- a/crates/kornia-io/src/error.rs
+++ b/crates/kornia-io/src/error.rs
@@ -20,7 +20,7 @@ pub enum IoError {
 
     /// Error to decode the JPEG image.
     #[error("Error with Jpeg decoding")]
-    JpegDecodingError(#[from] jpeg_decoder::Error),
+    JpegDecodingError(#[from] zune_jpeg::errors::DecodeErrors),
 
     /// Error to encode the JPEG image.
     #[error("Error with Jpeg encoding")]
--- a/crates/kornia-io/src/error.rs
+++ b/crates/kornia-io/src/error.rs
@@ -10,35 +10,35 @@ pub enum IoError {
     InvalidFileExtension(std::path::PathBuf),
 
     /// Error to open the file.
-    #[error("Failed to manipulate the file")]
+    #[error("Failed to manipulate the file. {0}")]
     FileError(#[from] std::io::Error),
 
     /// Error to map the file to memory.
     #[cfg(feature = "turbojpeg")]
-    #[error("Error with Jpeg encoding/decoding")]
+    #[error("Error with Jpeg encoding/decoding. {0}")]
     JpegTurboError(#[from] crate::jpegturbo::JpegTurboError),
 
     /// Error to decode the JPEG image.
-    #[error("Error with Jpeg decoding")]
+    #[error("Error with Jpeg decoding. {0}")]
     JpegDecodingError(#[from] zune_jpeg::errors::DecodeErrors),
 
     /// Error to encode the JPEG image.
-    #[error("Error with Jpeg encoding")]
+    #[error("Error with Jpeg encoding. {0}")]
     JpegEncodingError(#[from] jpeg_encoder::EncodingError),
 
     /// Error to create the image.
-    #[error("Failed to create image")]
+    #[error("Failed to create image. {0}")]
     ImageCreationError(#[from] kornia_image::ImageError),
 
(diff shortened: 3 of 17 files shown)


crates/kornia-io/src/jpegturbo.rs
```diff
@@ -162,6 +162,27 @@ impl JpegTurboDecoder {
...
+    /// The decoded data as Image<u8, 1>.
+    pub fn decode_gray8(&mut self, jpeg_data: &[u8]) -> Result<Image<u8, 1>, JpegTurboError> {
+        self.decode(jpeg_data, turbojpeg::PixelFormat::GRAY)
+    }
+
+    fn decode<const N: usize>(
+        &mut self,
+        jpeg_data: &[u8],
+        format: turbojpeg::PixelFormat,
+    ) -> Result<Image<u8, N>, JpegTurboError> {
```
edgarriba, line 185: i think you should update also the `pixels` buffer with `vec![0u8; image_size.height * image_size.width * N];` (call it `C` to  keep consistency with the rest of the library?)

One more observation, is that if we should provide in the function interface a `mut img` instead so that allocation eg can happen only once (and is up to the user)

crates/kornia-io/src/png.rs
```diff
@@ -127,8 +178,13 @@ fn read_png_impl(file_path: impl AsRef<Path>) -> Result<(Vec<u8>, [usize; 2]), I
...
     }
 
-    let file = File::open(file_path)?;
-    let mut reader = Decoder::new(file)
+    let file = fs::read(file_path)?;
+    decode_png_imp(file.as_slice())
+}
+
+// Utility function to decode png files from raw bytes
+fn decode_png_imp(bytes: &[u8]) -> Result<(Vec<u8>, [usize; 2]), IoError> {
```
edgarriba, line 186: wondering if we should raise an error if expected type doesn't match

kornia-py/src/io/functional.rs
```diff
@@ -26,3 +27,43 @@ pub fn read_image_any(file_path: &str) -> PyResult<PyImage> {
...
+/// The following modes are supported:
+/// 1. "rgb" -> 8-bit RGB
+/// 2. "mono" -> 8-bit Monochrome
+///
+/// ```py
+/// import kornia_rs as K
+///
+/// img = K.decode_image_jpeg(bytes(img_data), "rgb")
+/// ```
+pub fn decode_image_jpeg(jpeg_data: &[u8], mode: &str) -> PyResult<PyImage> {
```
edgarriba, line 43: I would like to keep the default jpegturbo as it is but want to hear your opinion on wether we support raw jpeg in python (same applied to the ImageEncode/ImageDecoder api)

crates/kornia-io/src/png.rs
AS1100K: I was thinking that later, we could add a function like `decode_image` that will automatically going to figure out the file type, mode and that could internally use functions like `decode_image_png_rgba16`.

And for these functions, they could be used If the user know the file type and mode before hand or want the `Image` directly instead of a enum.

crates/kornia-io/src/jpegturbo.rs
AS1100K: By taking `&mut Image` as parameter, the user needs to ensure that the `Vec`'s capacity is not less than buffer size. The user needs to read the header information of the file, to get this info.

Also, for `Image<u16, _>` there is going to be a allocation twice. We could update the `Image` struct to always use `u8` tensor and have a phantom data that corresponds to `u8` or `u16`, and could have iterator implementation that does calculate `u16` on the fly. The benefit here is double allocation can be avoided and is up to the user, if they want to view the image data in `u8` _(Doesn't require additional allocation)_ or `u16`.

_This might be good change, but not sure if the breaking change is worth it and benefits kornia-rs in future._

I believe the current behaviour is doing single allocation, but creating a different tensor instead of modifying existing one. If you think that have `&mut Image` would be more beneficial to the user, I will be happy to update the code.

kornia-py/src/io/functional.rs
AS1100K: > I would like to keep the default jpegturbo
Could you please clarify it further?

**For supporting raw jpeg in python:**
Since we already have jpeg support in `kornia-io`, I think we should support it in python, because we are just re-exporting our `kornia-io` function as python function.

I personally think that kornia-py is wonderful but currently lacks support for many functions/modules like gstreamer in `kornia-io/stream`. Keeping everything up-to date for kornia-py with kornia-rs could be challenging but would allow python users to enjoy fast 3D computer vision library

crates/kornia-io/src/jpegturbo.rs
edgarriba: > If you think that have &mut Image would be more beneficial to the user, I will be happy to update the code.

there a lot of use cases in prod environments where you already know the fixed size of your images and you want to allocate only once as it's usually an expensive operation. We can use the info from the header to raise and error if the mut dst size doesn't align with the header info. It's very similar to the rest of the imgproc functions which I believe make a difference of kornia-rs api to other existing libraries

edgarriba:
@AS1100K  ready to rebase

AS1100K:
I switched from `jpeg_decoder` to `zune-jpeg` crate because `jpeg_decoder` doesn't provide a way to update the buffer while `zune-jpeg` provides a function for the same.

AS1100K:
The CI failed because of it couldn't download a dependency 🤔. Maybe try rerunning the CI after some time.
```
 Err:2 http://security.ubuntu.com/ubuntu noble-updates/main amd64 liblzma-dev amd64 5.6.1+really5.4.5-1build0.1
  404  Not Found [IP: 52.252.75.106 80]
E: Failed to fetch http://security.ubuntu.com/ubuntu/pool/main/x/xz-utils/liblzma-dev_5.6.1%2breally5.4.5-1build0.1_amd64.deb  404  Not Found [IP: 52.252.75.106 80]
Fetched 2365 kB in 1s (3171 kB/s)
E: Unable to fetch some archives, maybe run apt-get update or try with --fix-missing?
```

crates/kornia-io/src/jpeg.rs
```diff
@@ -69,10 +70,30 @@ pub fn read_image_jpeg_rgb8(file_path: impl AsRef<Path>) -> Result<Image<u8, 3>,
...
 }
 
-fn read_image_jpeg_internal<const N: usize>(
+/// Decodes a JPEG image with three channel (rgb8) from Raw Bytes.
+///
+/// # Arguments
+///
+/// - `image` - A mutable reference to your `Image`
+/// - `bytes` - Raw bytes of the jpeg file
+pub fn decode_image_jpeg_rgb8(image: &mut Image<u8, 3>, bytes: &[u8]) -> Result<(), IoError> {
```
edgarriba, line 82:
```suggestion
pub fn decode_image_jpeg_rgb8(src: &[u8], dst: &mut Image<u8, 3>) -> Result<(), IoError> {
```
to keep the same convention order like this, first immutable followed by mutables

crates/kornia-io/src/jpeg.rs
```diff
@@ -106,10 +127,31 @@ fn read_image_jpeg_internal<const N: usize>(
     Ok(Image::new(image_size, img_data)?)
 }
 
+fn decode_jpeg_impl<const C: usize>(image: &mut Image<u8, C>, bytes: &[u8]) -> Result<(), IoError> {
+    let mut decoder = zune_jpeg::JpegDecoder::new(bytes);
+    decoder
+        .decode_headers()
+        .map_err(IoError::JpegDecodingError)?;
```
edgarriba, line 134: above you don't map the error for the same call, right ?

crates/kornia-io/src/jpeg.rs
```diff
@@ -106,10 +127,31 @@ fn read_image_jpeg_internal<const N: usize>(
...
+        IoError::JpegDecodingError(zune_jpeg::errors::DecodeErrors::Format(String::from(
+            "Failed to find image info from its metadata",
+        )))
+    })?;
+
+    let image_size = [image_info.height as usize, image_info.width as usize, C];
+    decoder.decode_into(image.as_slice_mut())?;
+
+    // Update the tensor shape and stride
+    image.0.shape = image_size;
```
edgarriba, line 146: i don't think it's necessary if the image has been properly allocated. You should check sizes and return error before

crates/kornia-io/src/jpeg.rs
AS1100K: I removed this mapping too, because it is automatically converted by the from trait

edgarriba:
I just tried to re-run this right now and still stalled

AS1100K:
@edgarriba I updated the `kornia-py` code from the last time you visited. You may need review this PR again 🙂

crates/kornia-io/src/error.rs
```diff
@@ -10,35 +10,35 @@ pub enum IoError {
     InvalidFileExtension(std::path::PathBuf),
 
     /// Error to open the file.
-    #[error("Failed to manipulate the file")]
+    #[error("Failed to manipulate the file. {0}")]
```
edgarriba, line 13: all this errors we should derive from `transparent` so that we forward directly the original error messages unless we want to override some of them

crates/kornia-io/src/jpeg.rs
```diff
@@ -137,4 +183,21 @@ mod tests {
...
     }
+
+    #[test]
+    fn decode_jpeg() -> Result<(), IoError> {
+        // This is the size of buffer and must be known before hand
+        // for the sake of testing, we are keeping it a constant
+        const BUFFER_SIZE: usize = 150930;
+
+        let bytes = read("../../tests/data/dog.jpeg")?;
+        let mut image: Image<u8, 3> = Image::new([258, 195].into(), vec![0; BUFFER_SIZE])?;
```
edgarriba, line 194: or from_size_val([258, 195].into(), 0.0)

crates/kornia-image/src/error.rs
```diff
@@ -14,7 +14,7 @@ pub enum ImageError {
     ImageDataNotContiguous,
 
     /// Error when the image shape is not valid.
-    #[error("Invalid image shape")]
+    #[error("Invalid image shape. {0}")]
```
edgarriba, line 17:
```suggestion
    #[error(transparent)]
```

crates/kornia-io/Cargo.toml
```diff
@@ -19,9 +19,10 @@ rustdoc-args = ["--cfg", "docsrs"]
 image = "0.25"
 circular-buffer = "1.1.0"
 kornia-image = { workspace = true }
+kornia-tensor = { workspace = true }
```
edgarriba, line 22: why do we need this dependency here ?

crates/kornia-io/src/jpeg.rs
```diff
@@ -106,10 +126,36 @@ fn read_image_jpeg_internal<const N: usize>(
...
+
+    let image_info = decoder.info().ok_or_else(|| {
+        IoError::JpegDecodingError(zune_jpeg::errors::DecodeErrors::Format(String::from(
+            "Failed to find image info from its metadata",
+        )))
+    })?;
+
+    if [image_info.height as usize, image_info.width as usize] != [dst.height(), dst.width()] {
+        return Err(IoError::JpegDecodingError(
+            zune_jpeg::errors::DecodeErrors::Format(format!(
```
edgarriba, line 141: i don't think it's an error from the decoder. We could have a dedicated error `IoError::DecodeMismatchResolution` or maybe reuse from ImageError

crates/kornia-io/src/png.rs
```diff
@@ -140,6 +199,40 @@ fn read_png_impl(file_path: impl AsRef<Path>) -> Result<(Vec<u8>, [usize; 2]), I
...
+    image_size: ImageSize,
+) -> Result<(), IoError> {
+    let mut reader = Decoder::new(src)
+        .read_info()
+        .map_err(|e| IoError::PngDecodeError(e.to_string()))?;
+
+    let image_info = reader.info();
+    if image_info.size() != (image_size.width as u32, image_size.height as u32) {
+        return Err(IoError::PngDecodeError(format!(
+            "The Image shape didn't matched. Expected H: {}, W: {}, but found H: {}, W: {}",
```
edgarriba, line 215: same as above

crates/kornia-io/src/png.rs
```diff
@@ -140,6 +199,40 @@ fn read_png_impl(file_path: impl AsRef<Path>) -> Result<(Vec<u8>, [usize; 2]), I
...
+            image_info.height, image_info.width, image_size.height, image_size.width
+        )));
+    }
+
+    if dst.len() < reader.output_buffer_size() {
+        return Err(IoError::PngDecodeError(format!(
```
edgarriba, line 221: this is not a decoding error at all. Same as before, maybe have a dedicated error

crates/kornia-io/src/png.rs
```diff
@@ -352,4 +430,21 @@ mod tests {
...
     }
+
+    #[test]
+    fn decode_png() -> Result<(), IoError> {
+        // This is the size of buffer and must be known before hand
+        // for the sake of testing, we are keeping it a constant
+        const BUFFER_SIZE: usize = 150930;
+
+        let bytes = read("../../tests/data/dog-rgb8.png")?;
+        let mut image: Image<u8, 3> = Image::new([258, 195].into(), vec![0; BUFFER_SIZE])?;
```
edgarriba, line 441: from size val

edgarriba (approved):
LGTM. Just minimum changes

crates/kornia-io/src/jpeg.rs
AS1100K: I went with `DecodeMismatchResolution`

crates/kornia-io/Cargo.toml
AS1100K: I must have probably forgot to remove it when I was trying something with direct tensors. I removed it

edgarriba (approved):

PR #338, opened 2025-04-03
Implement Zero copy images from gstreamer

### LINKED ISSUE #331: Zero copy images from gstreamer
Right now images coming from gstreamer are not zero copy.

A couple of solutions described here: https://discourse.gstreamer.org/t/zero-copy-video-frames/3856/2

AS1100K:
Fixes #331 

I followed the approach suggested by slomo at https://discourse.gstreamer.org/t/zero-copy-video-frames/3856/2

## Benchmarks
I have added `gstreamer.rs` benchmark, and this PR improves performance by <ins> **_~9.84%_** </ins> on my local machine.

```
Video downloaded to: "/tmp/.tmpJTJ1BC/video.mp4"
grab                    time:   [3.9711 ns 3.9861 ns 4.0036 ns]
                        change: [-9.8473% -9.3685% -8.8909%] (p = 0.00 < 0.05)
                        Performance has improved.
```

Also, I have used a video from [kornia/tutorials](https://github.com/kornia/tutorials) repository for the benchmark.

### DIFF
--- a/crates/kornia-io/src/stream/capture.rs
+++ b/crates/kornia-io/src/stream/capture.rs
@@ -1,7 +1,8 @@
 use crate::stream::error::StreamCaptureError;
 use circular_buffer::CircularBuffer;
 use gstreamer::prelude::*;
-use kornia_image::{Image, ImageSize};
+use kornia_image::Image;
+use kornia_tensor::{CpuAllocator, Tensor};
 use std::sync::{Arc, Mutex};
 
 // utility struct to store the frame buffer
@@ -90,20 +91,19 @@ impl StreamCapture {
             .lock()
             .map_err(|_| StreamCaptureError::MutexPoisonError)?;
         if let Some(frame_buffer) = circular_buffer.pop_front() {
-            // TODO: solve the zero copy issue
-            // https://discourse.gstreamer.org/t/zero-copy-video-frames/3856/2
             let buffer = frame_buffer
                 .buffer
                 .map_readable()
                 .map_err(|_| StreamCaptureError::GetBufferError)?;
-            let img = Image::<u8, 3>::new(
-                ImageSize {
-                    width: frame_buffer.width as usize,
-                    height: frame_buffer.height as usize,
-                },
-                buffer.to_owned(),
+
+            let tensor = Tensor::from_shape_slice(
+                [frame_buffer.width as usize, frame_buffer.height as usize, 3],
+                buffer.as_slice(),
+                CpuAllocator,
             )
             .map_err(|_| StreamCaptureError::CreateImageFrameError)?;
+
+            let img = Image::<u8, 3>(tensor);
             return Ok(Some(img));
         }
         Ok(None)
--- a/crates/kornia-io/Cargo.toml
+++ b/crates/kornia-io/Cargo.toml
@@ -33,6 +33,7 @@ turbojpeg = { version = "1.2", optional = true }
 [dev-dependencies]
 criterion = { workspace = true }
 tempfile = { workspace = true }
+reqwest = { version = "0.12", features = ["blocking"] }
 
 [features]
 gstreamer = ["dep:gstreamer", "dep:gstreamer-app"]
@@ -41,3 +42,9 @@ turbojpeg = ["dep:turbojpeg"]
 [[bench]]
 name = "bench_io"
 harness = false
+required-features = ["turbojpeg"]
+
+[[bench]]
+name = "gstreamer"
+harness = false
+required-features = ["gstreamer"]
--- a/crates/kornia-io/Cargo.toml
+++ b/crates/kornia-io/Cargo.toml
@@ -19,6 +19,7 @@ rustdoc-args = ["--cfg", "docsrs"]
 image = "0.25"
 circular-buffer = "1.1.0"
 kornia-image = { workspace = true }
+kornia-tensor = { workspace = true }
 png = "0.17"
 jpeg-encoder = "0.6"
 jpeg-decoder = "0.3"
--- a/crates/kornia-io/benches/gstreamer.rs
+++ b/crates/kornia-io/benches/gstreamer.rs
@@ -1,0 +1,52 @@
+use criterion::{black_box, Criterion};
+use reqwest::blocking::get;
+use std::io::copy;
+use std::path::PathBuf;
+use std::{fs::File, time::Duration};
+use tempfile::{tempdir, TempDir};
+
+const FILE_NAME: &str = "video.mp4";
+const VIDEO_LINK: &str =
+    "https://github.com/kornia/tutorials/raw/refs/heads/master/data/sharpening.mp4";
+
+fn download_video<'a>() -> (PathBuf, TempDir) {
+    let response = get(VIDEO_LINK).expect("Failed to download video");
+    let temp_dir = tempdir().expect("Failed to create temp directory");
+    let temp_file_path = temp_dir.path().join(FILE_NAME);
+    let mut temp_file = File::create(&temp_file_path).expect("Failed to create temp file");
+
+    copy(
+        &mut response.bytes().expect("Failed to read response").as_ref(),
+        &mut temp_file,
+    )
+    .expect("Failed to write video to temp file");
+
+    println!("Video downloaded to: {:?}", temp_file_path);
+    (temp_file_path, temp_dir)
+}
+
+fn benchmark_get_buffer(c: &mut Criterion) {
+    let (video_path, _temp_dir) = download_video();
+
+    let pipeline_desc = format!(
+        "filesrc location={} ! decodebin ! videoconvert ! video/x-raw,format=RGB,width=1024,height=688,framerate=8/1 ! appsink name=sink sync=false",
+        video_path.to_str().unwrap()
+    );
+
+    let mut stream_capture = kornia_io::stream::StreamCapture::new(&pipeline_desc).unwrap();
+    stream_capture.start().unwrap();
+
+    c.bench_function("grab", |b| {
+        b.iter(|| {
+            black_box(stream_capture.grab()).expect("Failed to grab the image");
+        });
+    });
+}
+
+criterion::criterion_group! {
+    name = benches;
+    config = Criterion::default().warm_up_time(Duration::from_secs(1));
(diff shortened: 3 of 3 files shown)


AS1100K:
The CI failed because of it couldn't find OpenSSL. Strange!

```
cargo:warning=Could not find directory of OpenSSL installation, and this `-sys` crate cannot proceed without this knowledge. If OpenSSL is installed and this crate had trouble finding it,  you can set the `OPENSSL_DIR` environment variable for the compilation process. See stderr section below for further information.
```

**FIXED**

crates/kornia-io/src/stream/capture.rs
```diff
@@ -90,20 +91,19 @@ impl StreamCapture {
...
                 .map_readable()
                 .map_err(|_| StreamCaptureError::GetBufferError)?;
-            let img = Image::<u8, 3>::new(
-                ImageSize {
-                    width: frame_buffer.width as usize,
-                    height: frame_buffer.height as usize,
-                },
-                buffer.to_owned(),
+
+            let tensor = Tensor::from_shape_slice(
```
edgarriba, line 99: I don't think it's as simple as this function calls `to_vec` internally. There are two options:
1) Restructure TensorStorage to active a generic for the image that implements `AsRef<u8>` (not sure the consequences os that at all
2) Use `Tensor::from_raw_parts` along with a custom `GstreamerAllocator` (I tried that at some point) which also came with other caveats

let me know if you have any other suggestions

crates/kornia-tensor/Cargo.toml
```diff
@@ -16,6 +16,7 @@ num-traits = { workspace = true }
 serde = { workspace = true, optional = true }
 bincode = { workspace = true, optional = true }
 thiserror = { workspace = true }
+auto_impl = "1.2"
```
edgarriba, line 19: i'd prefer if we don't include any extra dependency. This crate should be really light

AS1100K: removed `auto_impl`

edgarriba:
if you want to check, this is what i attempted some months ago in this direction https://github.com/kornia/kornia-rs/commit/7aac9c8594a52baf6ebbaa855982ff872335497f

crates/kornia-io/src/stream/capture.rs
```diff
@@ -90,21 +106,44 @@ impl StreamCapture {
...
+            let layout = unsafe { Layout::array::<u8>(frame_data_slice.len()).unwrap_unchecked() };
+
+            let length = frame_data_slice.len();
+            let shape = [height, width, 3];
+            let strides = get_strides_from_shape(shape);
+
+            // Drop the buffer_map as it is a reference of Buffer
+            drop(buffer_map);
+
+            let gst_parent_deallocator = GstParentDeallocator(frame_buffer.buffer);
```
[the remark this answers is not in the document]
AS1100K: I took some inspiration from it and made the `ParentDeallocator` logic in `CpuAllocator` because it depends a lot on the `Allocator` than `TensorStorage`. What do you say about this direction?

crates/kornia-tensor/src/allocator.rs
```diff
@@ -68,15 +88,20 @@ impl TensorAllocator for CpuAllocator {
     /// # Arguments
     ///
     /// * `ptr` - A non-null pointer to the allocated memory.
+    /// * `parent` - An optional parent deallocator.
```
gbin, line 91: you don't pass parent here

crates/kornia-tensor/src/allocator.rs
```diff
@@ -30,17 +31,36 @@ pub trait TensorAllocator: Clone {
...
+    /// * `ptr` - A non-null pointer to the allocated memory.
+    /// * `layout` - The layout of the tensor.
     fn dealloc(&self, ptr: *mut u8, layout: Layout);
 }
 
 #[derive(Clone)]
 /// A tensor allocator that uses the system allocator.
-pub struct CpuAllocator;
+pub struct CpuAllocator {
+    parent: Option<Arc<dyn ParentDeallocator>>,
```
gbin, line 45: Do we have benchmarks before/after as this adds: 1 branch to get the option, 1 sync with Arc, 1 virtual table access?

crates/kornia-tensor/src/parent_deallocator.rs
```diff
@@ -0,0 +1,11 @@
+use std::panic::RefUnwindSafe;
+
+/// A trait for deallocating memory of parent for tensor.
+///
+/// # Safety
+///
+/// The parent deallocator must be thread-safe.
+pub trait ParentDeallocator: Send + Sync + RefUnwindSafe {
```
gbin, line 8: Maybe using a generic instead of a virtual table would be better?

crates/kornia-tensor/src/storage.rs
```diff
@@ -340,12 +340,63 @@ mod tests {
...
         Ok(())
     }
+
+    #[test]
+    fn test_parent_deallocator() -> Result<(), TensorAllocatorError> {
+        use std::sync::atomic::{AtomicBool, Ordering};
+        use std::sync::Arc;
+
+        #[derive(Clone)]
+        struct TestParentDeallocator {
```
gbin, line 358: maybe use mockall or something similar, there are mock test frameworks that can trace calls like this

crates/kornia-tensor/src/storage.rs
```diff
@@ -166,13 +166,13 @@ mod tests {
...
         let size = 8;
-        let allocator = CpuAllocator;
+        let allocator = CpuAllocator::default();
         let layout = Layout::array::<u8>(size).map_err(TensorAllocatorError::LayoutError)?;
         let ptr =
             NonNull::new(allocator.alloc(layout)?).ok_or(TensorAllocatorError::NullPointer)?;
         let ptr_raw = ptr.as_ptr();
 
-        let buffer = TensorStorage {
+        let buffer: TensorStorage<_, _> = TensorStorage {
```
gbin, line 175: IMHO, it doesn't add much to add the type here, it is given 2 chrs later

crates/kornia-io/src/stream/capture.rs
```diff
@@ -90,21 +103,41 @@ impl StreamCapture {
...
+
+            let frame_data_slice = buffer_map.as_slice();
+            let frame_data_ptr = frame_data_slice.as_ptr();
+
+            let length = frame_data_slice.len();
+            let shape = [height, width, 3];
+            let strides = get_strides_from_shape(shape);
+
+            // Drop the buffer_map as it is a reference of Buffer
+            drop(buffer_map);
```
gbin, line 123: This is completely broken.

once you do that, frame_data_ptr becomes dangling.

crates/kornia-io/src/stream/capture.rs
```diff
@@ -90,21 +103,41 @@ impl StreamCapture {
...
+
+            let gst_parent_deallocator = Arc::new(GstParentDeallocator(frame_buffer.buffer));
+
+            let tensor = unsafe {
+                Tensor {
+                    shape,
+                    strides,
+                    storage: TensorStorage::from_raw_parts(
+                        frame_data_ptr,
```
gbin, line 132: then you tell Rust to ignore the dangling pointer and use it as a valid allocated memory.

gbin (changes requested):
Key part: you are releasing a pooled buffer and using its underlying  pointer after that -> there is no guarantees about what this pointer will contain by the time of use.

crates/kornia-io/src/stream/capture.rs
```diff
@@ -90,21 +103,41 @@ impl StreamCapture {
...
+                        length,
+                        CpuAllocator::with_parent_relation(gst_parent_deallocator),
```
edgarriba, line 134: Can we find a way that you just pass directly the GstAllocator ?

crates/kornia-tensor/src/allocator.rs
edgarriba: I feel adding the parent in this struct it’s unnecessary. The allocators are generics in the tensor storage which you can specialise (eg adding the gst buffer or anything) as soon as you implement the trait methods. I’d explore going to that route

crates/kornia-tensor/src/parent_deallocator.rs
edgarriba: Allocators are generics already in the tensor storage structure, if that’s what you meant

crates/kornia-io/src/stream/capture.rs
gbin: I believe the correct architecture is to keep this buffer handle as is and on the Kornia side wrap it in a generic warper for all types of memory.

edgarriba: @gbin how would the api look like in this case at a high level ?

crates/kornia-io/src/stream/mod.rs
```diff
@@ -22,3 +22,39 @@ pub use crate::stream::error::StreamCaptureError;
...
 pub use crate::stream::v4l2::V4L2CameraConfig;
 pub use crate::stream::video::VideoWriter;
+
+use kornia_image::allocator::ImageAllocator;
+use kornia_tensor::{allocator::TensorAllocatorError, TensorAllocator};
+
+#[allow(dead_code)]
+#[derive(Clone)]
+/// A [TensorAllocator] used for those images, whose memory is managed by gstreamer.
+pub struct GstAllocator(gstreamer::Buffer);
```
AS1100K, line 32: @edgarriba Should we have `Option<gstreamer::Buffer>` as it would simplify the `Default` as it is used in copper example?

crates/kornia-io/src/stream/capture.rs
gbin: For example on Copper we hold the actual Gst handle which keeps it out of the recycling on the backend until you naturally release it.
There is no secret here I believe, if this is RAII from gstreamer and you need a 0 copy then hold the original surface until your don't need it anymore, then release it.

AS1100K: already did it after merging #373 and updating this PR to be up-to date

AS1100K:
Continued and Merged in #387

PR #339, opened 2025-04-04
Python: make encoders and decoders frozen pyclasses

ngoldbaum:
Note this is targeted at the PR branch for #334.

Since you're using interior mutability via a `Mutex`, all this `mut` isn't necessary.


edgarriba:
thanks !

PR #343, opened 2025-04-05
Fix FAST implementations + NMS

kylekam:
## This PR contains
- a fix for the previous FAST implementation (see images below)
- an implementation of NMS using scoring from Sum of Absolute Differences

Goal is to augment FAST implementation that @edgarriba did here: https://github.com/kornia/kornia-rs/pull/240

## Future work:
- Use SIMD to optimize `get_fast_corner_score`
- Make NMS generic enough to be used for FAST and Harris corner
- Implement other scoring functions for NMS

This is the FAST response prior to this PR. Note the lack of corners along the ridge of the yellow house and the top corners of the windows.
before_fix

FAST response with this PR. No NMS in this image.
fix_no_nms

FAST + NMS.
fix_with_nms

### DIFF
--- a/crates/kornia-imgproc/src/features/fast.rs
+++ b/crates/kornia-imgproc/src/features/fast.rs
@@ -1,5 +1,27 @@
 use kornia_image::{Image, ImageError};
 use rayon::prelude::*;
+use std::collections::BinaryHeap;
+use std::cmp::Ordering;
+
+// Structure to represent a feature point with its score and coordinates. Useful for NMS.
+#[derive(Copy, Clone, Eq, PartialEq)]
+struct FeaturePoint {
+    score: i32,
+    x: usize,
+    y: usize,
+}
+
+impl Ord for FeaturePoint {
+    fn cmp(&self, other: &Self) -> Ordering {
+        self.score.cmp(&other.score)
+    }
+}
+
+impl PartialOrd for FeaturePoint {
+    fn partial_cmp(&self, other: &Self) -> Option<Ordering> {
+        Some(self.cmp(other))
+    }
+}
 
 /// Fast feature detector
 ///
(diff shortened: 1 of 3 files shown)


edgarriba:
Yep, let’s figure out the timings. Check that vpi for same image with cpu in a jetson (factor 2-3x slower than regular laptop) they report ~2ms https://docs.nvidia.com/vpi/algo_fast_corners_detector.html#algo_fast_corners_detector_overview

crates/kornia-imgproc/src/features/fast.rs
```diff
@@ -133,6 +155,236 @@ fn is_fast_corner(
...
+    let p8 = get_pixel_from_offset(7);
+    let p10 = get_pixel_from_offset(9);
+    let p11 = get_pixel_from_offset(10);
+    let p12 = get_pixel_from_offset(11);
+    let p14 = get_pixel_from_offset(13);
+    let p15 = get_pixel_from_offset(14);
+    let p16 = get_pixel_from_offset(15);
+    let pixels = [
+        p1, p2, p3, p4, p5, p6, p7, p8, p9, p10, p11, p12, p13, p14, p15, p16,
+        p1, p2, p3, p4, p5, p6, p7, p8, p9, p10, p11, p12, p13, p14, p15, p16
```
edgarriba, line 211: Why twice ? That might explain the big difference

crates/kornia-imgproc/src/features/fast.rs
```diff
@@ -133,6 +155,236 @@ fn is_fast_corner(
...
+        } else {
+            update_indices(&consecutive_brighter, &idx, &mut start_idx, &mut end_idx, &mut most_consecutive);
+            update_indices(&consecutive_darker, &idx, &mut start_idx, &mut end_idx, &mut most_consecutive);
+
+            consecutive_brighter = 0;
+            consecutive_darker = 0;
+        }
+
+        if consecutive_brighter >= arc_length || consecutive_darker >= arc_length {
+            is_corner = true;
```
edgarriba, line 253: Why don’t you break here already ?

crates/kornia-imgproc/src/features/fast.rs
```diff
@@ -133,6 +155,236 @@ fn is_fast_corner(
...
+        p1, p2, p3, p4, p5, p6, p7, p8, p9, p10, p11, p12, p13, p14, p15, p16,
+        p1, p2, p3, p4, p5, p6, p7, p8, p9, p10, p11, p12, p13, p14, p15, p16 
+    ]; // Values are repeated to handle circular indexing
+    
+    let mut consecutive_brighter = 0u8;
+    let mut consecutive_darker = 0u8;
+
+    let mut is_corner = false;
+    let mut start_idx = 0;
+    let mut end_idx = 0;
```
edgarriba, line 219: This is not updated right ?

crates/kornia-imgproc/src/features/fast.rs
```diff
@@ -133,6 +155,236 @@ fn is_fast_corner(
...
+pub fn fast_feature_nms_detector(
+    src: &Image<u8, 1>,
+    threshold: u8,
+    arc_length: u8,
+    nms: bool
+) -> Result<Vec<[i32; 2]>, ImageError> {
+    let (cols, rows) = (src.cols() as i32, src.rows() as i32);
+
+    // Precompute the offsets for the Bresenham circle
+    let offsets = [
```
edgarriba, line 300: One micro optimisation here could be to reduce half this array and invert just values

crates/kornia-imgproc/src/features/fast.rs
```diff
@@ -133,6 +155,236 @@ fn is_fast_corner(
...
+/// * `nms` - A boolean flag to enable or disable Non-Maximum Suppression.
+/// # Returns
+///
+/// A vector containing the coordinates of the detected keypoints.
+pub fn fast_feature_nms_detector(
+    src: &Image<u8, 1>,
+    threshold: u8,
+    arc_length: u8,
+    nms: bool
+) -> Result<Vec<[i32; 2]>, ImageError> {
```
edgarriba, line 296: I don’t think we are handling any error, so we might return directly the vec

crates/kornia-imgproc/src/features/fast.rs
```diff
@@ -133,6 +155,236 @@ fn is_fast_corner(
...
+        -2 * cols - 2, // 15
+        -3 * cols - 1,     // 16
+    ];
+
+    // Process rows in parallel
+    let (kp1, scores): (Vec<[i32; 2]>, Vec<i32>) = (3..rows - 3)
+        .into_par_iter()
+        .flat_map(|y| {
+            let row_start_idx = y * cols;
+            let mut row_keypoints = Vec::new();
```
edgarriba, line 324: Why the vec inside ? I would put outside and append each time we get a valid point

crates/kornia-imgproc/src/features/fast.rs
```diff
@@ -133,6 +155,236 @@ fn is_fast_corner(
...
+            let mut kp_scores = Vec::new();
+
+            for x in 3..cols - 3 {
+                let (is_corner, score) = get_fast_corner_score(
```
edgarriba, line 328:
```suggestion
                If let Some(score) = get_fast_corner_score(
```
Better use an optional and why not appending directly the FeaturePoint ?

crates/kornia-imgproc/src/features/fast.rs
```diff
@@ -133,6 +155,236 @@ fn is_fast_corner(
...
+        }
+    }
+
+    if !is_corner {
+        return (false, 0); // Not a corner found
+    }
+
+    // Sum of absolute differences for the corner score. 
+    let mut score = 0i32;
+    for offset in start_idx..start_idx + most_consecutive as usize {
```
edgarriba, line 271: This probably can be done while check the arc ? To avoid this extra loop

tests/data/houses.png
edgarriba, line 1: We should probably host this in GitHub and use reqwest to download. I’ll take care later of it

crates/kornia-imgproc/src/features/fast.rs
```diff
@@ -133,6 +155,236 @@ fn is_fast_corner(
...
+    let get_pixel_from_offset =
+        |off_idx: usize| unsafe { src.get_unchecked((pixel_idx + offsets[off_idx]) as usize) };
+
+    // Fast rejection test - check if at least 3 of the 4 high-speed test points are different enough
+    let p1 = get_pixel_from_offset(0);
+    let p5 = get_pixel_from_offset(4);
+    let p9 = get_pixel_from_offset(8);
+    let p13 = get_pixel_from_offset(12);
+
+    let m0 = (p1 >= lower_threshold) as u8 +
```
edgarriba, line 182: In the reference implementation of OpenCV they were checking with Boolean masks. Maybe the casting is adding overhead

kylekam:
> Yep, let’s figure out the timings. Check that vpi for same image with cpu in a jetson (factor 2-3x slower than regular laptop) they report ~2ms https://docs.nvidia.com/vpi/algo_fast_corners_detector.html#algo_fast_corners_detector_overview

I'll look into it right now.

crates/kornia-imgproc/src/features/fast.rs
kylekam: Some NMS functions (such as sum of absolute differences) requires knowledge of the entire arc. If we exit early here, we won't have that information.

crates/kornia-imgproc/src/features/fast.rs
kylekam: This is to account for the discontinuity from index 16->1 in the Bressenham circle. If we have a valid arc that starts continues through this section, the previous algorithm does not count it as valid. Hence, why many corners in the previous algorithm are missing. 

I agree this current method is unoptimized.

crates/kornia-imgproc/src/features/fast.rs
kylekam: All three are used to track the largest segment which is used calculate the NMS score for the center pixel.

crates/kornia-imgproc/src/features/fast.rs
kylekam: Cannot have multiple mutable references to because of rayon.

crates/kornia-imgproc/src/features/fast.rs
kylekam: Agreed.

kylekam:
> > Yep, let’s figure out the timings. Check that vpi for same image with cpu in a jetson (factor 2-3x slower than regular laptop) they report ~2ms https://docs.nvidia.com/vpi/algo_fast_corners_detector.html#algo_fast_corners_detector_overview
> 
> I'll look into it right now.

Hmm...VPI source code isn't open sourced. However, I found this C++ repo which seems pretty good: https://github.com/komrad36/KFAST

kylekam:
@edgarriba Based on exploring some other repos with optimized implementations of FAST and VPI, I plan on reworking the current implementation with manual AVX2 instructions. This may be a more involved effort, do you want this work to be captured in a separate PR or keep it in this one? The current implementation still returns the correct results.

kylekam:
I was able to get some improvements by refactoring parts of the `get_fast_corner_score` function and setting AVX2 as the compile target. There might be more improvements available in the cost function, but I think long term it would be ideal to use intrinsics.

image

crates/kornia-imgproc/src/features/fast.rs
kylekam: I modified it use the the first segment detected with arc_length. Less robust, but quicker execution time.

tests/data/houses.png
edgarriba, line 1: can we find a way to download the image from here ? https://github.com/kornia/data/blob/main/houses.png
want to avoid adding large blobs in the repo

crates/kornia-imgproc/src/features/fast.rs
```diff
@@ -133,6 +155,266 @@ fn is_fast_corner(
...
+    pixel_idx: i32,
+    offsets: [i32; 16],
+    threshold: u8,
+    arc_length: u8,
+) -> (bool, i32) {
+    get_fast_corner_score_inner(src, pixel_idx, offsets, threshold, arc_length)
+}
+
+/// Wrapper function for FAST corner score calculation.
+pub fn get_fast_corner_score(
```
edgarriba, line 270: should we make this private ?

edgarriba (commented):
sorry for so late reply; the implementation looks promising. What else is missing ?

PR #344, opened 2025-04-06
pref: solved TODO: improve performance by using matmul33

dhruvkjain:
# TODO: improve performance by using matmul33
At `crates/kornia-icp/src/ops.rs` inside function `fit_transformation` at [line 39](https://github.com/kornia/kornia-rs/blob/6ec2cca5285a1dcc5acd62c55882591c4eca8d6c/crates/kornia-icp/src/ops.rs#L39)

Currently, the ICP implementation uses `faer::linalg::matmul::matmul` for multiplying 3×3 matrices. While `faer` is optimized for general-purpose matrix operations, it's overkill for small fixed-size matrices like 3×3 and introduces unnecessary overhead.

So TODO was to replace this general purpose `matmul` with custom implementation 
Therefore, I added a function `matmul_faer_mat` in `kornia_3d::linalg` that takes 3x3 matrices of type `faer::Mat` and return an mutable pre-allocated matrix that is multiplication of both matrices.

## Implementation pattern:
```rust
 m.write(0, 0, a.read(0, 0) * b.read(0, 0) + a.read(0, 1) * b.read(1, 0) + a.read(0, 2) * b.read(2, 0));
```

## Results: 
Duration it took using my implementation i.e. `matmul_faer_mat`:
```console
233ns
123ns
496ns
35ns
351ns
48ns
34ns
```

Duration it took using `faer::linalg::matmul::matmul`:
```console
353ns
371ns
570ns
193ns
179ns
171ns
180ns
```

## My implementation took **~100ns** less than `faer::linalg::matmul::matmul`

### DIFF
--- a/crates/kornia-3d/src/linalg.rs
+++ b/crates/kornia-3d/src/linalg.rs
@@ -102,6 +102,45 @@ pub fn matmul33(a: &[[f64; 3]; 3], b: &[[f64; 3]; 3], m: &mut [[f64; 3]; 3]) {
     m[2][2] = a[2][0] * b[0][2] + a[2][1] * b[1][2] + a[2][2] * b[2][2];
 }
 
+use faer::{Mat, MatRef};
+/// Multiply two 3x3 faer::Mat matrices
+/// 
+/// # Arguments
+/// 
+/// * `a` - The left hand side 3x3 faer::Mat matrix.
+/// * `b` - The right hand side 3x3 faer::Mat matrix.
+/// * `m` - A pre-allocated 3x3 matrix to store the faer::Mat result.
+/// 
+/// PRECONDITION: m is a pre-allocated 3x3 faer::Mat matrix
+/// 
+/// # Example
+/// 
+/// ```
+/// use kornia_3d::linalg::matmul33_faer_mat;
+///
+/// let a_array = [[1.0, 2.0, 3.0], [4.0, 5.0, 6.0], [7.0, 8.0, 9.0]];
+/// let b_array = [[1.0, 2.0, 3.0], [4.0, 5.0, 6.0], [7.0, 8.0, 9.0]];
+/// let m_array = [[0.0; 3]; 3];
+/// 
+/// let a = Mat::from_fn(3, 3, |i, j| a_array[i][j]);
+/// let b = Mat::from_fn(3, 3, |i, j| b_array[i][j]);
+/// let mut m = Mat::from_fn(3, 3, |i, j| m_array[i][j]);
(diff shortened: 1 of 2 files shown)


edgarriba (changes requested):
ideally we use `glam-rs` for this. How did you get this timings ?

dhruvkjain:
> ideally we use `glam-rs` for this. How did you get this timings ?

I used `let now = std::time::Instant::now();`, because I saw a similar approach used at `kornia-rs/crates/kornia-icp/src/icp_vanilla.rs` [line 79](https://github.com/kornia/kornia-rs/blob/6ec2cca5285a1dcc5acd62c55882591c4eca8d6c/crates/kornia-icp/src/icp_vanilla.rs#L79), with different types of matrices (Rank-deficient, symmetric, skew symmetric etc.) as inputs.

About `glam-rs`, 
I read it's docs and found `DMat3` matrices multiplication: multiply two 3x3 DMat3 type matrices of f64 values using [`*` operator](https://docs.rs/glam/latest/glam/f64/struct.DMat3.html#method.mul-4) or [`mat_mul3` function](https://docs.rs/glam/latest/glam/f64/struct.DMat3.html#method.mul_mat3).

I feel like adding glam would increase a dependency and would require us to handle new `DMat3` type matrix, my implementation seems pretty straightforward.
I'm not very familiar with `glam-rs`, so I'd appreciate it if you could explain why `glam-rs` might be a better option.

dhruvkjain:
Hey @edgarriba, haven't heard back from you on this.

edgarriba:
glam is a very light dependency with nice and clean abstractions for small matrices operators including optimizations which we'll add in any case as dependency in the 3d crate

edgarriba (commented):
if you don't want to use glam; can we try at least with just raw implementation ? https://github.com/kornia/kornia-rs/blob/main/crates/kornia-3d/src/linalg.rs#L91

i don't think faer is the right think to use for this use case

dhruvkjain:
Sorry for late reply,
after doing a bit research how faer, glam and raw implementation would work,
I think for now raw implementation might be better because
- [https://docs.rs/glam/latest/glam/f64/struct.DVec3.html#associatedconstant.USES_WASM32_SIMD](https://docs.rs/glam/latest/glam/f64/struct.DVec3.html#associatedconstant.USES_WASM32_SIMD) : that support SIMD only on wasm32 (for f64 only) but rerun uses native platforms for that so even using glam particular for this case wouldn't improve performance 
- and as you already said faer is not the right thing to use for this use case

PR #345, opened 2025-04-07
Update tiff reading - writing methods for f32 and u8 datatype 

viktorho:
* add traits TiffValueConverter

* add write_image_tiff_internal(), and example for u8 only

* add read_image_tiff_internal(), and example for u8, f32 (experiments)


crates/kornia-io/src/tiff.rs
```diff
@@ -0,0 +1,207 @@
...
+/// file_path - The path to the TIFF file.
+pub fn read_image_tiff_gray8(
+    file_path: impl AsRef<Path>,
+) -> Result<Image<u8, 1>, IoError> {
+    read_image_tiff_internal::<u8,1>(file_path)
+}
+
+/// Reads a TIFF image with RGBA8 channels
+/// #Arguments
+/// file_path - The path to the TIFF file.
```
AS1100K, line 93:
```suggestion
/// Reads a TIFF image with four channels (rgba8).
///
/// # Arguments
///
/// - `file_path` - The path to the TIFF file.
```

Same

crates/kornia-io/src/tiff.rs
```diff
@@ -0,0 +1,207 @@
...
+        tiff::ColorType::RGB(_) => 3,
+        tiff::ColorType::RGBA(_) => 4,
+        _ => return Err(IoError::TiffEncodingError("Unsupported color type".to_string())),
+    };
+    if num_channels != N {
+        return Err(IoError::TiffEncodingError(format!(
+            "Image has {} channels, but expected {}",
+            num_channels, N
+        )));
+    }
```
AS1100K, line 164:
```suggestion
```

Since, it is a internal function, the public functions ensure that the correct number of channels are provided at compile time and this won't be necessary.

crates/kornia-io/src/jpeg.rs
```diff
@@ -119,6 +121,14 @@ mod tests {
...
     }
 
+    #[test]
+    fn read_jpeg_f32() -> Result<(), IoError> {
+        let image = read_image_jpeg_mono8("../../tests/data/dog.jpeg")?;
+        assert_eq!(image.cols(), 258);
+        assert_eq!(image.rows(), 195);
+        Ok(())
+    }
+
```
AS1100K, line 131:
```suggestion
```
Firstly, the name of the test doesn't make sense as jpeg doesn't support f32 and you are using `read_image_jpeg_mono8` in it.

And secondly, this doesn't work. It will give `ImageCreationError(InvalidChannelShape(150930, 50310))` error.

Make sure to run the tests first.

crates/kornia-io/src/tiff.rs
```diff
@@ -0,0 +1,207 @@
...
+    file_path: impl AsRef<Path>,
+) -> Result<Image<u8, 4>, IoError> {
+    read_image_tiff_internal::<u8,4>(file_path)
+}
+
+/// A trait to convert TIFF values to a specific type.
+/// 
+/// This trait is used to convert TIFF values (u8 or f32) to a specific type (e.g., u8 or f32).
+/// It provides two methods:
+pub trait TiffValueConverter: Sized {
```
AS1100K, line 104: I think we should have a simple trait _(could be internal?)_ with only one function like `type_` and should have it's implementation for all numeric primitive types that could be bound to generic `T` _(`P` in your case, change it to `T` so that it's globally same)_.
The `type_` function could be used to verify if the decoded image is of same type, if not return error.

This could be used to to only allow integer primitive types. Also, we should not deal with any conversion and make it up to the user if they want to covert the type.

Then your functions should look something like:
```rust
pub trait NumericTrait {
    fn type_() -> NumericType;
}

pub fn read_image_tiff_rgba8<T: NumericTrait>(file_path: impl AsRef<Path>) -> Result<Image<T, 4>, IoError> {
    todo!()
}

pub fn read_image_tiff_internal<T: NumericTrait>(/* arguments */) -> /* ... */
```

What do you say @edgarriba ?

crates/kornia-io/src/tiff.rs
```diff
@@ -0,0 +1,207 @@
...
+
+
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+    use std::fs;
+    use std::path::PathBuf;
+    use kornia_image::ImageSize;
+    #[test]
```
AS1100K, line 184: Also, add a read_write test

Refer https://github.com/kornia/kornia-rs/blob/d959fb39bd6aba6ab9c9af93ad307abca1a99f45/crates/kornia-io/src/jpeg.rs#L123-L139

crates/kornia-io/src/tiff.rs
```diff
@@ -0,0 +1,207 @@
...
+    fn test_write_image_tiff_rgb8() {
+        let image = Image::new(ImageSize { width: 100, height: 100 }, vec![0; 30000]).unwrap();
+        let file_path = PathBuf::from("test_rgb8.tiff");
+        assert!(write_image_tiff_rgb8(&file_path, &image).is_ok());
+        fs::remove_file(file_path).unwrap();
```
AS1100K, line 189: Use `tempfile` crate for this, as all the similar tests in this crate uses it

crates/kornia-io/src/tiff.rs
```diff
@@ -0,0 +1,207 @@
...
+    }
+
+    #[test]
+    fn test_write_image_tiff_gray8() {
+        let image = Image::new(ImageSize { width: 100, height: 100 }, vec![0; 10000]).unwrap();
+        let file_path = PathBuf::from("test_gray8.tiff");
+        assert!(write_image_tiff_gray8(&file_path, &image).is_ok());
+        fs::remove_file(file_path).unwrap();
```
AS1100K, line 197: Same

AS1100K (changes requested):
Make sure to format your code with rustfmt and run the tests as one of your test fails.

crates/kornia-io/src/error.rs
```diff
@@ -26,6 +26,15 @@ pub enum IoError {
     #[error("Error with Jpeg encoding")]
     JpegEncodingError(#[from] jpeg_encoder::EncodingError),
 
+    
+    /// Error decoding the TIFF image.
+    #[error("Error with TIFF decoding")]
```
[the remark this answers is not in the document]
edgarriba: i advocate more to use `#[error(transparent)]` to forward directly the original errors
https://github.com/kornia/kornia-rs/blob/main/crates/kornia-io/src/stream/error.rs#L5

crates/kornia-io/src/tiff.rs
```diff
@@ -0,0 +1,207 @@
...
+use kornia_image::{Image, ImageSize};
+use tiff::decoder::{Decoder,DecodingResult};
+use tiff::encoder::{
+    colortype::{ColorType,RGB8,Gray8}, 
+    TiffEncoder,
+    TiffValue,
+};
+use std::fs::File;
+use std::path::Path;
+use std::io::BufWriter;
```
edgarriba, line 12: group properly the imports, probably cargo fmt will do this for you

crates/kornia-io/src/tiff.rs
edgarriba: sounds good to me

crates/kornia-io/src/tiff.rs
```diff
@@ -0,0 +1,207 @@
...
+    }
+
+    if file_path.extension().map_or(true, |ext| {
+        !ext.eq_ignore_ascii_case("tif") && !ext.eq_ignore_ascii_case("tiff")
+    }) {
+        return Err(IoError::InvalidFileExtension(file_path.to_path_buf()));
+    }
+
+    let file = File::open(&file_path)?;
+    let mut decoder = Decoder::new(file).map_err(IoError::TiffError)?;
```
edgarriba, line 142:
```suggestion
    let mut decoder = Decoder::new(file)?;
```
forward error diectly like this

crates/kornia-io/src/tiff.rs
viktorho: Hello, I think we should provide more details when implementing images for our application. It could be a marker type instead of struct u8 in this context, for example:
`let rgb_image: Image<u8, 4> = read_image_any::<Rgba8>("image.jpg")?;`
From my vision, this could be more sustainable in case we would provide more features for the fourth channel, like depth or heat instead of alpha.

viktorho: @edgarriba , @AS1100K 
Would it be a good one in this case? If it's good to go, i'm gonna include these for the next pull request

edgarriba: for now we decided in a previous conversation in discord to hold on the read_any and leave generalizations for later

edgarriba:
merged in https://github.com/kornia/kornia-rs/pull/363

PR #349, opened 2025-04-08
Add frozen attribute to Python wrapper classes

### LINKED ISSUE #341: Make immutable pyclasses in `kornia-py` frozen
See https://github.com/kornia/kornia-rs/pull/334. 

This should be a (probably small) optimization, since there’s no need to do runtime atomic reference counting on immutable types.

AS1100K:
Fixes #341


edgarriba (approved):

ngoldbaum:
Awesome, I’m glad this was trivial. Definitely an argument in favor of making this the default for all pyclasses.

PR #352, opened 2025-04-11
feat: implement `draw_rect` and `draw_filled_rect`

Mojam847:
- implement kornia_imgproc::draw::`draw_rect` and `draw_filled_rect`

### DIFF
--- a/crates/kornia-imgproc/src/draw.rs
+++ b/crates/kornia-imgproc/src/draw.rs
@@ -1,6 +1,23 @@
 use kornia_image::Image;
+use std::cmp::{max, min};
 
-/// Draws a line on an image inplace.
+/// Helper function to set a pixel's color, handling bounds checking.
+#[inline]
+fn set_pixel<const C: usize>(img: &mut Image<u8, C>, x: i64, y: i64, color: [u8; C]) {
+    if x >= 0 && x < img.cols() as i64 && y >= 0 && y < img.rows() as i64 {
+        let pixel_linear_index = (y * img.cols() as i64 + x) * C as i64;
+        // Bounds check above makes this indexing safe.
+        let start = pixel_linear_index as usize;
+        // Check if the slice operation is within bounds (extra safety)
+        if start + C <= img.as_slice().len() {
+             img.as_slice_mut()[start..start + C].copy_from_slice(&color);
+        } else {
+             eprintln!("Warning: Attempted to write out of bounds at ({}, {})", x, y);
+        }
+    }
+}
+
+/// Draws a line on an image inplace using a standard Bresenham's line algorithm.
 ///
 /// # Arguments
 ///
@@ -8,7 +25,7 @@ use kornia_image::Image;
 /// * `p0` - The start point of the line as a tuple of (x, y).
 /// * `p1` - The end point of the line as a tuple of (x, y).
 /// * `color` - The color of the line as an array of `C` elements.
-/// * `thickness` - The thickness of the line.
+/// * `thickness` - The thickness of the line. (Note: thickness > 1 is approximate).
 pub fn draw_line<const C: usize>(
     img: &mut Image<u8, C>,
     p0: (i64, i64),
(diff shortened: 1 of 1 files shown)


archiv-nptel:
Hi @edgarriba,

Could you please provide your review and approve the workflow for this PR whenever you get a chance? Let me know if any changes are needed — I’m happy to address them.

Looking forward to your feedback so I can proceed further.

Thank you!

crates/kornia-imgproc/src/draw.rs
```diff
@@ -1,99 +1,256 @@
 use kornia_image::Image;
+use std::cmp::{max, min};
 
-/// Draws a line on an image inplace.
+/// Helper function to set a pixel's color, handling bounds checking.
+#[inline]
```
edgarriba, line 5: Why inline ?

crates/kornia-imgproc/src/draw.rs
```diff
@@ -1,99 +1,256 @@
...
+fn set_pixel<const C: usize>(img: &mut Image<u8, C>, x: i64, y: i64, color: [u8; C]) {
+    if x >= 0 && x < img.cols() as i64 && y >= 0 && y < img.rows() as i64 {
+        let pixel_linear_index = (y * img.cols() as i64 + x) * C as i64;
```
edgarriba, line 8: Think if there’s any way that y * cols can be precomputed or similar so that later we iterate over sums

crates/kornia-imgproc/src/draw.rs
```diff
@@ -1,99 +1,256 @@
...
+/// Draws a line on an image inplace using a standard Bresenham's line algorithm.
 ///
 /// # Arguments
 ///
 /// * `img` - The image to draw on.
 /// * `p0` - The start point of the line as a tuple of (x, y).
 /// * `p1` - The end point of the line as a tuple of (x, y).
 /// * `color` - The color of the line as an array of `C` elements.
-/// * `thickness` - The thickness of the line.
+/// * `thickness` - The thickness of the line. (Note: thickness > 1 is approximate).
```
edgarriba, line 28: Worth mimic this one https://github.com/pygame/pygame/blob/79807da84c9bacf8df5a177763e14c924e3b15e2/src_c/draw.c#L1624

edgarriba (changes requested):

crates/kornia-imgproc/src/draw.rs
Rejoan-Sardar: @edgarriba The current draw_line implementation approximates thickness > 1 by drawing a small filled square around each point on the Bresenham line. This can look blocky, especially for diagonal lines. The linked Pygame code likely implements a more geometrically accurate method for thick lines (e.g., drawing parallel lines or calculating the line's polygonal bounds and filling it).

For this iteration, we'll keep the current approximate thickness implementation in draw_line but acknowledge its limitation in the documentation more clearly. A more sophisticated implementation could be a future enhancement. The focus here is resolving the specific, actionable feedback points (inline, precomputation, bounds check).

crates/kornia-imgproc/src/draw.rs
Rejoan-Sardar: ok, we can remove the explicit #[inline] hint and trust the compiler's optimization decisions.

crates/kornia-imgproc/src/draw.rs
Mojam847: done

crates/kornia-imgproc/src/draw.rs
Mojam847: removed the explicit #[inline] hint and trust the compiler's optimization decisions.

crates/kornia-imgproc/src/draw.rs
Mojam847: @edgarriba Pls check your current `draw_line` implementation approximates thickness > 1 by drawing a small filled square around each point on the Bresenham line. This can look blocky, especially for diagonal lines. The linked Pygame code likely implements a more geometrically accurate method for thick lines (e.g., drawing parallel lines or calculating the line's polygonal bounds and filling it).
For this iteration, we'll keep the current approximate thickness implementation in `draw_line` but acknowledge its limitation in the documentation more clearly.

crates/kornia-imgproc/src/draw.rs
```diff
@@ -9,91 +21,238 @@ use kornia_image::Image;
...
+
+    let lx0 = min(x0, x1);
+    let ly0 = min(y0, y1);
+    let lx1 = max(x0, x1);
+    let ly1 = max(y0, y1);
+
+    let half_thickness = (thickness / 2) as i64;
+
+    // Draw horizontal lines
+    draw_line(img, (lx0, ly0), (lx1, ly0), color, thickness); // Top
```
edgarriba, line 109: Drawing the horizontal lines like this is very inefficient

crates/kornia-imgproc/src/draw.rs
```diff
@@ -9,91 +21,238 @@ use kornia_image::Image;
...
+    let img_rows = img.rows() as i64;
+    let x_min = max(0, x_min_coord);
+    let y_min = max(0, y_min_coord);
+    let x_max = min(img_cols, x_max_coord);
+    let y_max = min(img_rows, y_max_coord);
+
+    let data = img.as_slice_mut();
+    let data_len = data.len();
+
+    for y in y_min..y_max {
```
edgarriba, line 149: implement an Interator based draw_horizontal_line and just draw them sequentally

crates/kornia-imgproc/src/draw.rs
Mojam847: While `set_pixel` itself doesn't loop, the optimization concern is valid for its callers. The refactoring of `draw_rect` and `draw_filled_rect` to use `draw_horizontal_line_segment` addresses this effectively. Inside `draw_horizontal_line_segment`, the base row offset `(y_usize * cols_usize)` is calculated only once per row segment, and then pixels within that segment are accessed efficiently.

crates/kornia-imgproc/src/draw.rs
Mojam847: Also I've updated the documentation comment for draw_line to be much clearer about the fact that it uses a square approximation and that this can lead to blocky appearances, especially for diagonal lines, explicitly mentioning it differs from more precise algorithms.

crates/kornia-imgproc/src/draw.rs
Mojam847: Agreed. The previous nested loop calculated pixel indices individually. I've refactored `draw_filled_rect` to iterate only through the necessary rows (y coordinates). Inside the loop, it now calls the same efficient `draw_horizontal_line_segment` helper function created for `draw_rect`. This avoids redundant calculations and leverages optimized memory filling for each row segment.

crates/kornia-imgproc/src/draw.rs
Mojam847: @edgarriba You are absolutely right. Calling `draw_line` (which uses `Bresenham's algorithm`) for simple horizontal lines is inefficient. I've introduced a new helper function `draw_horizontal_line_segment`. This function directly calculates the start and end indices within the image's data slice for the given row (y) and x range, and then fills that memory segment efficiently using chunks_exact_mut and copy_from_slice. The draw_rect function now uses this helper for drawing the top and bottom horizontal parts of the rectangle, significantly improving performance for those segments.

edgarriba (changes requested):
looks good to me -- only update test to be more robust

edgarriba:
can you rebase form main ?

edgarriba:
Are you still interested in merging this ?

Issue #351, opened 2025-04-09
add `draw_polygon` and `draw_filled_polygon` functions.


TomatoChocolate12:
I will work on this. Opening an issue just because I had two questions:

- should i add these functions to `draw.rs` or should they be somewhere else? maybe in the `kornia-imgproc::color` crate.
- i intend to use the *scanline-fill* algorithm to make the filled polygons. would you suggest something else?

edgarriba:
the `draw.rs` should be the right place

archiv-nptel:
HI @edgarriba I was thinking after draw_line, we could also add other basic drawing functions in draw.rs like:

`draw_rect, draw_filled_rect`
`draw_circle, draw_filled_circle`
`draw_ellipse, draw_filled_ellipse`
These would follow — in-place drawing, basic primitives, minimal dependencies.

If that sounds good, I’d like to work on adding some of these.

TomatoChocolate12:
Drawing a finite-sided polygon like a rectangle is fairly simple and should be achieved by just providing it's vertices as argument.

But, drawing an ellipse or a circle would be far more complicated as I assume there would be an irrational number of pixels if we were to try to draw it perfectly. So we might have to make some estimations and assumptions to design such functions.

 @Mojam847 if you have any ideas on how to handle shapes like ellipses and circles, I would love to discuss with you and then try to implement them!

edgarriba:
all sounds good, please make sure one function per PR choosing a good reference implementation to compare with

PR #353, opened 2025-04-11
adds `draw_polygon` and `draw_filled_polygon` functions.

TomatoChocolate12:
Solves issue #351 .

Adds `draw_polygon` and `draw_filled_polygon` functions. Uses the scanline fill algorithm to implement `draw_filled_polygon`. Added tests as well.

Reference implementation can be found here:

https://github.com/pygame/pygame/blob/79807da84c9bacf8df5a177763e14c924e3b15e2/src_c/draw.c#L2322


crates/kornia-imgproc/src/draw.rs
```diff
@@ -98,4 +198,47 @@ mod tests {
...
+                width: 5,
+                height: 5,
+            },
+            vec![0; 25],
+        )?;
+        let points: [(i64, i64); 3] = [(0, 0), (0, 3), (4, 0)];
+        draw_polygon(&mut img, &points, [255], 1);
+        assert_eq!(
+            img.as_slice(),
+            vec![
```
edgarriba, line 215: Can you format this so that is more readable? You might need to use the skipfmt macro (check in other examples)

edgarriba (changes requested):

crates/kornia-imgproc/src/draw.rs
TomatoChocolate12: yes it appears, the rustfmt during pre-commit messed it up. i'll use the macro like you suggested.

crates/kornia-imgproc/src/draw.rs
```diff
@@ -70,9 +70,109 @@ pub fn draw_line<const C: usize>(
...
+    for i in 0..n {
+        let (x0, y0) = points[i];
+        let (x1, y1) = points[(i + 1) % n];
+        edges.push((x0, y0, x1, y1));
+    }
+
+    // bounds for y to find range of scanlines
+    let min_y = points.iter().map(|&(_, y)| y).min().unwrap();
+    let max_y = points.iter().map(|&(_, y)| y).max().unwrap();
+
```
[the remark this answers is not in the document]
TomatoChocolate12: the reason i left this out was because i wanted to draw the outline of the polygon anyway and a single pixel width polygon would have had color because of the polygon perimeter anyway.

crates/kornia-imgproc/src/draw.rs
```diff
@@ -70,9 +70,109 @@ pub fn draw_line<const C: usize>(
...
+    // for each scanline
+    for y in min_y..=max_y {
+        let mut x_intersections = Vec::new();
+
+        for &(x0, y0, x1, y1) in &edges {
+            // checking if the scanline intersects the edges.
```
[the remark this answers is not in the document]
TomatoChocolate12: i did not understand, could you please elaborate?

crates/kornia-imgproc/src/draw.rs
```diff
@@ -70,9 +70,114 @@ pub fn draw_line<const C: usize>(
...
+                        let base = (y * img.cols() as i64 + x) * C as i64;
+                        let pixel_slice = &mut img.as_slice_mut()[base as usize..base as usize + C];
+                        pixel_slice.copy_from_slice(&fill_color);
+                    }
+                }
+            }
+        }
+    }
+    // drawing the polygon outline
+    draw_polygon(img, points, line_color, thickness);
```
edgarriba, line 175: i think this is not valuable to have it here; as any user can call themselves first `draw_filled_polygon` and later `draw_polygon` and in this way you are forcing anyone that wants a filled polygon to draw also the outline.

Please, update the function header without the `line_color` and remove this last `draw_polygon` call to have very decoupled functionality.

crates/kornia-imgproc/src/draw.rs
```diff
@@ -98,4 +203,79 @@ mod tests {
...
         Ok(())
     }
+
+    #[rustfmt::skip]
+    #[test]
+    fn test_draw_polygon() -> Result<(), ImageError> {
+        let mut img = Image::new(
+            ImageSize {
+                width: 5,
+                height: 5,
```
edgarriba, line 213:
```suggestion
                height: 6,
```
can you try with non-squared images (it's usually a good testing practice with working with images)

edgarriba (changes requested):

edgarriba (approved):

PR #357, opened 2025-04-15
feat: implemented otsu threshold

aayushhzz:
Implemented Otsu Threshold using this [wikipedia](https://en.wikipedia.org/wiki/Otsu%27s_method#:~:text=In%20computer%20vision%20and%20image,two%20classes%2C%20foreground%20and%20background ) article as the reference

### DIFF
--- a/crates/kornia-imgproc/src/threshold.rs
+++ b/crates/kornia-imgproc/src/threshold.rs
@@ -5,6 +5,19 @@ use kornia_image::{Image, ImageError};
 
 use crate::parallel;
 
+/// The type of thresholding to apply.
+pub enum ThresholdType {
+    /// Binary thresholding
+    Binary,
+    /// Inverse binary thresholding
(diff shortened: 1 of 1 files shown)


crates/kornia-imgproc/src/threshold.rs
```diff
@@ -365,7 +378,140 @@ where
...
+            src.rows(),
+            dst.cols(),
+            dst.rows(),
+        ));
+    }
+
+    let height = src.height();
+    let width = src.width();
+    const BINS: usize = 256;
+    let mut histogram = [0u32; BINS];
```
edgarriba, line 441: we have already an histogram function

edgarriba (changes requested):
Can you implement based to a specific implementation and compare byte to byte the results ? This paper could be a good reference https://www.ipol.im/pub/art/2016/158/article_lr.pdf

PR #359, opened 2025-05-01
implement v4l2 crate

edgarriba:
This pull request introduces significant updates to the `kornia-imgproc` and `kornia-io` crates, focusing on adding YUV-to-RGB conversion functionality, improving FPS calculation, and integrating V4L2 (Video for Linux 2) support for video capture and camera control. Below is a categorized summary of the most important changes:

### YUV-to-RGB Conversion:
* Added a new module `yuv` with a function `convert_yuyv_to_rgb_u8` to convert YUYV image data to RGB format. This includes helper functions for YUV-to-RGB color space conversion and clamping values. (`crates/kornia-imgproc/src/color/yuv.rs`, [crates/kornia-imgproc/src/color/yuv.rsR1-R67](diffhunk://#diff-c63f00564c6a2a0cfc398974d90b4437a565c579ff56dcafd50a3c3abebf1783R1-R67))
* Exposed the `convert_yuyv_to_rgb_u8` function in the `color` module for public use. (`crates/kornia-imgproc/src/color/mod.rs`, [crates/kornia-imgproc/src/color/mod.rsR3-R7](diffhunk://#diff-871df256bec93e288fa520c95fc0b8843388e2bc2f082a141c888a6632259e1cR3-R7))

### FPS Counter Refactor:
* Replaced the previous smoothing-based FPS calculation with a sliding window approach using a `VecDeque`. This allows for more accurate and configurable FPS computation by tracking frame timestamps. (`crates/kornia-io/src/fps_counter.rs`, [[1]](diffhunk://#diff-3e5701c6ce762ab30d689cc42eee63dace1c57a73a447879f50eb84b0c89e880R1-L5) [[2]](diffhunk://#diff-3e5701c6ce762ab30d689cc42eee63dace1c57a73a447879f50eb84b0c89e880L20-R58)

### V4L2 Integration:
* Introduced a new `v4l` module for real-time video capture and camera control, including:
  - `V4LVideoCapture` for handling video streams.
  - `CameraControl` and `AutoExposureMode` enums for managing camera settings like exposure, brightness, and white balance.
  - `PixelFormat` enum for supported camera pixel formats (e.g., YUYV, MJPG). (`crates/kornia-io/src/v4l/`, [[1]](diffhunk://#diff-f35c62538f975eb049035a39b623262baff01ece3a53d6427c1678ddf920d098R1-R138) [[2]](diffhunk://#diff-6ab693adbf0de6269f568b5cbdda70b6fe13b0787898f3d7714c375a660d5a98R1-R65) [[3]](diffhunk://#diff-a75a742aae862f4c63a5350112c8742c05535ac2738e9906ceafc7bdea29f0cbR1-R82) [[4]](diffhunk://#diff-6e827b97c037afaba152568d875e165551e1cef0c2f5223735e30d2d4fa6db6cR1-R126)
* Added optional dependencies (`v4l`, `libc`) and a corresponding feature flag (`v4l`) in the `Cargo.toml` file to enable V4L2 integration. (`crates/kornia-io/Cargo.toml`, [[1]](diffhunk://#diff-f4747ba2294b00897d87c3ed094f6d22befc421aae9cb0f7c89c448d82fc59e4R31-R33) [[2]](diffhunk://#diff-f4747ba2294b00897d87c3ed094f6d22befc421aae9cb0f7c89c448d82fc59e4R47)

### Miscellaneous Improvements:
* Updated the `Cargo.toml` file to explicitly define workspace attributes for metadata fields like `authors`, `edition`, and `repository`. (`crates/kornia-io/Cargo.toml`, [crates/kornia-io/Cargo.tomlL3-R11](diffhunk://#diff-f4747ba2294b00897d87c3ed094f6d22befc421aae9cb0f7c89c448d82fc59e4L3-R11))
* Added a minor code cleanup in `jpeg.rs` by removing an unnecessary newline. (`crates/kornia-io/src/jpeg.rs`, [crates/kornia-io/src/jpeg.rsR169](diffhunk://#diff-60a1d9bf7a0b66a3ae525ba7725968faf2813212ae1d822b4376ab77043f24dfR169))

These changes enhance the functionality and flexibility of the `kornia` libraries, particularly for image processing and video capture use cases.


crates/kornia-io/src/stream/v4l.rs
```diff
@@ -0,0 +1,86 @@
...
+}
+
+/// A video capture object for a V4L2 camera.
+pub struct V4LVideoCapture {
+    stream: Stream<'static>,
+    size: ImageSize,
+}
+
+/// A frame of video from a V4L2 camera.
+pub struct ImageFrame<'a> {
```
edgarriba, line 36: @AS1100K something in this direction could work. Or we can implement a ImageView (immutable) struct with a lifetime param. The timestamp is a feature I would like to support as it's very important for robotics and iot applications

AS1100K: I can think of two possible solutions that can be used here:

1. Making `Image`, have any `Allocator` and not only `CpuAllocator`. This would allow having more control over the ownership and could simplify other stuff like in #338. And we could have a metadata field in `Image` that can be used to have metadata info like timestamp.
2. Other option would be having a `Image` trait instead of a struct that would allow having custom image types.

I think that we should look into something similar to first approach, as it would integrate much better with existing tensor and imgproc apis

edgarriba: probably allowing to have any Allocator is the way to go. I would like to decouple the timestamp from the image struct, we can just return a tuple for now

crates/kornia-io/src/stream/v4l.rs
```diff
@@ -0,0 +1,119 @@
...
+/// A video capture object for a V4L2 camera.
+pub struct V4LVideoCapture<'a> {
+    stream: Stream<'a>,
+    size: ImageSize,
+}
+
+/// An allocator for a V4L2 camera
+#[allow(dead_code)]
+#[derive(Clone)]
+pub struct V4lAllocator<'a>(&'a [u8]);
```
edgarriba: @AS1100K @gbin how do you feel about this api here ?

crates/kornia-io/src/stream/v4l.rs
```diff
@@ -0,0 +1,119 @@
...
+            size: ImageSize {
+                width: config.size.width,
+                height: config.size.height,
+            },
+        })
+    }
+
+    /// Grab a frame from the V4L2 camera.
+    pub fn grab(&mut self) -> Option<Result<ImageFrame, V4L2Error>> {
+        let Ok((buffer, metadata)) = self.stream.next() else {
```
AS1100K: The [`next`](https://docs.rs/v4l/latest/v4l/io/mmap/stream/struct.Stream.html#method.next) method uses different lifetime then the Stream lifetime.

The Stream has lifetime `'a` while the buffer returned by `next` has lifetime `'b`, but we are assuming it to be `'a`, in this case user can move this value anywhere until `V4LVideoCapture` is valid, but if the next method is called in the meantime, the buffer vale is changed which could be an unexpected behavior and could also result to other issues.

crates/kornia-imgproc/src/color/yuv.rs
```diff
@@ -0,0 +1,64 @@
...
+///
+/// # Arguments
+///
+/// * `src` - The YUYV image data.
+/// * `dst` - The RGB image to store the result.
+///
+/// # Returns
+///
+/// The RGB image in HxWx3 format.
+pub fn convert_yuyv_to_rgb_u8<A: ImageAllocator>(src: &[u8], dst: &mut Image<u8, 3, A>) {
```
AS1100K: Can we use `Image<u8, 2, A>` instead of `&[u8]` as that would allow us to verify if `src` and `dst` are of same dimensions, and `chunks_exact_mut` won't skip the last pixels if they don't fit in the chunks

crates/kornia-io/src/v4l/stream.rs
```diff
@@ -0,0 +1,221 @@
...
+    /// use v4l::buffer::Type;
+    /// use v4l::device::Device;
+    /// use v4l::io::userptr::Stream;
+    ///
+    /// let dev = Device::new(0);
+    /// if let Ok(dev) = dev {
+    ///     let stream = Stream::new(&dev, Type::VideoCapture);
+    /// }
+    /// ```
+    #[allow(dead_code)]
```
AS1100K, line 46: I don't think you do need to pass `#[allow(dead_code)]` for public functions

crates/kornia-imgproc/src/color/yuv.rs
AS1100K: Or maybe we can just verify the length of both src and dst?

crates/kornia-io/src/v4l/arena.rs
```diff
@@ -0,0 +1,132 @@
+// Modified from https://github.com/raymanfx/libv4l-rs/blob/30f6dbaeed5cdb1b33760fc5b35c1b545cfe0e46/src/io/userptr/arena.rs
+use std::{io, mem, sync::Arc};
+use v4l::{buffer, device::Handle, memory::Memory, v4l2, v4l_sys::*};
+
+#[derive(Clone)]
+pub struct V4lBuffer(Arc<Vec<u8>>);
```
AS1100K, line 7:
```suggestion
pub type V4lBuffer = Arc<Vec<u8>>;
```

Maybe let it be a type, as it doesn't provide any separate functionality?

crates/kornia-imgproc/src/color/yuv.rs
```diff
@@ -0,0 +1,64 @@
...
+    let width = dst.width();
+    let rgb_data = dst.as_slice_mut();
+
+    rgb_data
+        .chunks_exact_mut(width * 3)
```
AS1100K: Should we use [`par_chunks_mut`](https://docs.rs/rayon/latest/rayon/slice/trait.ParallelSliceMut.html#method.par_chunks_mut) instead?

crates/kornia-io/src/v4l/mod.rs
```diff
@@ -0,0 +1,138 @@
...
+    pixel_format: PixelFormat,
+    device: Device,
+}
+
+/// Represents a captured frame
+pub struct EncodedFrame {
+    /// The buffer of the frame
+    pub buffer: arena::V4lBuffer,
+    /// The fourcc of the frame
+    pub pixel_format: PixelFormat,
```
AS1100K, line 71: Should we instead hold the buffer instead `PixelFormat` like

```rust
pub enum PixelFormat<A: ImageAllocator> {
    YUYV(Image<u8, 2, A>),
    MJPG(/* Unsure here */),
    // ...
}
```

crates/kornia-io/src/v4l/pixel_format.rs
```diff
@@ -0,0 +1,82 @@
...
+use std::str::FromStr;
+use v4l::FourCC;
+
+/// Supported camera pixel formats
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub enum PixelFormat {
+    /// YUYV 4:2:2 format (uncompressed, good quality, high bandwidth)
+    YUYV,
+    /// Motion JPEG (compressed, excellent quality, lower bandwidth)
+    MJPG,
```
AS1100K, line 11: There are a lot of formats supported by v4l https://docs.kernel.org/userspace-api/media/v4l/pixfmt.html#pixfmt which includes RGB and RGBA, IMO we should support them at least.

crates/kornia-imgproc/src/color/yuv.rs
edgarriba: Encoded images do need a different api from the current Image. Think of this function similar to the JPEG decoder and similar

crates/kornia-io/src/v4l/pixel_format.rs
edgarriba: Step by step, besides i don’t have for now a camera that supports other formats which I can test

crates/kornia-io/src/v4l/stream.rs
edgarriba: This supposed to be private to general user. I’ll look into it

crates/kornia-io/src/v4l/arena.rs
edgarriba: In this way it’s more abstract

crates/kornia-io/src/v4l/mod.rs
edgarriba: Yuyv format doesn’t fit in the existing image api

AS1100K (approved):

PR #363, opened 2025-05-02
io: implement image tiff read and write

edgarriba:


edgarriba:
@AS1100K if you want to review

AS1100K (commented):
Everything lgtm. Why not we also add functions for `u32`, `f32`, `f64`, and all other signed integer type too?

edgarriba:
added for now only f32 to have the most common used formats

PR #373, opened 2025-05-19
Add support for custom Allocator for `Image`

AS1100K:
## Future Work
This PR would allow using custom Allocators for `Image`, which will be essential for zero-copy stuff:
- Would unblock #338 for zero-copy gstreamer frame 
- Zero-copy for kornia-py

### DIFF
--- a/kornia-py/src/color.rs
+++ b/kornia-py/src/color.rs
@@ -1,7 +1,7 @@
 use pyo3::prelude::*;
 
 use crate::image::{FromPyImage, PyImage, ToPyImage};
-use kornia_image::Image;
+use kornia_image::{allocator::CpuAllocator, Image};
 use kornia_imgproc::color;
 
 #[pyfunction]
@@ -9,7 +9,7 @@ pub fn rgb_from_gray(image: PyImage) -> PyResult<PyImage> {
     let image_gray = Image::from_pyimage(image)
         .map_err(|e| PyErr::new::<pyo3::exceptions::PyException, _>(format!("src image: {}", e)))?;
 
-    let mut image_rgb = Image::from_size_val(image_gray.size(), 0u8)
+    let mut image_rgb = Image::from_size_val(image_gray.size(), 0u8, CpuAllocator)
         .map_err(|e| PyErr::new::<pyo3::exceptions::PyException, _>(format!("dst image: {}", e)))?;
 
     color::rgb_from_gray(&image_gray, &mut image_rgb).map_err(|e| {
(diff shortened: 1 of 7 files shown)


crates/kornia-image/src/image.rs
```diff
@@ -56,11 +57,11 @@ impl From<ImageSize> for [u32; 2] {
 /// Represents an image with pixel data.
 ///
 /// The image is represented as a 3D Tensor with shape (H, W, C), where H is the height of the image,
-pub struct Image<T, const C: usize>(pub Tensor3<T, CpuAllocator>);
+pub struct Image<T, const C: usize, A: ImageAllocator + 'static = CpuAllocator>(pub Tensor3<T, A>);
```
edgarriba, line 60: why is the static needed here ? can we do not give a default ?

crates/kornia-image/src/ops.rs
```diff
@@ -12,26 +12,28 @@ use crate::{Image, ImageError};
...
 ///
 /// cast_and_scale(&image, &mut image_f32, 1. / 255.0).unwrap();
 ///
 /// assert_eq!(image_f32.get_pixel(0, 0, 0).unwrap(), &0.0f32);
 /// assert_eq!(image_f32.get_pixel(1, 0, 0).unwrap(), &1.0f32);
 /// ```
-pub fn cast_and_scale<T, U, const C: usize>(
-    src: &Image<T, C>,
-    dst: &mut Image<U, C>,
+pub fn cast_and_scale<T, U, const C: usize, A1: ImageAllocator, A2: ImageAllocator>(
```
edgarriba, line 34: are we gonna have any conflicts if we give images with two different and incompatibles allocator types ?

crates/kornia-image/src/image.rs
AS1100K: When not adding `'static` bound the compiler gives the following error:
```
error[E0310]: the parameter type `A` may not live long enough
  --> crates/kornia-image/src/image.rs:60:71
   |
56 | #[derive(Clone)]
   |          ----- in this derive macro expansion
...
60 | pub struct Image<T, const C: usize, A: ImageAllocator = CpuAllocator>(pub Tensor3<T, A>);
   |                                                                       ^^^^^^^^^^^^^^^^^
   |                                                                       |
   |                                                                       the parameter type `A` must be valid for the static lifetime...
   |                                                                       ...so that the type `A` will meet its required lifetime bounds
   |
help: consider adding an explicit lifetime bound
   |
60 | pub struct Image<T, const C: usize, A: ImageAllocator + 'static = CpuAllocator>(pub Tensor3<T, A>);
   |                                                       +++++++++
```
therefore, I added `'static` bound

crates/kornia-image/src/ops.rs
AS1100K: By the current `TensorAllocator` design, they are only responsible for allocation and deallocation, and not have any influence on the retrieval data stored. I will add some tests for the same, but unsure how would incompatibles allocators look like?

There, are some functions like https://github.com/kornia/kornia-rs/blob/53a9ba465626a90eb9a1db05b8b8c04a0c8bfb8d/crates/kornia-tensor/src/storage.rs#L66-L86 doesn't even utilize `alloc`.

Should we explore utilizing those allocators for `Vec` by requiring _(or even optional)_ [`Allocator` Trait](https://doc.rust-lang.org/stable/std/alloc/trait.Allocator.html) or something similar

edgarriba (approved):

crates/kornia-image/src/image.rs
edgarriba: i'll address in a separated PR

PR #377, opened 2025-05-26
chore: bump version of rerun and criterion, fix warnings

edgarriba:


AS1100K (changes requested):
`criterion::black_box` is now deprecated and it is suggested to use `std::hint::black_box`

PR #384, opened 2025-06-04
Add Adaptive Threshold Binary function

AS1100K:
- Input Image
preprocess

- Result of Adaptive Threshold Image
threshold

</details>

## TODOs

- [x] Tests
- [x] Document the `adaptive_threshold` function


crates/kornia-apriltag/src/threshold.rs
```diff
@@ -0,0 +1,198 @@
...
+    let dst_data = dst.as_slice_mut();
+
+    let tiles_x_len = (src.width() as f32 / tile_size as f32).ceil() as usize; // number of horizontal tiles
+    let tiles_y_len = (src.height() as f32 / tile_size as f32).ceil() as usize; // number of vertical tiles
+
+    // pixels available in the tiles at edge
+    let last_tile_x_px = tile_size - (src.width() % tile_size);
+    let last_tile_y_px = tile_size - (src.height() % tile_size);
+
+    let mut tile_min: Vec<T> = Vec::with_capacity(tile_size * tile_size);
```
edgarriba: Should we implement this as a stateful struct so that we don't have to allocate this buffers every time ?

crates/kornia-apriltag/src/threshold.rs
```diff
@@ -0,0 +1,198 @@
...
+    let mut tile_max: Vec<T> = Vec::with_capacity(tile_size * tile_size);
+
+    // Calculate extrema (i.e. min & max grayscale value) of each tile
+    for y in 0..tiles_y_len {
```
edgarriba: Can we implement this with iterators instead?

edgarriba (commented):
Add basic unit tests

crates/kornia-apriltag/src/threshold.rs
AS1100K: I believe we can, I would explore [`ParallelIterator`](https://docs.rs/rayon/latest/rayon/iter/trait.ParallelIterator.html) and will try to implement it by today

edgarriba: You can start with plane iterative first

crates/kornia-imgproc/src/iter.rs
```diff
@@ -0,0 +1,398 @@
...
+    TileIteratorMut
+);
+
+/// An iterator over tiles of an image, yielding non-overlapping rectangular regions as slices.
+///
+/// Each item yielded by the iterator is a slice of rows, where each row is a slice of pixel data.
+/// The tile size is specified at construction, and the iterator will yield tiles of the given size,
+/// except for tiles at the image edges, which may be smaller if the image dimensions are not
+/// multiples of the tile size.
+pub struct TileIterator<'a, T> {
```
AS1100K: @edgarriba Should we keep the tile iterator stuff in kornia-imgproc or in other crate?

Also, should we add the benchmarks for the iterator (probably in another PR). I did experiment and benchmarks with some iterators you can [see here](https://github.com/AS1100K/rust-experiments/tree/master/tile_iterator_benchmark)

crates/kornia-apriltag/src/threshold.rs
```diff
@@ -0,0 +1,285 @@
...
+
+    let tile_iterator = TileIterator::from_image(src, tile_size);
+    let tiles_x_len = tile_iterator.tiles_x_len();
+    let tiles_y_len = tile_iterator.tiles_y_len();
+
+    // Binarize the image
+    for ((y, x), tile) in tile_iterator.tile_enumerator() {
+        let tile_index = (y * tiles_x_len) + x;
+
+        let mut neighbor_min = tile_buffers.tile_min[tile_index];
```
cjpurackal: I think this crashes if `TileBuffers` was created for different image size, so maybe add a validation?
```
let tile_buffers = TileBuffers::new(ImageSize { width: 2, height: 2 }, 2); // 1 tile
adaptive_threshold(&larger_image, &mut dst, &mut tile_buffers, 2, 20); // Tries to access tile[1], [2], 
```

AS1100K: Added validation for length of `TileBuffers` 👍️

crates/kornia-apriltag/src/threshold.rs
```diff
@@ -0,0 +1,330 @@
...
+
+                for px in dst_data.iter_mut().take(end_index).skip(start_index) {
+                    *px = T::SKIP_PROCESSING;
+                }
+            }
+
+            continue;
+        }
+
+        macro_rules! neighbor_min_max {
```
edgarriba: why a macro ? that should be an implement function in the Tilesbuffer that includes some tests

crates/kornia-apriltag/src/threshold.rs
```diff
@@ -0,0 +1,330 @@
...
+        let mut dst = Image::from_size_val(src.size(), 0u8, CpuAllocator).unwrap();
+
+        let mut tile_buffers = TileBuffers::new(src.size(), 2);
+        adaptive_threshold(&src, &mut dst, &mut tile_buffers, 2, 20).unwrap();
+        assert_eq!(dst.as_slice(), &[u8::SKIP_PROCESSING; 16]);
+    }
+
+    #[test]
+    #[should_panic(
+        expected = "called `Result::unwrap()` on an `Err` value: InvalidTileBufferSize(2, 4)"
```
edgarriba: why ? tests should be deterministic

crates/kornia-apriltag/src/utils.rs
```diff
@@ -0,0 +1,35 @@
+/// Trait representing a pixel value
+pub trait PixelTrait {
```
edgarriba: why don't just an enum ? then `Image<Pixel, C>`
you can always implement a return value for each enum value

```rust
#[derive(Debug, Clone, Copy, PartialEq, Eq, Default)]
    enum Pixel {
        Black,
        White,
        #[default]
        Skip,
    }

    impl Pixel {
        fn value(&self) -> f64 {
            match self {
                Pixel::Black => 0.0,
                Pixel::White => 1.0,
                Pixel::Skip => 0.5,
            }
        }
    }

    let img =
        Image::<Pixel, 1, _>::from_size_val(dst.size(), Pixel::Skip, kornia_tensor::CpuAllocator)
            .unwrap();
```

edgarriba: start simple not with a lots of generics. We want something working first with common types

crates/kornia-imgproc/src/iter.rs
```diff
@@ -0,0 +1,398 @@
...
+    last_tile_y_px: usize,
+    /// Index of the next tile to yield horizontally.
+    next_tile_x_index: usize,
+    /// Index of the next tile to yield vertically.
+    next_tile_y_index: usize,
+    /// Buffer holding mutable references to the rows of the current tile.
+    buffer: Vec<&'a mut [T]>,
+}
+
+impl<'a, T> TileIteratorMut<'a, T> {
```
edgarriba: why do we need a mutable version ?

crates/kornia-imgproc/src/iter.rs
edgarriba: let's keep it in april tag for now and make it specific to that. If needed we generalize later

crates/kornia-apriltag/src/threshold.rs
```diff
@@ -0,0 +1,330 @@
...
+        return Err(AprilTagError::InvalidTileBufferSize(
+            tile_buffers.tile_min.len(),
+            expected_tile_count,
+        ));
+    }
+
+    let dst_data = dst.as_slice_mut();
+
+    // Calculate extrema (i.e. min & max grayscale value) of each tile
+    for (i, tile) in tile_iterator.enumerate() {
```
edgarriba: can we simplify first with plain iterators ? so that it's easy to expand lwater to par_iter

edgarriba: i think that you can just iterate per pixel and have pre-computed offsets (as you know the image cols) to get the max/min per tile in a 3x3 neighbourhood

crates/kornia-apriltag/src/threshold.rs
```diff
@@ -0,0 +1,330 @@
...
+        + Sub<T, Output = T>
+        + Div<T, Output = T>
+        + From<u8>,
+    A1: ImageAllocator,
+    A2: ImageAllocator,
+>(
+    src: &Image<T, 1, A1>,
+    dst: &mut Image<T, 1, A2>,
+    tile_buffers: &mut TileBuffers<T>,
+    tile_size: usize,
```
edgarriba: this is part already fro mthe tile  buffer , right ?

edgarriba (changes requested):
i'd simplify the code to make it also simple to read/maintain, we can come up with generalizations later in an optimization phase

crates/kornia-apriltag/src/threshold.rs
AS1100K: yup, I removed `tile_size` parameter from `adaptive_threshold` function and instead store it in `TileBuffers` now

crates/kornia-apriltag/src/utils.rs
AS1100K: I did implemented `Pixel` struct and made the adaptive threshold function only support u8 images for now. As you said we can generalise it later.

crates/kornia-apriltag/Cargo.toml
```diff
@@ -14,3 +14,9 @@ rust-version.workspace = true
 version.workspace = true
 
 [dependencies]
+kornia-image = { workspace = true }
+thiserror = { workspace = true }
+kornia-imgproc = { workspace = true }
```
edgarriba, line 18: alphabetical order (maybe we should find a tool to correct toml fles?)

edgarriba: https://crates.io/crates/cargo-toml-lint

crates/kornia-apriltag/src/errors.rs
```diff
@@ -0,0 +1,13 @@
+/// Errors that can occur when working with AprilTag detection.
+#[derive(Debug, thiserror::Error)]
+pub enum AprilTagError {
+    /// Error related to image.
+    #[error(transparent)]
+    ImageError(#[from] kornia_image::error::ImageError),
```
edgarriba, line 6:
```suggestion
    ImageError(#[from] kornia_image::error::ImageError),

```

crates/kornia-apriltag/src/iter.rs
```diff
@@ -0,0 +1,204 @@
+use kornia_image::{allocator::ImageAllocator, Image, ImageSize};
+
+/// An enumerator over tiles of an image, yielding the `(y, x)` tile indices and the tile data as slices.
+/// Each item is a tuple of `((tile_y, tile_x), tile)`, where tile is a slice of rows, and each row is a slice of pixel data.
+#[repr(transparent)]
```
edgarriba: why is this needed ?

crates/kornia-apriltag/src/iter.rs
```diff
@@ -0,0 +1,204 @@
...
+pub struct TileEnumerator<'a, T>(pub TileIterator<'a, T>);
```
edgarriba: why inner needs to be public?

crates/kornia-apriltag/src/iter.rs
```diff
@@ -0,0 +1,204 @@
...
+
+impl<'a, T> Iterator for TileEnumerator<'a, T> {
+    type Item = ((usize, usize), <TileIterator<'a, T> as Iterator>::Item);
+
+    fn next(&mut self) -> Option<Self::Item> {
+        let current_y = self.0.next_tile_y_index;
+        let current_x = self.0.next_tile_x_index;
+
+        if let Some(item) = self.0.next() {
+            return Some(((current_y, current_x), item));
```
edgarriba: should we return better a small private `struct Point2d { x: usize, y: size }` to avoid potenial access issues?

crates/kornia-apriltag/src/iter.rs
```diff
@@ -0,0 +1,204 @@
...
+    ///
+    /// Returns a `TileIterator` that yields non-overlapping tiles of the specified size from the image.
+    /// Tiles at the image edges may be smaller if the image dimensions are not multiples of the tile size.
+    pub fn from_image<const C: usize, A: ImageAllocator>(
+        img: &'a Image<T, C, A>,
+        tile_size: usize,
+    ) -> Self {
+        let img_size = img.size();
+
+        let tiles_x_len = (img_size.width as f32 / tile_size as f32).ceil() as usize;
```
edgarriba: make as separated function `compute_tiles_sizes` or similar name and add unittest with corner cases e.g non-squared  images

crates/kornia-apriltag/src/iter.rs
```diff
@@ -0,0 +1,204 @@
...
+    /// Number of horizontal pixels in the last tile column.
+    last_tile_x_px: usize,
+    /// Number of vertical pixels in the last tile row.
+    last_tile_y_px: usize,
+    /// Index of the next tile to yield horizontally.
+    next_tile_x_index: usize,
+    /// Index of the next tile to yield vertically.
+    next_tile_y_index: usize,
+    /// Buffer holding references to the rows of the current tile.
+    buffer: Vec<&'a [T]>,
```
edgarriba, line 60: why not an owned vec ? another alternative could be to give the tile size as a generic so that the buffer could be an sized array in the stack

crates/kornia-apriltag/src/iter.rs
```diff
@@ -0,0 +1,204 @@
...
+        }
+
+        // Update indices
+        self.next_tile_x_index += 1;
+        if self.next_tile_x_index >= self.tiles_x_len {
+            self.next_tile_x_index = 0;
+            self.next_tile_y_index += 1;
+        }
+
+        Some(unsafe { std::slice::from_raw_parts(self.buffer.as_mut_ptr(), tile_y_px) })
```
edgarriba: can we find a way to avoid unsafe here ?

crates/kornia-apriltag/src/iter.rs
```diff
@@ -0,0 +1,204 @@
...
+    }
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+    use kornia_image::{allocator::CpuAllocator, Image, ImageSize};
+
+    #[test]
+    fn test_tile_iterator() {
```
edgarriba:
```suggestion
    fn test_tile_iterator() -> Result<(), AprilTagError> {
```
and handle properly the errors

crates/kornia-apriltag/src/iter.rs
```diff
@@ -0,0 +1,204 @@
...
+        let data = vec![127u8; 100];
```
edgarriba, line 187: add more meaningful tests with hardcoded values to easily understand if errors occur

crates/kornia-apriltag/src/utils.rs
```diff
@@ -0,0 +1,29 @@
...
+    /// A black pixel.
+    Black = 0,
+    /// A pixel to be skipped.
+    #[default]
+    Skip = 127,
+}
+
+impl Pixel {
+    /// Returns the numeric value of the pixel.
+    pub fn value(&self) -> u8 {
```
edgarriba: i believe that with the `repr` we are already giving the actua value. Maybe add a small unit tes to verify this

crates/kornia-apriltag/src/threshold.rs
```diff
@@ -0,0 +1,445 @@
+use crate::utils::Pixel;
+use crate::{errors::AprilTagError, iter::TileIterator};
+use kornia_image::{allocator::ImageAllocator, Image, ImageError, ImageSize};
+
+/// Stores the minimum and maximum pixel values for each tile for [adaptive_threshold]
+pub struct TileBuffers {
+    tile_min: Vec<u8>,
```
edgarriba: maybe specify in the docs the layout for the order of the tiles ids

crates/kornia-apriltag/src/threshold.rs
```diff
@@ -0,0 +1,445 @@
...
+
+    fn neighbor_blur(&self, current_tile: (usize, usize), tiles_len: (usize, usize)) -> (u8, u8) {
+        let (tiles_y_len, tiles_x_len) = tiles_len;
+        let (y, x) = current_tile;
+        let index = y * tiles_x_len + x;
+
+        let mut neighbor_min = self.tile_min[index];
+        let mut neighbor_max = self.tile_max[index];
+
+        if y != 0 {
```
edgarriba:
```suggestion
        if y > 0 {
```
more intuitive to read that is within the bounds of the image tiles, check the others too

crates/kornia-apriltag/src/threshold.rs
```diff
@@ -0,0 +1,445 @@
...
+        }
+
+        (neighbor_min, neighbor_max)
+    }
+
+    fn neighbor_min_max(
+        &self,
+        neighbor_min: &mut u8,
+        neighbor_max: &mut u8,
+        neighbor_tile: (usize, usize), // (y, x)
```
edgarriba: use a small x/y struct to have better semantics

crates/kornia-apriltag/src/threshold.rs
```diff
@@ -0,0 +1,445 @@
...
+        tiles_x_len: usize,            // (y, x)
```
edgarriba: comment is wrong

crates/kornia-apriltag/src/threshold.rs
```diff
@@ -0,0 +1,445 @@
...
+    ) {
+        let neighbor_index = neighbor_tile.0 * tiles_x_len + neighbor_tile.1;
```
edgarriba: this can be precomputed in the parent function and then for consecutives pixels eg in the x direction just perform index addition to avoid the mul operator every time

crates/kornia-apriltag/src/threshold.rs
```diff
@@ -0,0 +1,445 @@
...
+/// let mut dst = Image::from_size_val(src.size(), Pixel::Skip, CpuAllocator).unwrap();
+///
+/// let mut tile_buffers = TileBuffers::new(src.size(), 2);
+/// adaptive_threshold(&src, &mut dst, &mut tile_buffers, 20).unwrap();
+/// assert_eq!(dst.as_slice(), &[0, 0, 255, 255, 255, 255]);
+/// ```
+// TODO: Add support for parallelism
+pub fn adaptive_threshold<A1: ImageAllocator, A2: ImageAllocator>(
+    src: &Image<u8, 1, A1>,
+    dst: &mut Image<Pixel, 1, A2>,
```
edgarriba, line 188: worth to double check if there's any differenc between having Pixel type than just u8

crates/kornia-apriltag/src/threshold.rs
```diff
@@ -0,0 +1,445 @@
...
+    if tile_buffers.tile_min.len() != expected_tile_count {
+        // It is guaranteed for tile_min and tile_max to have same length by design
+        // so, avoiding additional check for tile_max
+        return Err(AprilTagError::InvalidTileBufferSize(
+            tile_buffers.tile_min.len(),
+            expected_tile_count,
+        ));
+    }
+
+    // let src_data = src.as_slice();
```
edgarriba: remove commented code

crates/kornia-apriltag/src/threshold.rs
```diff
@@ -0,0 +1,445 @@
...
+        if tile.len() != tile_buffers.tile_size || tile[0].len() != tile_buffers.tile_size {
+            // Skip non-full tiles
+            return;
+        }
+
+        let mut local_min = 255;
+        let mut local_max = 0;
+
+        for row in tile {
+            for px in row as &[u8] {
```
edgarriba, line 224: why don't we return the whole tile as a flat slice, then will be easier to use builtin iterators semantics to find min/max in a more efficient way

crates/kornia-apriltag/src/threshold.rs
```diff
@@ -0,0 +1,445 @@
...
+                    local_min = *px;
+                }
+
+                if px > &local_max {
+                    local_max = *px;
+                }
+            }
+        }
+
+        let index = y * tiles_full_x_len + x;
```
edgarriba: i feel the iterator should return a `struct ImageTile { data: &[u8], index: usize }` where the index probably is already a computed value inside the iterator with extra indices that can be used later

crates/kornia-apriltag/src/threshold.rs
```diff
@@ -0,0 +1,445 @@
...
+        .for_each(|((y, x), tile)| {
+            let not_full_tile =
+                tile.len() != tile_buffers.tile_size || tile[0].len() != tile_buffers.tile_size;
+
+            let tile_index = y * tiles_full_x_len + x;
+
+            let (neighbor_min, neighbor_max) = if not_full_tile {
+                if tile.len() != tile_buffers.tile_size && tile[0].len() != tile_buffers.tile_size {
+                    // Bottom-Right tile of image
+                    tile_buffers.neighbor_blur((y - 1, x - 1), (tiles_full_y_len, tiles_full_x_len))
```
edgarriba: why is the interpolation needed here ?

crates/kornia-apriltag/src/threshold.rs
```diff
@@ -0,0 +1,445 @@
...
+                } else if tile[0].len() != tile_buffers.tile_size {
+                    // Right tile of image
+                    tile_buffers.neighbor_blur((y, x - 1), (tiles_full_y_len, tiles_full_x_len))
+                } else {
+                    // Bottom tile of image
+                    tile_buffers.neighbor_blur((y - 1, x), (tiles_full_y_len, tiles_full_x_len))
+                }
+            } else {
+                let (neighbor_min, neighbor_max) =
+                    tile_buffers.neighbor_blur((y, x), (tiles_full_y_len, tiles_full_x_len));
```
edgarriba: same why blur here ? don't see that in the original  code

crates/kornia-apriltag/src/threshold.rs
```diff
@@ -0,0 +1,445 @@
...
+
+                im_min[tile_index] = neighbor_min;
```
edgarriba: i don't get this. Should we use the min/max computed in the previous step per each tile ?

crates/kornia-apriltag/src/iter.rs
```diff
@@ -0,0 +1,204 @@
+use kornia_image::{allocator::ImageAllocator, Image, ImageSize};
+
+/// An enumerator over tiles of an image, yielding the `(y, x)` tile indices and the tile data as slices.
+/// Each item is a tuple of `((tile_y, tile_x), tile)`, where tile is a slice of rows, and each row is a slice of pixel data.
+#[repr(transparent)]
+pub struct TileEnumerator<'a, T>(pub TileIterator<'a, T>);
+
+impl<'a, T> Iterator for TileEnumerator<'a, T> {
```
edgarriba: why can't you implement the Iterator directly on the TileIterator ? less confusing naming

crates/kornia-apriltag/src/threshold.rs
```diff
@@ -0,0 +1,445 @@
...
+                im_max[tile_index] = neighbor_max;
+
+                if neighbor_max - neighbor_min < min_white_black_diff {
+                    for y_px in 0..tile.len() {
+                        let row = ((y * tile_buffers.tile_size) + y_px) * src.width();
```
edgarriba: can't we reuse somehow all this from the iterator ? we are recomputing indices to many times

edgarriba: maybe as suggested before return the needed info in a struct

crates/kornia-apriltag/src/threshold.rs
```diff
@@ -0,0 +1,445 @@
...
+                    return;
+                }
+
+                (neighbor_min, neighbor_max)
+            };
+
+            let thresh = neighbor_min + (neighbor_max - neighbor_min) / 2;
+
+            for (y_px, row) in tile.iter().enumerate() {
+                let row_index = ((y * tile_buffers.tile_size) + y_px) * src.width()
```
edgarriba: same, get this info from the iterator to avoid computing again

crates/kornia-apriltag/src/threshold.rs
```diff
@@ -0,0 +1,445 @@
...
+        assert_eq!(neighbor_max, 70);
+    }
+
+    #[test]
+    fn test_adaptive_threshold_basic() {
+        #[rustfmt::skip]
+        let src = Image::new(
+            ImageSize {
+                width: 5,
+                height: 5,
```
edgarriba: test with non suared images

crates/kornia-apriltag/src/threshold.rs
```diff
@@ -0,0 +1,445 @@
...
+            vec![
+                0,   50,  100, 150, 200,
+                250, 0,   50,  100, 150,
+                200, 250, 0,   50,  100,
+                150, 200, 250, 0,   50,
+                100, 150, 200, 250, 0,
+            ],
+            CpuAllocator,
+        )
+        .unwrap();
```
edgarriba: hande errors

crates/kornia-apriltag/src/errors.rs
AS1100K: You might have forgot to edit the suggestion code

crates/kornia-apriltag/src/iter.rs
AS1100K: will do the same

crates/kornia-apriltag/src/iter.rs
AS1100K: I have implemented it on `TileIterator`, this enumerator just also returns the x and y values of the tile according to image height and width

crates/kornia-apriltag/src/iter.rs
AS1100K: without this we can't satisfy the lifetime

crates/kornia-apriltag/src/threshold.rs
AS1100K: Currently `Pixel` is repr(u8), so it's just like u8 in memory while also ensuring the value can be 127, 255, and 0 only.

crates/kornia-apriltag/src/threshold.rs
AS1100K: Returning the flat slice of fragmented data would involve moving data from the image to buffer, but by returning the slice of tile rows we are returning the fragmented data in a much efficient manner

crates/kornia-apriltag/src/threshold.rs
AS1100K: Forgot to remove it, I used this for debugging the values and compare it with the C code. I will remove this

crates/kornia-apriltag/src/threshold.rs
AS1100K: It's there in the original code, but the approach is different. In the C code, they are storing all the blur values of only full sized tiles into a array, and for the full-sized tiles. For the remaining non-full sized tile, they use the blur value of it's nearest full sized tile.
Instead of reading the image 3 times (as in the C code), we are only reading those 2 times and computing there blur again as it is not that expensive operation cause we are only computing 5 neighbors of the nearest tile

crates/kornia-apriltag/src/errors.rs
edgarriba: It’s a new line suggestion

AS1100K: okay

crates/kornia-apriltag/src/threshold.rs
```diff
@@ -0,0 +1,442 @@
...
+                }
+            }
+        }
+
+        tile_buffers.tile_min[tile.full_index] = local_min;
+        tile_buffers.tile_max[tile.full_index] = local_max;
+    });
+
+    // Binarize the image
+    TileIterator::from_image(src, tile_buffers.tile_size)?.for_each(|tile| {
```
edgarriba: you cannot use the same iterator declared before ?

edgarriba (approved):

crates/kornia-apriltag/src/threshold.rs
AS1100K: We can't as the `for_each` function consumes the `TileIterator`

crates/kornia-apriltag/src/iter.rs
```diff
@@ -0,0 +1,285 @@
...
+///
+/// Each item yielded by the iterator is a slice of rows, where each row is a slice of pixel data.
+/// The tile size is specified at construction, and the iterator will yield tiles of the given size,
+/// except for tiles at the image edges, which may be smaller if the image dimensions are not
+/// multiples of the tile size.
+pub struct TileIterator<'a, T> {
+    img_data: &'a [T],
+    img_size: ImageSize,
+    tile_size: usize,
+    tiles_len: Point2d,
```
cjpurackal: `tiles_len` suggests a scalar while this is a `Point2d`. For readability, I suggest renaming it to something like `tiles_dim`

crates/kornia-apriltag/src/iter.rs
```diff
@@ -0,0 +1,285 @@
...
+
+        // Update indices
+        self.next_tile_index.x += 1;
+        if self.next_tile_index.x >= self.tiles_len.x {
+            self.next_tile_index.x = 0;
+            self.next_tile_index.y += 1;
+        }
+
+        self.next_index += 1;
+        let data = unsafe { std::slice::from_raw_parts(self.buffer.as_mut_ptr(), tile_y_px) };
```
cjpurackal: @edgarriba any thoughts on this?

crates/kornia-apriltag/src/threshold.rs
```diff
@@ -0,0 +1,442 @@
+use crate::iter::ImageTile;
+use crate::utils::{find_full_tiles, Pixel, Point2d};
+use crate::{errors::AprilTagError, iter::TileIterator};
+use kornia_image::{allocator::ImageAllocator, Image, ImageError, ImageSize};
+
+/// Stores the minimum and maximum pixel values for each tile for [adaptive_threshold]
+///
+/// The tiles are indexed in row-major order, i.e., tile IDs increase first along the x-axis (columns),
+/// then along the y-axis (rows) only for the full tiles.
+pub struct TileBuffers {
```
cjpurackal: Maybe change the name to  `TileExtrema` or `TileMinMax` for better readability ?

cjpurackal:
minor comments, other than that LGTM

crates/kornia-apriltag/src/threshold.rs
edgarriba: +1

crates/kornia-apriltag/src/iter.rs
```diff
@@ -0,0 +1,285 @@
...
+
+        // Update indices
+        self.next_tile_index.x += 1;
+        if self.next_tile_index.x >= self.tiles_dim.x {
+            self.next_tile_index.x = 0;
+            self.next_tile_index.y += 1;
+        }
+
+        self.next_index += 1;
+        let data = unsafe { std::slice::from_raw_parts(self.buffer.as_ptr(), tile_y_px) };
```
edgarriba, line 157: Oh yes, right, let’s find a way to avoid unsafe code

andrew-shc (approved):

PR #387, opened 2025-06-08
Implement Zero copy images from gstreamer

edgarriba:
merge of https://github.com/kornia/kornia-rs/pull/338

- implement a GstAllocator for gstreamer pipelines that holds the `gst::Buffer`
- update examples to use the allocator
- add benchesfor gstreamer


crates/kornia-io/benches/bench_gstreamer.rs
```diff
@@ -0,0 +1,52 @@
+use criterion::{black_box, Criterion};
```
AS1100K:
```suggestion
use criterion::Criterion;
use std::hint::black_box;
```

crates/kornia-io/src/stream/capture.rs
```diff
@@ -131,30 +131,48 @@ impl StreamCapture {
...
+            return Ok(None);
+        };
+
+        // unpack the frame buffer
+        let width = frame_buffer.width;
+        let height = frame_buffer.height;
+        let buffer = frame_buffer.buffer;
+
+        let mapped_buffer = buffer
+            .into_mapped_buffer_readable()
```
AS1100K, line 153: We are converting it into `MappedBuffer<Readable>`, while our mutability test later works. I think that in some cases we could have an undefined behavior when trying to mutate the value of underlying slice.

edgarriba: i'm getting an error with the writtable map 

```bash
(webcam:262907): GStreamer-CRITICAL **: 20:09:55.691: write map requested on non-writable buffer
Error: GetBufferError
```

AS1100K: Then, maybe we should just say in the grab docs that the image should mutated with caution as it might cause undefined behavior.
Although I did tried the mutation on webcam example and it works, not sure in what case this would cause undefined behavior.
Or, we could look into something that would enforce the image to be immutable (maybe do this later)

edgarriba: image

quick experiment seems to be alright

AS1100K: Should we also remove the mutability test?

edgarriba: done

PR #396, opened 2025-06-16
lie: fix col major issues and relax aarch64 epsilon tests

edgarriba:
- fix wrong column major access members
- relax tests for aarch64 with higher epsilon

### DIFF
--- a/crates/kornia-lie/src/se2.rs
+++ b/crates/kornia-lie/src/se2.rs
@@ -150,7 +150,7 @@ mod tests {
     const EPSILON: f32 = 1e-6;
 
     #[cfg(target_arch = "aarch64")]
-    const EPSILON: f32 = 1e-3;
+    const EPSILON: f32 = 1e-2;
 
     fn make_random_se2() -> SE2 {
         SE2::from_random()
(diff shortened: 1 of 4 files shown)


crates/kornia-lie/src/se2.rs
```diff
@@ -81,24 +81,32 @@ impl SE2 {
...
         };
+
         let v_inv = Mat2::from_cols_array(&[a, -half_theta, half_theta, a]);
-        let upsilon = v_inv * self.t;
+        let upsilon = v_inv.mul_vec2(self.t);
 
         (upsilon, theta)
     }
 
     pub fn hat(upsilon: Vec2, theta: f32) -> Mat3A {
```
andrew-shc: should we just use a uniform vector `v: Vec3` for the parameter?
https://github.com/kornia/kornia/blob/a63f4f0f44da0b89b276cbc81ea467ec181d5ca2/kornia/geometry/liegroup/se2.py#L226

andrew-shc:
I just realized there's some inconsistencies here:
https://github.com/kornia/kornia-rs/blob/be3e16eb77edb265988f075eede641a12def44e7/crates/kornia-lie/src/se3.rs#L27
vs.
https://github.com/kornia/kornia-rs/blob/be3e16eb77edb265988f075eede641a12def44e7/crates/kornia-lie/src/se2.rs#L24

crates/kornia-lie/src/se2.rs
edgarriba: done

PR #398, opened 2025-06-16
vlm: implement paligemma using candle

edgarriba:
import paligemma from https://github.com/kornia/kornia-paligemma

### DIFF
--- a/crates/kornia-vlm/src/paligemma/mod.rs
+++ b/crates/kornia-vlm/src/paligemma/mod.rs
@@ -1,0 +1,176 @@
+mod model;
+mod utils;
+
+use candle_core::{DType, Device, Tensor};
+use candle_nn::VarBuilder;
+use candle_transformers::models::paligemma::{Config, Model};
+use hf_hub::{api::sync::Api, Repo, RepoType};
+use kornia_image::{allocator::CpuAllocator, Image};
+use kornia_imgproc::{interpolation::InterpolationMode, resize::resize_fast};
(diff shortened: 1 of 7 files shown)


andrew-shc (approved):
how are we going to test the models in the future? or it's just impractical. i assume we would be testing the individual low-level components in the `kornia-nn` crate

edgarriba:
yep, possibly at individual level. Also, something that worked in the past with the pytorch library, is just to have synthetic tests to verify determinism.

crates/kornia-vlm/src/paligemma/mod.rs
```diff
@@ -0,0 +1,176 @@
...
+            .expect("Failed to generate text");
+
+        Ok(response)
+    }
+
+    // utility function to load the model
+    fn load_model(dtype: DType, device: &Device) -> Result<(Model, Tokenizer), PaligemmaError> {
+        let api = Api::new()?;
+        // TODO: add a way to load the model from a local path
+        let model_id = "google/paligemma-3b-mix-224".to_string();
```
cjpurackal, line 158: if we make this dynamic (maybe part of `PaligemmaConfig` ?), Maybe we can support other models like [google/paligemma-3b-mix-44](https://huggingface.co/google/paligemma-3b-mix-448)

edgarriba: sure, let me open issue for later PR

PR #401, opened 2025-06-17
feat: add se3 jacobains and fix log->exp->log roundtrip issue

cjpurackal:


edgarriba:
@cjpurackal can you relax the epsilon for the failing test ? I tried to find out the root cause but couldn't figure out the reason at all

cjpurackal:
@edgarriba The latest commit should fix the issue. The reason was that we were using an approximation of the SE2 Log fn, so for certain angles the the `roundtrip` test (`log->exp->log`) would fail because it drifts outside of the tolerance.

PR #406, opened 2025-06-19
Add Connected Components for AprilTag detection

AS1100K:
Part 1 of the Segmentation Process


crates/kornia-apriltag/src/segmentation.rs
```diff
@@ -0,0 +1,106 @@
...
+        let mut uf = find_connected_components(&bin);
+
+        let mut union_representatives = String::new();
+        let expected =
+            std::fs::read_to_string("../../tests/data/apriltag_pixel_representatives.txt")?;
+
+        for y in 0..src.height() {
+            for x in 0..src.width() {
+                let i = y * src.width() + x;
+                let representative = if x == src.width() - 1 { i } else { uf.find(i) }.to_string();
```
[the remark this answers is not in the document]
AS1100K: ~The C implementation assumes that the last pixel of each row has it's left pixel as the representative by how the union is created, and sets are formed~

Outdated

crates/kornia-apriltag/src/segmentation.rs
```diff
@@ -0,0 +1,151 @@
...
+        let bin_data = vec![
+            Black, Black, White, White,
+            Black, Black, White, White,
+            White, White, Black, Black,
+            White, White, Black, Black
+        ];
+
+        let bin = Image::new(
+            ImageSize {
+                width: 4,
```
edgarriba: test with non squared images

crates/kornia-apriltag/src/union_find.rs
```diff
@@ -0,0 +1,118 @@
...
+pub struct UnionFind {
+    parent: Vec<usize>,
+    size: Vec<usize>,
+}
+
+impl UnionFind {
+    /// Creates a new UnionFind structure with length `len`.
+    pub fn new(len: usize) -> Self {
+        Self {
+            parent: vec![usize::MAX; len],
```
edgarriba, line 11:
```suggestion
            parent: vec![usize::MAX; (len + 1) * 2],
```
i don't think initialization is correct. And then only the first half of the array is set to 0xff aka usize::MAX
https://github.com/AprilRobotics/apriltag/blob/master/common/unionfind.h#L51

similar then for the size vec

crates/kornia-apriltag/src/union_find.rs
```diff
@@ -0,0 +1,118 @@
...
+            return id;
+        }
+
+        // Chase down the root
+        while self.parent[root] != root {
+            root = self.parent[root];
+        }
+
+        // Go back and collapse the tree
+        let mut id = id;
```
edgarriba: why don't you define directly as mut in the function signature?

crates/kornia-apriltag/src/union_find.rs
```diff
@@ -0,0 +1,118 @@
...
+    /// Unites the sets containing `aid` and `bid`.
+    pub fn union(&mut self, aid: usize, bid: usize) {
+        let aroot = self.get_representative(aid);
+        let broot = self.get_representative(bid);
+
+        if aroot == broot {
+            return;
+        }
+
+        let asize = self.size[aroot];
```
edgarriba, line 49:
```suggestion
        let asize = self.size[aroot] + 1;
```
https://github.com/AprilRobotics/apriltag/blob/master/common/unionfind.h#L128

crates/kornia-apriltag/src/union_find.rs
```diff
@@ -0,0 +1,118 @@
...
+        root
+    }
+
+    /// Unites the sets containing `aid` and `bid`.
+    pub fn union(&mut self, aid: usize, bid: usize) {
+        let aroot = self.get_representative(aid);
+        let broot = self.get_representative(bid);
+
+        if aroot == broot {
+            return;
```
edgarriba:
```suggestion
            return aroot;
```

crates/kornia-apriltag/src/union_find.rs
```diff
@@ -0,0 +1,118 @@
...
+        }
+
+        let asize = self.size[aroot];
+        let bsize = self.size[broot];
+
+        if asize > bsize {
+            self.parent[broot] = aroot;
+            self.size[aroot] += bsize;
```
edgarriba, line 54:
```suggestion
            self.size[aroot] += bsize;
            return aroot;
```

crates/kornia-apriltag/src/union_find.rs
```diff
@@ -0,0 +1,118 @@
...
+        } else {
+            self.parent[aroot] = broot;
+            self.size[broot] += asize;
```
edgarriba, line 58:
```suggestion
            self.size[broot] += asize;
            return broot;
```

crates/kornia-apriltag/src/segmentation.rs
```diff
@@ -0,0 +1,151 @@
...
+                height: 4,
+            },
+            bin_data,
+            CpuAllocator,
+        )?;
+
+        let mut uf = UnionFind::new(16);
+        find_connected_components(&bin, &mut uf)?;
+
+        assert_eq!(uf.get_representative(0), 0);
```
edgarriba, line 105: shouldn't the results give only 4 regions ?

crates/kornia-apriltag/src/union_find.rs
```diff
@@ -0,0 +1,118 @@
...
+            let tmp = self.parent[id];
+            self.parent[id] = root;
+            id = tmp;
+        }
+
+        root
+    }
+
+    /// Unites the sets containing `aid` and `bid`.
+    pub fn union(&mut self, aid: usize, bid: usize) {
```
edgarriba:
```suggestion
    pub fn cpnnect(&mut self, aid: usize, bid: usize) {
```

crates/kornia-apriltag/src/segmentation.rs
AS1100K: No, we skip the fist and last pixel and assume it to be itself representative, for the rest of the pixels we consider the left and the top pixels. And for the diagonal we only consider them if the current pixel is White. By this rules, it is expected for it to have more than 4 regions

crates/kornia-apriltag/src/union_find.rs
AS1100K: In the C code, they initialize the array in one go for twice the length and consider the first half for parent and the second half to be size array, that's why they set 0xff for the first half and 0 for the rest

crates/kornia-apriltag/src/union_find.rs
AS1100K: Since, we are initializing the size array with default value 1, we don't need to +1 here as the reason to do so, is to avoid subtract with overflow. I verified it without having +1, and the value initialized with 1 instead of 0, and it works as same.

In this way, we avoid unnecessary addition time in every union

cjpurackal (approved):

PR #410, opened 2025-06-22
Add gradient cluster detection

AS1100K:
Final part 2 of segmentation process

## TODOs

- [x] Add gradient cluster detection
- [x] Code cleanup
- [x] Add deterministic test


crates/kornia-apriltag/src/segmentation.rs
```diff
@@ -72,6 +81,156 @@ pub fn find_connected_components<A: ImageAllocator>(
...
+/// * `clusters` - Mutable reference to a map where the gradient information between component pairs will be stored.
+///     Make sure to call [`HashMap::clear`] if you are using this function multiple times with the same `clusters`
+pub fn find_gradient_clusters<A: ImageAllocator>(
+    src: &Image<Pixel, 1, A>,
+    uf: &mut UnionFind,
+    clusters: &mut HashMap<(usize, usize), Vec<GradientInfo>>,
+) {
+    let src_slice = src.as_slice();
+
+    (1..src.height() - 1).into_iter().for_each(|y| {
```
cjpurackal: `into_iter` is redundant here

crates/kornia-apriltag/src/segmentation.rs
```diff
@@ -72,6 +81,156 @@ pub fn find_connected_components<A: ImageAllocator>(
...
+        let mut connected_last = false;
+
+        (1..src.width() - 1).into_iter().for_each(|x| {
```
cjpurackal: same here

crates/kornia-apriltag/src/segmentation.rs
edgarriba: maybe just a regular for / range expression ?

crates/kornia-apriltag/src/segmentation.rs
```diff
@@ -72,6 +81,156 @@ pub fn find_connected_components<A: ImageAllocator>(
...
+            let i = y * src.width() + x;
+            let current_pixel = src_slice[i];
+
+            if current_pixel == Pixel::Skip {
+                connected_last = false;
+                return;
+            }
+
+            let current_pixel_representative = uf.get_representative(i);
+            if uf.get_set_size(current_pixel_representative) < 25 {
```
edgarriba, line 174: add comment why this 25 here

crates/kornia-apriltag/src/segmentation.rs
```diff
@@ -72,6 +81,156 @@ pub fn find_connected_components<A: ImageAllocator>(
...
+            let mut do_conn =
+                |dx: isize, dy: isize, neighbor_i: usize, any_connected: &mut bool| {
+                    let neighbor_pixel = src_slice[neighbor_i];
+                    if neighbor_pixel == Pixel::Skip {
+                        return;
+                    }
+
+                    if current_pixel != neighbor_pixel {
+                        let neighbor_pixel_representative = uf.get_representative(neighbor_i);
+                        if uf.get_set_size(neighbor_pixel_representative) > 24 {
```
edgarriba: why here 24 ?

crates/kornia-apriltag/src/segmentation.rs
edgarriba: or wondering if just with iterators directly on the slice would work using proper skip / take semnatics

AS1100K: I have made the range expression directly call `for_each`.

Using just standard lib iterators, would not provide us x, and y coordinates of the pixels. We could use `TileIterator`, but this would be just overkill and IMO slower for tile size 1

crates/kornia-apriltag/src/union_find.rs
```diff
@@ -58,23 +58,24 @@ impl UnionFind {
...
     pub fn reset(&mut self) {
-        (0..self.parent.len()).into_iter().for_each(|i| {
+        (0..self.parent.len()).for_each(|i| {
             self.parent[i] = usize::MAX;
             self.size[i] = 1;
         });
     }
 
     /// Returns the number of elements in the UnionFind structure.
+    #[allow(clippy::len_without_is_empty)]
```
edgarriba, line 79: maybe just provide the `is_empty` method too

AS1100K: Both the vectors in `UnionFind` will never be empty but will contain default values, so there `Vec::is_empty` will always return `false`. So, IMO `is_empty` is reductant

edgarriba: ok, add this statement as a comment then

PR #414, opened 2025-06-26
deprecate image-rs crate

edgarriba:


cjpurackal:
Is this just aimed at the `kornia-io` crate? or do we plan to remove it from other crates as well, like `kornia-imgproc` ?

cjpurackal (approved):

edgarriba:
The image crate it was only used in kornia-io , the rest is for testing which it’s totally fine

PR #419, opened 2025-06-28
Add Quad Fitting

AS1100K:
Screenshot from 2025-06-30 04-54-00

## TODOs

- [x] Add Initial Quad Fitting Logic
- [x] Add Tests
- [x] Code Cleanup
- [x] Add Docs

### DIFF
--- a/crates/kornia-apriltag/src/quad.rs
+++ b/crates/kornia-apriltag/src/quad.rs
@@ -5,8 +5,10 @@ use std::{collections::HashMap, f32};
 /// TODO
 #[derive(Debug, Default, Clone)]
 pub struct Quad {
-    point: [[f32; 2]; 4],
-    // reversed_bool: bool,
+    /// TODO
+    pub point: [[f32; 2]; 4],
+    /// TODO
+    pub reversed_border: bool,
     // h: Vec<f32>,
     // hinv: Vec<f32>,
 }
@@ -25,7 +27,7 @@ pub fn fit_quads<A: ImageAllocator>(
 
     let mut quads = Vec::new();
 
-    clusters.iter_mut().for_each(|(_, mut cluster)| {
+    clusters.iter_mut().for_each(|(_, cluster)| {
         if cluster.len() < min_cluster_pixels {
             return;
         }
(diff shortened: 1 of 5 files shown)


crates/kornia-apriltag/src/quad.rs
```diff
@@ -0,0 +1,1052 @@
...
+                dx = -dx;
+            }
+
+            if dx < 0.0 {
+                let tmp = dx;
+                dx = dy;
+                dy = -tmp;
+            }
+
+            *slope = quadrant as f32 + dy / dx;
```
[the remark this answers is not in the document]
AS1100K: the `dx` can never be zero, even if the `pos.x` is zero, we are also subtracting a small number which results in the value to be never zero as we know that `pos.x` is a integer

edgarriba (approved):

crates/kornia-apriltag/src/quad.rs
```diff
@@ -0,0 +1,1059 @@
...
+            lfps,
+            (i + len - window_size) % len,
+            (i + window_size) % len,
+            None,
+            errors.get_mut(i),
+            None,
+        );
+    });
+
+    const SIGMA: f32 = 1.0;
```
cjpurackal, line 438: Can you check if we can use the guassian kernel defined here: 
https://github.com/kornia/kornia-rs/blob/171965ad2f41042c49ad21f904eb41afdda6e3c6/crates/kornia-imgproc/src/filter/kernels.rs#L25

crates/kornia-apriltag/src/quad.rs
```diff
@@ -0,0 +1,1059 @@
...
+
+        if pos.y > y_max {
+            y_max = pos.y;
+        } else if pos.y < y_min {
+            y_min = pos.y;
+        }
+    });
+
+    // Reject clusters whose bounding box is too small
+    if (x_max - x_min) * (y_max - y_min) < min_tag_width {
```
cjpurackal, line 152: `min_tag_width` is being compared to an area, so change name to `min_tag_area`

AS1100K: `min_tag_width` actually represents the width not the area, as we are later finding the area of tag at L291. We are comparing area with the width so that we can exit early for very small quads, as their area after accounting for rotation is calculated at later step.

cjpurackal (approved):

PR #420, opened 2025-07-01
Add Apriltag Decoding

AS1100K:
⚠️ **DO NOT MERGE** ⚠️
_Currently this PR is a successor of #419 and once that is merged, I will cherry pick commits and change the base of the PR to main_

## TODOs

- [x] Implement Tag Decoding
- [ ] Add tests
- [ ] Code Cleanup
- [ ] Add Docs


crates/kornia-apriltag/src/utils.rs
```diff
@@ -63,6 +63,163 @@ impl PartialEq<u8> for Pixel {
...
+            sum += a[col * 9 + i] * a[i * 9 + 8];
+        }
+        a[col * 9 + 8] = (a[col * 9 + 8] - sum) / a[col * 9 + col];
+    }
+
+    Some([a[8], a[17], a[26], a[35], a[44], a[53], a[62], a[71], 1.0])
+}
+
+/// TODO
+pub const fn inverse_3x3_matrix(a: &[f32; 9]) -> [f32; 9] {
```
SeaOtocinclus: Why not using an existing framework for such common functions?

edgarriba: We are trying to go with as less deps as possible in the first phase, there will be be a second optimisation phase which if we find a lib that make a difference then yes (eg glam). We tried in other parts faer and wasnˋt performing that well for small matrices 3x3, 4x4. See https://github.com/kornia/kornia-rs/pull/241#discussion_r1984719826

In a later stage we might port all this small functionality under kornia-linalg or similar to experiment more with glam or wide

crates/kornia-apriltag/src/decode.rs
```diff
@@ -0,0 +1,1161 @@
...
+
+/// Applies a sharpening filter to the input values using a Laplacian kernel.
+///
+/// # Arguments
+///
+/// * `sharpening_buffer` - Mutable reference of `SharpeningBuffer`.
+/// * `decode_sharpening` - The sharpening factor to apply.
+fn sharpen(sharpening_buffer: &mut SharpeningBuffer, decode_sharpening: f32) {
+    #[rustfmt::skip]
+    const KERNEL: [f32; 9] = [
```
cjpurackal: I'm thinking if this should be added to `kornia-imgproc` and used from there ?

crates/kornia-apriltag/src/decode.rs
```diff
@@ -0,0 +1,1161 @@
...
+            (0..3).for_each(|i| {
+                (0..3).for_each(|j| {
+                    let yi = y + i;
+                    let xj = x + j;
+                    if yi == 0
+                        || xj == 0
+                        || (yi - 1) > sharpening_buffer.size - 1
+                        || (xj - 1) > sharpening_buffer.size - 1
+                    {
+                        return;
```
cjpurackal: continue?

AS1100K: We are using `Iterator::for_each` which takes a closure. Here return behaves like continue

crates/kornia-apriltag/src/decode.rs
cjpurackal, line 782: I think this can be added to the

crates/kornia-apriltag/src/family/mod.rs
```diff
@@ -0,0 +1,22 @@
...
+    /// The total width of the tag, including border, in units.
+    pub total_width: usize,
+    /// The number of bits in the tag code.
+    pub nbits: usize,
+    /// The x-coordinates of each bit in the tag.
+    pub bit_x: &'static [u8],
+    /// The y-coordinates of each bit in the tag.
+    pub bit_y: &'static [u8],
+    /// The code data for the tag family.
+    pub code_data: &'static [usize],
```
edgarriba: probably is worth to have owned data here as result and we avoid lifetime parameters

crates/kornia-apriltag/src/family/mod.rs
```diff
@@ -0,0 +1,22 @@
...
+    // TODO: more properties
+}
+
+#[doc(hidden)]
```
edgarriba, line 50: why hidden docs here ?

crates/kornia-apriltag/src/lib.rs
```diff
@@ -24,3 +24,6 @@ pub mod family;
 
 /// Quad detection utilities for AprilTag detection.
 pub mod quad;
+
+/// Decoding utilities for AprilTag detection.
+pub mod decode;
```
edgarriba:
```suggestion
pub mod decoder;
```
?

crates/kornia-apriltag/src/quad.rs
```diff
@@ -38,14 +38,39 @@ impl Default for FitQuadOpts {
...
+#[derive(Debug, Default, Clone, PartialEq)]
 pub struct Quad {
     /// The four corners of the quadrilateral, in image coordinates.
     ///
     /// Order: [Bottom-left, Bottom-right, Top-right, Top-left]
     pub corners: [Point2d<f32>; 4],
     /// Indicates whether the border is reversed (black border inside white border).
     pub reversed_border: bool,
+    /// The 3x3 homography matrix (row-major order) mapping tag coordinates to image coordinates.
+    pub homography: [f32; 9],
```
edgarriba, line 51: add as todo to use glam 3x3 mat representation

crates/kornia-apriltag/src/quad.rs
```diff
@@ -38,14 +38,39 @@ impl Default for FitQuadOpts {
...
+    ///
+    /// # Arguments
+    ///
+    /// * `x` - The x coordinate in tag space.
+    /// * `y` - The y coordinate in tag space.
+    ///
+    /// # Returns
+    ///
+    /// A `Point2d<f32>` representing the projected point in image coordinates.
+    pub fn homography_project(&self, x: f32, y: f32) -> Point2d<f32> {
```
edgarriba, line 65: add todo to use glam 3x3

crates/kornia-apriltag/src/utils.rs
edgarriba, line 1: add some notes here wether to place this under a more general purpose utilities, lets say kornia-linalg or wherever we decide later

crates/kornia-apriltag/src/utils.rs
```diff
@@ -63,9 +63,164 @@ impl PartialEq<u8> for Pixel {
...
+///
+/// # Arguments
+///
+/// * `src` - The source grayscale image.
+/// * `p` - The floating-point pixel coordinate.
+///
+/// # Returns
+///
+/// An `Option<f32>` containing the interpolated pixel value, or `None` if the coordinate is out of bounds.
+pub(crate) fn value_for_pixel<A: ImageAllocator>(
```
edgarriba, line 192: we have an interpolation routine in imgproc. Check if we can reuse

crates/kornia-apriltag/src/family/tag36h11.rs
```diff
@@ -0,0 +1,610 @@
+use super::*;
+
+impl TagFamily {
+    /// The Tag36H11 AprilTag family.
+    pub const TAG36_H11: Self = Self {
+        width_at_border: 8,
+        reversed_border: false,
+        total_width: 10,
+        nbits: 36,
+        bit_x: &[
```
edgarriba: use owned data, i don't see any value here to have referenced

crates/kornia-apriltag/src/family/tag36h11.rs
```diff
@@ -0,0 +1,610 @@
+use super::*;
+
+impl TagFamily {
+    /// The Tag36H11 AprilTag family.
+    pub const TAG36_H11: Self = Self {
```
edgarriba: an alternaitve could be to have a factory method:

```rust
pub fn new(TagFamilyType: type) -> Self { }
```

and use enum instead that implements a constuctor with the const values ?

crates/kornia-apriltag/src/decode.rs
```diff
@@ -0,0 +1,1161 @@
...
+        });
+    }
+}
+
+/// Options for decoding tags from an image.
+///
+/// This struct contains configuration parameters and precomputed data
+/// required for decoding tags, such as the tag family, quick decode table,
+/// edge refinement flag, and sharpening factor.
+pub struct DecodeTagsOpts<'a> {
```
edgarriba: better if we can avoid the lifetimeparam

edgarriba: also for consistency with the rest of the library let's agree if `config` structs we call it `FooBarOpts` or `FooBarConfing` -- i've been using the later

crates/kornia-apriltag/src/decode.rs
```diff
@@ -0,0 +1,1161 @@
...
+                quad.homography = matrix_3x3_mul(&quad.homography, &r);
+                let center = quad.homography_project(0.0, 0.0);
+
+                let detection = Detection {
+                    tag_family: opts.tag_family,
+                    id: entry.id,
+                    hamming: entry.hamming,
+                    decision_margin,
+                    center,
+                    quad: std::mem::take(quad), // TODO: Should we take or copy the value? Benchmark it
```
edgarriba: as end user api makes sense to return owned data for the results

crates/kornia-apriltag/src/decode.rs
```diff
@@ -0,0 +1,1161 @@
...
+    detections
+}
+
+/// Refines the edges of a quadrilateral in the image by adjusting its corners based on local image gradients.
+///
+/// # Arguments
+///
+/// * `src` - Reference to the grayscale source image.
+/// * `quad` - Mutable reference to the quadrilateral to refine.
+fn refine_edges<A: ImageAllocator>(src: &Image<u8, 1, A>, quad: &mut Quad) {
```
edgarriba: add todo to move this somewhere in imgproc

crates/kornia-apriltag/src/decode.rs
```diff
@@ -0,0 +1,1161 @@
...
+/// Updates the homography matrix for the given quadrilateral based on its corners.
+///
+/// # Arguments
+///
+/// * `quad` - Mutable reference to the quadrilateral whose homography will be updated.
+///
+/// # Returns
+///
+/// Returns `true` if the homography was successfully computed and updated, `false` otherwise.
+fn quad_update_homographies(quad: &mut Quad) -> bool {
```
edgarriba: feels like this should be a method from Quad

crates/kornia-apriltag/src/decode.rs
edgarriba: yes, and see if we can reuse some of the filtering ops

crates/kornia-apriltag/src/decode.rs
```diff
@@ -0,0 +1,1161 @@
...
+    #[rustfmt::skip]
+    const KERNEL: [f32; 9] = [
+         0.0, -1.0,  0.0,
+        -1.0,  4.0, -1.0,
+         0.0, -1.0,  0.0,
+    ];
+
+    (0..sharpening_buffer.size).for_each(|y| {
+        (0..sharpening_buffer.size).for_each(|x| {
+            let idx = y * sharpening_buffer.size + x;
```
edgarriba:
```rust
let idx = y * sharpening_buffer.size ;
```
can be in the loop above

crates/kornia-apriltag/src/family/tag36h11.rs
AS1100K: maybe let's just look into High Level APIs later once we have implemented multiple tag families

crates/kornia-apriltag/src/utils.rs
AS1100K: The interpolate function in the imgproc doesn't support u8 images, and converting the image to f32 slice might not be ideal.
Maybe let's find a way in imgproc to support multiple types as many other functions are also restricting just due to the slice type in another PR, I will add a TODO comment for now

crates/kornia-apriltag/src/decode.rs
AS1100K: By having it as a reference we are avoiding copy as later we are also returning the same reference in the detection which will be valid for as long as tag family remains, otherwise copying TagFamily contents could greatly impact performance

crates/kornia-apriltag/src/decode.rs
AS1100K: I don't think that the imgproc crate have Laplacian Filter operations

crates/kornia-apriltag/src/family/mod.rs
AS1100K: For now, the modules are mostly empty just and just the `CODE_DATA` is public, for the tag it's an impl. I am just doing this for now, as this is the intermediate stage, and the family will be changed in the next PR adding multiple familes.

So, I though since there isn't a lot of stuff here maybe hide it for now and later work on it

crates/kornia-apriltag/src/decoder.rs
```diff
@@ -0,0 +1,1134 @@
...
+        self.0[bucket].id = id;
+        self.0[bucket].hamming = hamming;
+    }
+}
+
+/// Represents a detected tag in the image, including its family, ID, decoding quality, and geometric information.
+#[derive(Debug, Clone, PartialEq)]
+pub struct Detection<'a> {
+    /// Reference to the tag family this detection belongs to.
+    pub tag_family: &'a TagFamily,
```
edgarriba: why a reference if the data is owned inside the tagfamily ? as an alternative i think we should just retturn an enum and if the tag family data is needed, the user can get it easily

AS1100K: We are returning the tag family as once we support multiple tags, the detection could detect multiple tag families and having a reference will let them later distinguish between the tags identified.

Having an enum would only work for predefined tag families, but for custom tags the enum would look something like:

```rust
enum TagFamily<'a> {
    Tag36H11,
    /* Other Families */
    Custom(TagFamilySpecs<'a>)
}
```
which would still include the lifetime and returning the owned data wouldn't be efficient as it is large in size.

edgarriba: custom could contain a string or similar as uuid where the actual values could come from a factory method matching that  id

PR #421, opened 2025-07-01
dora: use upstream dora nodes for the examples

edgarriba:


examples/dora/dataflow.yml
```diff
@@ -1,5 +1,7 @@
 nodes:
   - id: v4l-camera
+    git: https://github.com/kornia/dora-nodes-hub.git
+    rev: 5b309b6c6f95074f8d38f2a236142e2e0751ee19
```
edgarriba, line 4: @haixuanTao could be `tag` supported ? or how to set a especifc version for a crate here

PR #425, opened 2025-07-07
vlm: fix lint with print and logging

edgarriba:


Cargo.toml
```diff
@@ -50,5 +50,6 @@ num-traits = "0.2"
 rand = "0.9"
 rerun = "0.23"
 serde = { version = "1", features = ["derive"] }
+serde_json = "1"
```
AS1100K, line 53: Are we using this crate?

AS1100K (approved):

Cargo.toml
edgarriba: yes, it's needed for now to load the models

andrew-shc (approved):
Was this an issue before? Didn't see it on my local linting nor on the CI/CDs

edgarriba:
Good point! I got those errors with the ‘just lint’ command.

PR #426, opened 2025-07-08
implement smolvlm preprocessor struct

edgarriba:


crates/kornia-vlm/src/smolvlm/preprocessor.rs
```diff
@@ -1,260 +1,242 @@
...
+        )?;
+
         for y in 0..height {
             for x in 0..width {
                 for c in 0..3 {
-                    padded_img
-                        .set_pixel(x, y, c, *global_img.get_pixel(x, y, c).unwrap())
-                        .unwrap();
+                    // TODO: this method are slow, we should use a more efficient way to set the pixel maybe with iterators
+                    padded_img.set_pixel(x, y, c, *img.get_pixel(x, y, c)?)?;
```
andrew-shc: We can do something like the `image` crate https://docs.rs/image/latest/src/image/image.rs.html#1026 for kornia-image API-wise.

Performance-wise, we can do something like [copy_from_slice](https://doc.rust-lang.org/std/primitive.slice.html#method.copy_from_slice) (if the underlying CPU allocator is just `Vec`). It should be more direct on cuda

crates/kornia-vlm/src/smolvlm/preprocessor.rs
```diff
@@ -1,260 +1,242 @@
...
+        img: &Image<u8, 3, A>,
+        target_size: u32,
+    ) -> Result<Image<u8, 3, A>, SmolVlmError> {
         let (width, height) = (img.width() as u32, img.height() as u32);
         let longest_edge = width.max(height);
-        if longest_edge <= outer_patch_size {
-            img.clone()
+
+        if longest_edge <= target_size {
+            // TODO: can we avoid cloning here? and return a reference?
```
andrew-shc: conditionally, we might have to create a new image for resizing. we can have the `img` moved into here to avoid unnecessary cloning

andrew-shc (changes requested):

crates/kornia-vlm/src/smolvlm/preprocessor.rs
edgarriba: Need to work out this but maybe this function just returns the ref to the original or ref to a cached image

crates/kornia-vlm/src/smolvlm/preprocessor.rs
edgarriba: Yes and work directly with the linear indices as now we are computing several time the same index in each call. Ideally we compute once per row i * cols and we keep adding at each column position

PR #429, opened 2025-07-10
Add support for multiple AprilTag Families

AS1100K:


crates/kornia-apriltag/src/decoder.rs
```diff
@@ -236,9 +231,7 @@ impl SharpeningBuffer {
 /// edge refinement flag, and sharpening factor.
 pub struct DecodeTagsConfig<'a> {
     /// Reference to the tag family used for decoding.
-    pub tag_family: &'a TagFamily,
-    /// Precomputed quick decode table for fast tag lookup.
-    pub quick_decode: QuickDecode,
+    pub tag_family: &'a [TagFamily],
```
edgarriba: check my prev PR, no lifetime parameters are needed

edgarriba: for the config we should only pass the enum

crates/kornia-apriltag/src/family/tag36h10.rs
```diff
@@ -0,0 +1,2347 @@
+use super::*;
+
+impl TagFamily {
+    /// TODO
```
edgarriba: add some meaningful docs

AS1100K:
@edgarriba Should we store the `CODE_DATA` in a binary file and use [`include_bytes!`](https://doc.rust-lang.org/stable/std/macro.include_bytes.html) to load them as these values are constant and generated from [`AprilRobotics/apriltag-generation`](https://github.com/AprilRobotics/apriltag-generation), or we can port the generation code to build.rs and generate those values here.

As this PR has a addition of 161K lines with most of the lines pre-generated constant.

crates/kornia-apriltag/src/decoder.rs
AS1100K: What do you say about the current implementation of using `Vec<TagFamily>` instead

.gitmodules
```diff
@@ -0,0 +1,3 @@
+[submodule "tests/data/apriltag-imgs"]
+	path = tests/data/apriltag-imgs
+	url = https://github.com/AS1100K/apriltag-imgs.git
```
AS1100K, line 3: @edgarriba Should we use submodules? Also, instead of having a seperate repo should we move it's content in [data repo](https://github.com/kornia/data) or keep it in different repo and move it's ownership from my account to kornia

AS1100K:
All the Decoding tests works fine on x86_64 but takes a very long time on aarch64, should I make the tests not run on aarch64 by doing `#[cfg(not(target_arch="aarch64"))]`

crates/kornia-apriltag/src/lib.rs
```diff
@@ -27,3 +45,455 @@ pub mod quad;
 
 /// Decoding utilities for AprilTag detection.
 pub mod decoder;
+
+#[derive(Debug, Clone, PartialEq)]
+/// Configuration for decoding AprilTags.
+pub struct DecodeTagsConfig {
+    /// List of tag families to detect.
+    pub tag_families: Vec<TagFamily>,
```
edgarriba, line 53: the tag family object is too heavy to be passed in a config as it's too heavy for serialization scenarion. Pass a list of enum instead

crates/kornia-apriltag/src/lib.rs
```diff
@@ -27,3 +45,455 @@ pub mod quad;
...
+        self.tag_families.iter_mut().for_each(|family| {
+            family.sharpening_buffer.reset();
+        });
+    }
+}
+
+/// Decoder for AprilTag detection and decoding.
+pub struct AprilTagDecoder {
+    /// Configuration for decoding AprilTags.
+    pub config: DecodeTagsConfig,
```
edgarriba: keep all private and implement a pub inline getter for consistency with the rest of the struct

.gitmodules
edgarriba: yep better to have a separated repo here as seems large

crates/kornia-apriltag/src/lib.rs
```diff
@@ -27,3 +45,455 @@ pub mod quad;
...
+            reversed_border,
+            min_tag_width,
+            min_white_black_difference: 20,
+        }
+    }
+
+    /// Creates a `DecodeTagsConfig` with all supported tag families.
+    pub fn all() -> Self {
+        let tag_families = vec![
+            TagFamily::tag16_h5(),
```
edgarriba, line 105: maybe the all() method should be implemented at the TagFamily level to easily keep it updated

crates/kornia-apriltag/src/lib.rs
AS1100K: Even if we store `TagFamilyKind` enum, we would still need to store the `TagFamily` somewhere as creating it could be very time consuming (at least at this point), we could pass a enum in `new` function.

For serialization part, we should skip the values of `code_data`, `quick_decode`, and maybe `sharpening_buffer`, as `code_data`, and `quick_decode` are already known values.

crates/kornia-apriltag/src/lib.rs
AS1100K: Should we have the function something like:

```rust
impl TagFamily {
    fn all() -> Vec<TagFamily> { todo!() }
    // Or,
    fn all() -> DecodeTagsConfig { todo!() }
}
```

cjpurackal:
I'm seeing a typo in `tagcircle49_h12.rs` where the tag name is set as `tagcircle19_h12` and minor issues. It's hard to review with all the hex data. Any way to move the hex data out of the source tree?

crates/kornia-apriltag/src/lib.rs
edgarriba: the first, to avoid ciruclar deps

crates/kornia-apriltag/src/lib.rs
edgarriba: enum in the new sounds good

AS1100K:
> Any way to move the hex data out of the source tree?

We could store all these in a binary file and then use `include_bytes!`

PR #431, opened 2025-07-12
fix: v4l no longer borking macos

m0sys:
kr_v4l_no_bork


m0sys:
Tests work too now: 

kr_test_pass_macos_bork

m0sys:
@edgarriba try running CI again - locked the fast_image_resize to 5.1 and curious to see if that's gonna fix it. If so you can report a bug to them.

m0sys:
This should def fix this CI failing and #429 failing based on this: https://github.com/Cykooz/fast_image_resize/compare/v5.1.4...v5.2.0

crates/kornia-io/Cargo.toml
```diff
@@ -38,6 +34,11 @@ gstreamer-app = { version = "0.23.5", optional = true }
 # for fast jpeg encoding/decoding
 turbojpeg = { version = "1.2", optional = true }
 
+[target.'cfg(linux)'.dependencies]
```
AS1100K:
```suggestion
[target.'cfg(target_os="linux")'.dependencies]
```

examples/webcam/src/main.rs
```diff
@@ -1,135 +1,11 @@
...
-                rerun::ColorModel::RGB,
-            ),
-        )?;
-
-        if args.debug {
-            fps_counter.update();
-            println!("FPS: {:.2}", fps_counter.fps());
-        }
-    }
+    webcam::webcam_demo()?
```
AS1100K:
```suggestion
    webcam::webcam_demo()
```

examples/webcam/src/webcam.rs
```diff
@@ -0,0 +1,136 @@
+#![cfg(not(target_os = "linux"))]
```
AS1100K:
```suggestion
```

It is redundant here

AS1100K (changes requested):
LGTM, I have commented the changes so that it compiles on linux but I haven't tried it on MacOS yet but it seems it would work fine there

examples/webcam/src/webcam.rs
m0sys: Nice catch, redundant it is

m0sys:
Addressed your comments. We're good to go.

examples/webcam/src/main.rs
```diff
@@ -1,135 +1,11 @@
-use argh::FromArgs;
```
edgarriba: Maybe adding this will make more clean and keep this example code simple in a single file

'''Rust
#[cfg(not(target_os = "linux"))]
compile_error!("This crate only supports Linux.");
'''

m0sys: That would result in a compilation error for non-linux devs rendering `cargo build`, `cargo test`, etc ... unusable. I wanted to avoid the complexity of a second file - but making a dump main and everything in `webcam.rs` seems to be the most cleanest for this example.

AS1100K: It would also fail the CI, as we check if the examples can be compiled

edgarriba: Sounds good. I was planning to rename this example to v4l instead of webcam if you want here or in a separated PR

edgarriba (approved):

examples/webcam/src/main.rs
m0sys: It's done.

AS1100K: @edgarriba Also, in the webcam (now v4l) example the following text needs to be changed:

```diff
- NOTE: This example requires the gstremer backend to be enabled. To enable the gstreamer backend, use the gstreamer feature flag when building the kornia crate and its dependencies.
+ NOTE: This example only works on Linux.
```

m0sys: Readme updated too

examples/v4l/README.md
```diff
@@ -0,0 +1,21 @@
+An example showing how to use V4L to drive a webcam with the `kornia_io` module with the ability to cancel the feed after a certain amount of time. This example will display the webcam feed in a [`rerun`](https://github.com/rerun-io/rerun) window.
+
+NOTE: This example is only supported on Linux. It also requires the gstremer backend to be enabled. To enable the gstreamer backend, use the `gstreamer` feature flag when building the `kornia` crate and its dependencies.
```
AS1100K: This example doesn't require gstreamer now. Only the statement `NOTE: This example is only supported on Linux.` will be enough. Although this change was meant to be done in #359 but can be done now

m0sys: Ahh ok cool

AS1100K (approved):

examples/v4l/src/main.rs
```diff
@@ -0,0 +1,11 @@
+#[cfg(target_os = "linux")]
+mod webcam;
```
edgarriba: this should be v4l now

Issue #432, opened 2025-07-15
Gaussian Blur Auto Compute KSize


m0sys:
Hello, it seems that `imgproc::filter:gaussian_blur` does not auto compute the kernel size from the given sigmas if ksize is 0. This is the expected behavior when using opencv::gaussian_blur as per the [documentation](https://docs.opencv.org/4.x/d4/d86/group__imgproc__filter.html#gae8bdcd9154ed5ca3cbc1766d960f45c1). 

I would like to extend the implementation to support this use case.

edgarriba:
feel free to implement it !

PR #434, opened 2025-07-16
Blur ksize auto

m0sys:
PR for #432


m0sys:
Few comments:
- not sure if you want to avoid any asserts outside of tests
- I wrote a python script to get values for your previous test using cv2 to see if that's how you got the values for 3x3 @0.5 gaussian blur but they didn't match up. So I wrote regression tests i.e. ran the tests and asserted against what was outputted rather than what's expected
- Also added sigma auto-compute based on this [opencv::getGaussianKernel](https://docs.opencv.org/4.12.0/d4/d86/group__imgproc__filter.html#gac05a120c1ae92a6060dd0db190a61afa)
- Used [opencv::createGaussianKernels](https://github.com/opencv/opencv/blob/4813d1cd3245013df505af2e036c075562187bc2/modules/imgproc/src/smooth.dispatch.cpp#L280) as ref

crates/kornia-imgproc/src/filter/ops.rs
```diff
@@ -33,17 +33,53 @@ pub fn box_blur<const C: usize, A1: ImageAllocator, A2: ImageAllocator>(
...
+    //       bit is always ON.
+    if kernel_size.0 == 0 && sigma.0 > 0.0 {
+        kernel_size.0 =
+            (sigma.0 * (if C == 1 { 3.0 } else { 4.0 }) * 2.0 + 1.0).round() as usize | 1;
+    }
+    if kernel_size.1 == 0 && sigma.1 > 0.0 {
+        kernel_size.1 =
+            (sigma.1 * (if C == 1 { 3.0 } else { 4.0 }) * 2.0 + 1.0).round() as usize | 1;
+    }
+    assert!(
```
AS1100K: Don't use assert outside of tests, as it will panic at runtime instead use `Result` and handle errors nicely

crates/kornia-imgproc/src/filter/ops.rs
```diff
@@ -33,17 +33,53 @@ pub fn box_blur<const C: usize, A1: ImageAllocator, A2: ImageAllocator>(
...
-/// * `kernel_size` - The size of the kernel (kernel_x, kernel_y).
-/// * `sigma` - The sigma of the gaussian kernel.
+/// * `kernel_size` - The size of the kernel (kernel_x, kernel_y). Or, they can
+///                   be zero and they will be computed from sigma values.
+/// * `sigma` - The sigma of the gaussian kernel. sigma.1 can be zero and it
+///             will take on the same value as sigma.0. Or, they can both be zero
+///             and they will be computed based on:
+///             [sigma = 0.3 * ((ksize - 1) * 0.5 - 1) + 0.8](https://docs.opencv.org/4.12.0/d4/d86/group__imgproc__filter.html#gac05a120c1ae92a6060dd0db190a61afa).
 ///
 /// PRECONDITION: `src` and `dst` must have the same shape.
```
AS1100K, line 45: Add a check for their size, and return `Err` if it doesn't matches

AS1100K (changes requested):

crates/kornia-imgproc/src/filter/ops.rs
m0sys: Technically if both sigmas are 0, the auto-compute-ksize will result in kernel size = 1. This will then feed into the auto-compute-sigma and result in both sigmas being 0.5. So we don't need to handle this case

crates/kornia-imgproc/src/filter/ops.rs
m0sys: Also you think I should return a ImageError::InvalidKernelLength or create a new type of error ImageError:InvalidSigmaValue?

crates/kornia-imgproc/src/filter/ops.rs
```diff
@@ -33,17 +33,55 @@ pub fn box_blur<const C: usize, A1: ImageAllocator, A2: ImageAllocator>(
...
 /// NOTE: This function uses a constant border type.
 pub fn gaussian_blur<const C: usize, A1: ImageAllocator, A2: ImageAllocator>(
     src: &Image<f32, C, A1>,
     dst: &mut Image<f32, C, A2>,
-    kernel_size: (usize, usize),
-    sigma: (f32, f32),
+    mut kernel_size: (usize, usize),
```
edgarriba: not mut here, create a tmp variable inside

crates/kornia-imgproc/src/filter/ops.rs
```diff
@@ -407,6 +445,68 @@ mod tests {
...
     }
 
+    #[test]
+    fn test_gaussian_blur_autocompute_ksize() -> Result<(), ImageError> {
+        let size = ImageSize {
+            width: 5,
+            height: 5,
+        };
+
+        #[rustfmt::skip]
```
edgarriba: nitpick: cargo fmt should take care of this. We use skip only in tests to have more human readable formatted matrices as you did well a bit below

edgarriba (changes requested):
@ducha-aiki can you double check that the math is correct here ?

crates/kornia-imgproc/src/filter/ops.rs
```diff
@@ -33,17 +33,55 @@ pub fn box_blur<const C: usize, A1: ImageAllocator, A2: ImageAllocator>(
...
 /// * `dst` - The destination image with shape (H, W, C).
-/// * `kernel_size` - The size of the kernel (kernel_x, kernel_y).
-/// * `sigma` - The sigma of the gaussian kernel.
+/// * `kernel_size` - The size of the kernel (kernel_x, kernel_y). They can differ,
+///                   but they both have to be positive and odd. Or, they can be zero
+///                   and they will be computed from sigma values.
+/// * `sigma` - The sigma of the gaussian kernel (sigma_x, sigma_y). sigma_y can
+///             be zero and it will take on the same value as sigma_x. Or, they
+///             can both be zero and they will be computed based on:
+///             [sigma = 0.3 * ((ksize - 1) * 0.5 - 1) + 0.8](https://docs.opencv.org/4.12.0/d4/d86/group__imgproc__filter.html#gac05a120c1ae92a6060dd0db190a61afa).
```
edgarriba: don't add this link here. If you want add in as comment near the function signature

m0sys: resolved them

crates/kornia-imgproc/src/filter/ops.rs
```diff
@@ -44,8 +49,44 @@ pub fn gaussian_blur<const C: usize, A1: ImageAllocator, A2: ImageAllocator>(
...
+    if sigma_cpy.1 <= 0.0 {
+        sigma_cpy.1 = sigma_cpy.0;
+    }
+
+    // Auto-compute the kernel sizes based on sigma if 0 or negative.
+    // NOTE: the `| 1` is to ensure that the number is always odd i.e. the 2^0
+    //       bit is always ON.
+    if ksize_cpy.0 == 0 && sigma_cpy.0 > 0.0 {
+        ksize_cpy.0 =
+            (sigma_cpy.0 * (if C == 1 { 3.0 } else { 4.0 }) * 2.0 + 1.0).round() as usize | 1;
```
ducha-aiki: Why do we have different kernel size depending on the number of channels?
The correct formula is with 4 = as in scipy:
```
`2*radius + 1``
```
https://github.com/scipy/scipy/blob/v1.16.0/scipy/ndimage/_filters.py#L706

crates/kornia-imgproc/src/filter/ops.rs
```diff
@@ -44,8 +49,44 @@ pub fn gaussian_blur<const C: usize, A1: ImageAllocator, A2: ImageAllocator>(
...
+    }
+    if ksize_cpy.1 == 0 && sigma_cpy.1 > 0.0 {
+        ksize_cpy.1 =
+            (sigma_cpy.1 * (if C == 1 { 3.0 } else { 4.0 }) * 2.0 + 1.0).round() as usize | 1;
```
ducha-aiki: same here

crates/kornia-imgproc/src/filter/ops.rs
```diff
@@ -44,8 +49,44 @@ pub fn gaussian_blur<const C: usize, A1: ImageAllocator, A2: ImageAllocator>(
...
+    }
+
+    // Sigma should be always positive.
+    sigma_cpy.0 = sigma_cpy.0.max(0.0);
+    sigma_cpy.1 = sigma_cpy.1.max(0.0);
+
+    // Auto-compute the sigma values based on 0.3*((ksize-1)*0.5 - 1) + 0.8 formula.
+    // See https://stackoverflow.com/a/14060998 for more details.
+    if sigma_cpy.0 == 0.0 {
+        sigma_cpy.0 = 0.3 * (((ksize_cpy.0 as f32) - 1.0) * 0.5 - 1.0) + 0.8;
```
ducha-aiki: I do understand, that this matches the OpenCV convention, but interestingly enough, this is not compatible with the formula for ks:

get_ks(3) -> 25
get_s(25) -> 4.1

crates/kornia-imgproc/src/filter/ops.rs
m0sys: So I'm using https://github.com/opencv/opencv/blob/4813d1cd3245013df505af2e036c075562187bc2/modules/imgproc/src/smooth.dispatch.cpp#L280 as my reference. 

Now that you put it that way I do I agree that it doesn't make sense for it to depend on channels. In my mind I quickly replaced depth for channel - but it's actually `data type of elements in matrix`. It seems like they use $ksize = 6\sigma + 1$ when type is `u8` and $ksize = 8\sigma + 1$ when it's anything else. Given $radius = 3\sigma$ the `u8` agrees with yours.

crates/kornia-imgproc/src/filter/ops.rs
m0sys: I asks Grok4 and o3 and here are some nice tables: 

| Method | Formula | Use Case | Precision | Performance |
| :-- | :-- | :-- | :-- | :-- |
| 3-Sigma Rule | kernel_size = 6σ + 1 | General purpose | High | Good |
| OpenCV Formula | σ = 0.3(n/2-1) + 0.8 | Small kernels | Medium | Excellent |
| Practical Trade-off | σ = (size-1)/4 | Balanced approach | Medium | Good |
| 4-Sigma Truncation | kernel_size = 8σ + 1 | High precision | Very High | Lower |

| Convention (who uses it) | From σ → kernel size *k* | From *k* → σ | Rationale |
| :-- | :-- | :-- | :-- |
| 6 σ rule (MATLAB, many texts) | `k = 2*ceil(3*σ) + 1` | `σ = (k − 1) / 6` | 6 σ spans ≈ 99.7% of Gaussian energy, so tails are negligible |
| SciPy `gaussian_filter` | `k = 2*round(t*σ) + 1` (default `t = 4`) | `σ = (k − 1) / (2*t)` | Keeps all samples farther than *t σ* below a chosen cut-off |
| OpenCV `getGaussianKernel` / `GaussianBlur` | `k = 2*round(((σ − 0.8)/0.3 + 1)) + 1` | `σ = 0.3*((k − 1)/2 − 1) + 0.8` | Tuned so the edge weight falls below 1/256 for 8-bit images |
| Radius-based (DSP texts) | Choose radius *r* (e.g., *r = 3 σ*), then `k = 2*r + 1` | `σ = r / 3` | Makes the kernel as wide as three standard deviations in every direction |

crates/kornia-imgproc/src/filter/ops.rs
m0sys: So in our case I guess since we always have an f32 image I just use $8\sigma + 1$? That would align with scipy since default `t=4` as per the below tables.

m0sys:
@edgarriba I fixed the math by using scipy convention while making them inverses of each other which should align with @ducha-aiki comments now

crates/kornia-imgproc/src/filter/ops.rs
```diff
@@ -44,8 +50,41 @@ pub fn gaussian_blur<const C: usize, A1: ImageAllocator, A2: ImageAllocator>(
...
+
+    // Satisfy setting sigma_y = sigma_x if sigma_y is zero.
+    if sigma_cpy.1 <= 0.0 {
+        sigma_cpy.1 = sigma_cpy.0;
+    }
+
+    // Auto-compute the kernel sizes based on sigma if 0 or negative using SciPy convention.
+    // NOTE: the `| 1` is to ensure that the number is always odd i.e. the 2^0
+    //       bit is always ON.
+    if ksize_cpy.0 == 0 && sigma_cpy.0 > 0.0 {
```
edgarriba, line 64:
```suggestion
    if kernel_x == 0 && sigma_x > 0.0 {
```
would be more readible with something like

edgarriba: you could do in the first place

```rust
let (mut kernel_x, mut kernel_y) = kernel_size;
```

edgarriba (changes requested):
i suggest also to have also a `create_gaussian_kernels` function which has a unit test to verify the formulation

PR #444, opened 2025-08-02
Add image downscaling support in AprilTag Detection

AS1100K:
## Benchmarks

1. NVIDIA Orin
    ```txt
    kornia-apriltag         time:   [11.525 ms 11.526 ms 11.528 ms]
                            change: [−68.450% −68.443% −68.436%] (p = 0.00 < 0.05)
    ```

3. AMD Ryzen 7 5800X
    ```txt
    kornia-apriltag         time:   [5.4888 ms 5.4963 ms 5.5049 ms]
                            change: [−68.551% −68.469% −68.392%] (p = 0.00 < 0.05)
    ```

### DIFF
--- a/crates/kornia-apriltag/src/lib.rs
+++ b/crates/kornia-apriltag/src/lib.rs
@@ -7,7 +7,7 @@ use kornia_image::{
     allocator::{CpuAllocator, ImageAllocator},
     Image, ImageSize,
 };
-use kornia_imgproc::resize::resize_fast;
+use kornia_imgproc::resize::resize_fast_gray;
 
 use crate::{
     decoder::{decode_tags, Detection, GrayModelPair},
@@ -219,7 +219,7 @@ impl AprilTagDecoder {
             )?;
         } else {
             let downscale_img = unsafe { self.downscale_img.as_mut().unwrap_unchecked() };
-            resize_fast(
+            resize_fast_gray(
                 &src,
                 downscale_img,
                 kornia_imgproc::interpolation::InterpolationMode::Nearest,
--- a/crates/kornia-apriltag/src/lib.rs
+++ b/crates/kornia-apriltag/src/lib.rs
@@ -7,7 +7,7 @@ use kornia_image::{
     allocator::{CpuAllocator, ImageAllocator},
     Image, ImageSize,
 };
-use kornia_imgproc::resize::resize_fast_gray;
+use kornia_imgproc::resize::resize_fast_mono;
 
 use crate::{
     decoder::{decode_tags, Detection, GrayModelPair},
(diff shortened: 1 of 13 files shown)


crates/kornia-apriltag/src/lib.rs
```diff
@@ -185,15 +209,30 @@ impl AprilTagDecoder {
...
-        // TODO: Add support for downscaling image
-
-        // Step 1: Adaptive Threshold
-        adaptive_threshold(
-            src,
-            &mut self.bin_img,
-            &mut self.tile_min_max,
-            self.config.min_white_black_difference,
-        )?;
+        if self.config.downscale_factor <= 1 {
```
edgarriba: maybe structure based on the existence of the optional
`if let Some(mut &downscale_img) = ...`

crates/kornia-apriltag/src/lib.rs
```diff
@@ -275,14 +314,14 @@ mod tests {
 
                 for (point, expected) in detection.quad.corners.iter().zip(expected_quads.iter()) {
                     assert!(
-                        (point.y - expected.y).abs() <= 0.001,
+                        (point.y - expected.y).abs() <= 0.1,
```
edgarriba, line 316: quite heavy change. What did we change here ?

crates/kornia-imgproc/src/resize.rs
```diff
@@ -150,26 +150,55 @@ pub fn resize_fast<A1: ImageAllocator, A2: ImageAllocator>(
...
+/// * `interpolation` - The interpolation mode to use.
+///
+/// # Returns
+///
+/// The resized image with the new size.
+///
+/// # Errors
+///
+/// The function returns an error if the image cannot be resized.
+pub fn resize_fast_gray<A1: ImageAllocator, A2: ImageAllocator>(
```
edgarriba:
```suggestion
pub fn resize_fast_mono<A1: ImageAllocator, A2: ImageAllocator>(
```
and the other rename to `color` ?

crates/kornia-imgproc/src/resize.rs
```diff
@@ -150,26 +150,64 @@ pub fn resize_fast<A1: ImageAllocator, A2: ImageAllocator>(
...
-        src.as_slice(),
-        fr::PixelType::U8x3,
-    )
-    .map_err(|_| ImageError::InvalidChannelShape(src_data_len, src_cols * src_rows * 3))?;
+    let pixel_type = match C {
+        4 => fr::PixelType::U8x4,
+        3 => fr::PixelType::U8x3,
+        1 => fr::PixelType::U8,
+        // TODO: Find a way to generalise it further by supporting multiple types other than u8
+        _ => panic!("Unsupported number of channels: {C}. Only 1, 3, or 4 channels are supported for resize_fast."),
```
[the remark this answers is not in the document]
edgarriba: +1 to return proper error

edgarriba:
@strasdat what's your bandwidth to setup a small example using sophus-rs optim for calibration ?

crates/kornia-apriltag/src/lib.rs
AS1100K: The quad detection is done on downscale image, and I believe that's the reason it's happens as the quad position is later multiplied by the `downscale factor` which increase the error slightly.

Also, what do you say if we return the position after calling the round function

crates/kornia-imgproc/src/resize.rs
AS1100K: Maybe use `resize_fast_rgb` for the color one as we are use `_rgb` everywhere

PR #447, opened 2025-08-03
pref(histogram.rs): parallelized histogram generation

dhruvkjain:
#### Implemented parallel generation of histogram using `rayon` crate optimizing performance as measured below:

1. for kornia-rs/tests/data/apriltags_tag36h11.jpg    
    - without parallelized (RGB)
       - elapsed: 6.328555ms
       - elapsed: 5.98668ms
       - elapsed: 5.968598ms
    - with parallelization (RGB)
       - elapsed: 2.482211ms
       - elapsed: 1.67441ms
       - elapsed: 1.264231ms

2. for kornia-rs/tests/data/dogs.jpeg
    - without parallelized (RGB)
       - elapsed: 749.052µs
       - elapsed: 719.1µs
       - elapsed: 712.953µs
    - with parallelization (RGB)
       - elapsed: 1.045121ms
       - elapsed: 409.218µs
       - elapsed: 413.271µs

#### I measured the elapsed time using:
```rust
let now = std::time::Instant::now();
// logic
let elapsed = now.elapsed();
println!("elapsed: {:?}", elapsed);
```

### DIFF
--- a/crates/kornia-imgproc/src/histogram.rs
+++ b/crates/kornia-imgproc/src/histogram.rs
@@ -1,4 +1,5 @@
 use kornia_image::{allocator::ImageAllocator, Image, ImageError};
+use rayon::prelude::*;
 
 /// Compute the pixel intensity histogram of an image.
 ///
@@ -55,12 +56,29 @@ pub fn compute_histogram<A: ImageAllocator>(
     // we assume 8-bit images for now and range [0, 255]
     let scale = 256.0 / num_bins as f32;
 
-    // TODO: check if this can be done in parallel
-    src.as_slice().iter().fold(hist, |histogram, &pixel| {
-        let bin_pos = (pixel as f32 / scale).floor();
-        histogram[bin_pos as usize] += 1;
-        histogram
-    });
+    // detect number of threads hardware can allocate to rayon
+    let num_threads = rayon::current_num_threads();
+
+    // parallaized computation of histogram on local threads 
+    let src_slice = src.as_slice();
+    let partials: Vec<Vec<usize>> = src_slice
+        .par_chunks(src_slice.len() / num_threads + 1)
+        .map(|chunk| {
+            let mut local_hist = vec![0_usize; num_bins];
+            for &pixel in chunk {
+                let bin = (pixel as f32 / scale).floor() as usize;
+                local_hist[bin] += 1;
+            }
+            local_hist
+        })
+        .collect();
+
+    // merge all local histograms to one
+    for partial in partials {
+        for (i, count) in partial.into_iter().enumerate() {
+            hist[i] += count;
+        }
+    }
 
     Ok(())
 }
@@ -69,7 +87,7 @@ pub fn compute_histogram<A: ImageAllocator>(
 mod tests {
     use kornia_image::{Image, ImageError, ImageSize};
     use kornia_tensor::CpuAllocator;
-
+    
     #[test]
     fn test_compute_histogram() -> Result<(), ImageError> {
         let image = Image::new(


crates/kornia-imgproc/src/histogram.rs
```diff
@@ -55,12 +56,29 @@ pub fn compute_histogram<A: ImageAllocator>(
...
-        histogram[bin_pos as usize] += 1;
-        histogram
-    });
+    // detect number of threads hardware can allocate to rayon
+    let num_threads = rayon::current_num_threads();
+
+    // parallaized computation of histogram on local threads 
+    let src_slice = src.as_slice();
+    let partials: Vec<Vec<usize>> = src_slice
+        .par_chunks(src_slice.len() / num_threads + 1)
```
edgarriba: why not to parallelize over the image rows instead ?

dhruvkjain: I tried row-wise approach using the code below:

```rust
let width = src.width();
let height = src.height();
let src_slice = src.as_slice();

let partials: Vec<Vec<usize>> = (0..height)
    .into_par_iter()
    .map(|row| {
        let mut local_hist = vec![0_usize; num_bins];
        let start = row * width;
        let end = start + width;

        for &pixel in &src_slice[start..end] {
            let bin = (pixel as f32 / scale).floor() as usize;
            local_hist[bin] += 1;
        }

        local_hist
    })
    .collect();

for partial in partials {
    for (i, count) in partial.into_iter().enumerate() {
        hist[i] += count;
     }
}
```


and the results clearly shows that it's even slower than non-parallelized version
1. for kornia-rs/tests/data/apriltags_tag36h11.jpg
    elapsed: 9.955734ms
    elapsed: 8.736156ms
    elapsed: 9.15354ms
2. for kornia-rs/tests/data/dogs.jpeg
    elapsed: 3.092654ms
    elapsed: 2.111322ms
    elapsed: 2.049358ms

This is because in row-wise we are not utilizing the available threads efficiently. Rather we are increasing the overhead of spawning and synchronizing parallel tasks of more threads.

edgarriba: i think to do properly by rows you should iterate using the chunk iterator directly on the slice. Check how is done in other functions in the library

would like to see what's the performance with something like
```rust
let hist = src_slice
    .par_chunks_exact(width)
    .map(|row| {
        let mut local_hist = vec![0_usize; num_bins];
        for &pixel in row {
            let bin = (pixel as f32 / scale).floor() as usize;
            local_hist[bin] += 1;
        }
        local_hist
    })
    .reduce(
        || vec![0; num_bins],
        |mut a, b| {
            for (i, val) in b.into_iter().enumerate() {
                a[i] += val;
            }
            a
        },
    );
```

edgarriba: besides, in the rest of the library we have taken the approach to parallelize by rows as a tradeoff to not satured all the available threads as usually when you use libraries like kornia, especially with constrained devices you want to keep space for other stuff happening in your stack. That's a common approach followed by vision libraries

dhruvkjain: Yeah you are right, my bad for not realizing it, this approach make much more sense.
This code make chunks of image width and performed better  
 
```rust
let partial_hist = src_slice
    .par_chunks_exact(width)
    .map(|row| {
        let mut local_hist = vec![0_usize; num_bins];
        for &pixel in row {
            let bin = (pixel as f32 / scale).floor() as usize;
            local_hist[bin] += 1;
        }
        local_hist
    })
    .reduce(
        || vec![0; num_bins],
        |mut a, b| {
            for (i, val) in b.into_iter().enumerate() {
                a[i] += val;
            }
            a
        },
    );

for (i, val) in partial_hist.into_iter().enumerate() {
    hist[i] += val;
}
```

1. for kornia-rs/tests/data/apriltags_tag36h11.jpg
elapsed: 2.093162ms
elapsed: 2.310861ms
elapsed: 1.946358ms
2. for kornia-rs/tests/data/dogs.jpeg
elapsed: 654.845µs
elapsed: 613.694µs
elapsed: 570.008µs

Issue #394, opened 2025-06-14
se2 exp_log_roundtrip fails CI aarch64


edgarriba:
https://github.com/kornia/kornia-rs/actions/runs/15536812342/job/43737743324?pr=389

```bash
---- se2::tests::test_exp_log_roundtrip stdout ----

thread 'se2::tests::test_exp_log_roundtrip' panicked at crates/kornia-lie/src/se2.rs:376:13:
assert_relative_eq!(log_upsilon.x, upsilon.x, epsilon = EPSILON)

    left  = 0.7224934
    right = 0.7228841


note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace


failures:
    se2::tests::test_exp_log_roundtrip

test result: FAILED. 80 passed; 1 failed; 0 ignored; 0 measured; 0 filtered out; finished in 0.16s
```

edgarriba:
#396 partially fix the CI but not the real issue. Might be something related to glam with aarch64 ? Needs to be investigated

Issue #393, opened 2025-06-14
SE3  adjoint properties test random fail in CI for aarch64


edgarriba:
https://github.com/kornia/kornia-rs/actions/runs/15637922163/job/44057827653?pr=391

```bash
---- se3::tests::test_adjoint_properties stdout ----

thread 'se3::tests::test_adjoint_properties' panicked at crates/kornia-lie/src/se3.rs:507:17:
assert_relative_eq!(adj_composed[i][j], adj_product[i][j], epsilon = EPSILON)

    left  = 0.34697977
    right = 0.34698108


note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace


failures:
    se3::tests::test_adjoint_properties

test result: FAILED. 80 passed; 1 failed; 0 ignored; 0 measured; 0 filtered out; finished in 0.16s
```

edgarriba:
#396 partially fix the CI but not the real issue. Might be something related to glam with aarch64 ? Needs to be investigated

AS1100K:
Why don't make the `EPSILON` larger, currently it is `1e-6`, why not make it `1e-4`

edgarriba:
we should find the real root case, relaxing the epsilon would be just a hack for the ci

PR #448, opened 2025-08-06
feat(image): added is_contiguous and to_contiguous for Image solving kornia-py/image.rs todo

dhruvkjain:
Added is_contiguous and to_contiguous for Image that solve [kornia-py/image.rs line 32, 43, 54 todos](https://github.com/kornia/kornia-rs/blob/0995adf0b516dfe61a2e98c331af56fd9fc4cf15/kornia-py/src/image.rs#L32)

`is_contiguous` function checks if the actual stride matches what a tightly packed layout would require
`to_contiguous` function if not already contiguous converts the data from Image to contiguous and returns a new Image

also added test cases and docs for is_contiguous and to_contiguous

### DIFF
--- a/crates/kornia-image/src/image.rs
+++ b/crates/kornia-image/src/image.rs
@@ -370,6 +370,28 @@ impl<T, const C: usize, A: ImageAllocator> Image<T, C, A> {
     }
 
     /// Checks if Image data is stored in contiguous memory
+    ///
+    /// # Returns
+    ///
+    /// boolean, true if contiguous and false if not
+    ///
+    /// # Examples
+    ///
+    /// ```
+    /// use kornia_image::{Image, ImageSize};
+    /// use kornia_image::allocator::CpuAllocator;
+    ///
+    /// let image = Image::<f32, 2, _>::from_size_val(
+    ///   ImageSize {
+    ///    width: 10,
+    ///   height: 20,
+    /// },
+    /// 0.0f32,
+    /// CpuAllocator).unwrap();
+    ///
+    /// let channels = image.split_channels().unwrap();
+    /// assert_eq!(channels.len(), 2);
+    /// ```
     pub fn is_contiguous(&self) -> bool {
         let mut expected_stride = 1;
         for i in (0..C).rev() {
@@ -938,6 +960,30 @@ mod tests {
         image_u8.strides[0] = 10;
 
         assert!(!image_u8.is_contiguous());
+        Ok(())
+    }
+
+    #[test]
+    fn test_image_to_contiguous() -> Result<(), ImageError> {
+        let mut image = Image::<u8, 3, CpuAllocator>::new(
+            ImageSize {
+                width: 10,
+                height: 20,
+            },
+            vec![0u8; 10 * 20 * 3],
+            CpuAllocator,
+        )?;
+
+        // Manually break the stride to simulate non-contiguous layout
+        image.strides[0] = 10;
+
+        assert!(!image.is_contiguous());
+
+        let contiguous = image.to_contiguous(CpuAllocator);
+
+        assert!(contiguous.is_contiguous());
+        assert_eq!(contiguous.as_slice(), &vec![0u8; 10 * 20 * 3]);
+
         Ok(())
     }
 }
--- a/crates/kornia-image/src/image.rs
+++ b/crates/kornia-image/src/image.rs
@@ -382,7 +382,7 @@ impl<T, const C: usize, A: ImageAllocator> Image<T, C, A> {
     }
 
     /// Copy Image data into contiguous memory if not already
-    pub fn to_contiguous(&self) -> Vec<T> 
+    pub fn to_contiguous(&self, alloc: A) -> Self 
     where
         T: Clone,
     {
@@ -390,7 +390,11 @@ impl<T, const C: usize, A: ImageAllocator> Image<T, C, A> {
         let mut flat = Vec::with_capacity(self.shape.iter().product());
         let data = self.as_slice();
         if self.is_contiguous() {
-            return data[..total_eles].to_vec();
+            return Self(Tensor3::from_shape_vec(
+                self.shape,
+                data[..total_eles].to_vec(),
+                alloc,
+            ).unwrap());
         }
 
         let mut idx = [0usize; C];
@@ -414,7 +418,12 @@ impl<T, const C: usize, A: ImageAllocator> Image<T, C, A> {
             }
         }
 
-        flat
+        
+        Self(Tensor3::from_shape_vec(
+                self.shape,
+                flat,
+                alloc,
+            ).unwrap())
     }
 
     /// Cast the pixel data to a different type and scale it.
--- a/crates/kornia-image/src/image.rs
+++ b/crates/kornia-image/src/image.rs
@@ -369,6 +369,56 @@ impl<T, const C: usize, A: ImageAllocator> Image<T, C, A> {
         C
     }
 
+    /// Checks if Image data is stored in contiguous memory
+    pub fn is_contiguous(&self) -> bool {
+        let mut expected_stride = 1;
+        for i in (0..C).rev() {
+            if self.strides[i] != expected_stride {
+                return false;
+            }
+            expected_stride *= self.shape[i];
+        }
+        true
+    }
+
+    /// Copy Image data into contiguous memory if not already
+    pub fn to_contiguous(&self) -> Vec<T> 
+    where
+        T: Clone,
+    {
+        let total_eles = self.shape.iter().product();
+        let mut flat = Vec::with_capacity(self.shape.iter().product());
+        let data = self.as_slice();
+        if self.is_contiguous() {
+            println!("here");
+            return data[..total_eles].to_vec();
+        }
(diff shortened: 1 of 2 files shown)


crates/kornia-image/src/image.rs
```diff
@@ -369,6 +369,109 @@ impl<T, const C: usize, A: ImageAllocator> Image<T, C, A> {
         C
     }
 
+    /// Checks if Image data is stored in contiguous memory
```
edgarriba: this comment it's not accurate as the memory is assumed to be contiguous all the time as tensors are mainly constructed by Vec which already fulfill that condition. What this function is checking is another thing

edgarriba: why don't we name this different, something like `is_standard_layout()` ?

crates/kornia-image/src/image.rs
```diff
@@ -369,6 +369,109 @@ impl<T, const C: usize, A: ImageAllocator> Image<T, C, A> {
...
+    ///         width: 1,
+    ///     },
+    ///     vec![0, 128],
+    ///     CpuAllocator,
+    /// ).unwrap();
+    /// assert!(image_u8.is_contiguous());
+    /// ```
+    pub fn is_contiguous(&self) -> bool {
+        let mut expected_stride = 1;
+        for i in (0..C).rev() {
```
edgarriba: this is wrong i think, it should iterate over the strides len. Besides, I would go further an implement this functio at the Tensor level and we just expose to the image api

crates/kornia-image/src/image.rs
```diff
@@ -369,6 +369,109 @@ impl<T, const C: usize, A: ImageAllocator> Image<T, C, A> {
...
+    ///     CpuAllocator,
+    /// ).unwrap();
+    /// // Manually break the stride to simulate non-contiguous layout
+    /// image.strides[0] = 10;
+    /// assert!(!image.is_contiguous());
+    /// let contiguous = image.to_contiguous(CpuAllocator);
+    /// assert!(contiguous.is_contiguous());
+    /// assert_eq!(contiguous.as_slice(), &vec![0u8; 10 * 20 * 3]);
+    /// ```
+    pub fn to_contiguous(&self, alloc: A) -> Self
```
edgarriba:
```suggestion
    pub fn to_standard_layout(&self, alloc: A) -> Self 
```

crates/kornia-image/src/image.rs
```diff
@@ -850,4 +953,61 @@ mod tests {
...
+            ImageSize {
+                height: 2,
+                width: 1,
+            },
+            vec![0, 128],
+            CpuAllocator,
+        )?;
+
+        // Manually break the stride to simulate non-contiguous layout
+        image_u8.strides[0] = 10;
```
edgarriba: seeing probably we should consider not allowing the user modify like that but instead provide an ImageView e.g to create ROIs

edgarriba (commented):
@AS1100K can you check this one ?

crates/kornia-image/src/image.rs
dhruvkjain: Yes, as this function is checking for reordering or padding, `is_standard_layout` makes sense

maybe this comment would clarify it's functionality instead of the previous one ?
```
Checks if the image has a standard contiguous layout according to its `shape` and `strides`.
```

crates/kornia-image/src/image.rs
dhruvkjain: as both shape and strides are declared of `C` length ([declaration here](https://github.com/kornia/kornia-rs/blob/95e616cfda9b13705601a791781f99e300ac5f5d/crates/kornia-tensor/src/tensor.rs#L82)) the code will loop correctly but,
for better understanding maybe something like this for Tensor level implementation ?

```rust
let mut expected_stride: usize = 1;
for (&dim, &stride) in self.shape.iter().rev().zip(self.strides.iter().rev()) {
   if stride != expected_stride {
         return false;
   }
   expected_stride *= dim;
}
```

or instead add `expected_stride = expected_stride.saturating_mul(dim);` ,  clamping the value at `usize::MAX`

AS1100K (commented):
I think we should only convert the data from non-contiguous to contiguous data in the python side. As the rust side doesn't support non-contiguous data _(technically it can be created but will cause a lot of issue as some functions like as_slice and other won't work)_.

If we want to support non-contiguous data on the rust side, the Tensor API needs a change

crates/kornia-image/src/image.rs
edgarriba: sure

crates/kornia-image/src/image.rs
edgarriba: yep, makes sense

dhruvkjain:
Hey @edgarriba ,
whenever you are free can you please look at the test which is failing, 
on my local setup all test in `kornia-lie` crate are passing.

AS1100K:
> whenever you are free can you please look at the test which is failing,
> on my local setup all test in `kornia-lie` crate are passing.

This is a known issue (#393 and #394). The CI randomly fails on aarch64 due to floating point precision error on the CI. I will rerun the failed CI jobs for now

crates/kornia-tensor/src/tensor.rs
```diff
@@ -618,6 +618,84 @@ impl<T, const N: usize, A: TensorAllocator> Tensor<T, N, A> {
...
+                if idx[dim] < self.shape[dim] {
+                    break;
+                } else {
+                    idx[dim] = 0;
+                }
+            }
+        }
+
+        Tensor::from_shape_vec(self.shape, flat, alloc)
+            .expect("to_standard_layout(tensor.rs): construction failed")
```
edgarriba: handle properly errors

crates/kornia-tensor/src/tensor.rs
```diff
@@ -618,6 +618,84 @@ impl<T, const N: usize, A: TensorAllocator> Tensor<T, N, A> {
...
+    /// // altering strides
+    /// t.strides = [1, 6, 2];
+    /// assert!(!t.is_standard_layout());
+    /// let t2 = t.to_standard_layout(CpuAllocator);
+    /// assert!(t2.is_standard_layout());
+    /// ```
+    pub fn to_standard_layout(&self, alloc: A) -> Self
+    where
+        T: Clone,
+    {
```
edgarriba, line 676: should we check wether it's altready in standard layout ? or at least especify in the docs

edgarriba (changes requested):

edgarriba:
i guess you need to rebase from main

dhruvkjain:
In `kornia-py/image.rs` the function to convert to python image has function definition as `fn to_pyimage(self) -> PyImage`

There can be an error while converting data to standard layout but current function definition doesn't have a return type of ImageError like other functions (`fn from_pyimage(image: PyImage) -> Result<Image<u8, C, CpuAllocator>, ImageError>`), so to handle error I have wrote a match and simply logged the error there and return array 

But I suggest it should be `fn to_pyimage(self) -> Result<PyImage, ImageError>`

Would it be okay to update the function definition and all it's references to return Result instead?

edgarriba:
> In `kornia-py/image.rs` the function to convert to python image has function definition as `fn to_pyimage(self) -> PyImage`
> 
> There can be an error while converting data to standard layout but current function definition doesn't have a return type of ImageError like other functions (`fn from_pyimage(image: PyImage) -> Result<Image<u8, C, CpuAllocator>, ImageError>`), so to handle error I have wrote a match and simply logged the error there and return array
> 
> But I suggest it should be `fn to_pyimage(self) -> Result<PyImage, ImageError>`
> 
> Would it be okay to update the function definition and all it's references to return Result instead?

yes, i think makes sense to update that function signature to return an error

kornia-py/src/io/jpeg.rs
```diff
@@ -81,20 +93,33 @@ pub fn decode_image_jpeg(src: &[u8]) -> PyResult<PyImage> {
...
+            })?;
+            output_pyimage
         }
         ch => {
-            return Err(PyErr::new::<pyo3::exceptions::PyValueError, _>(
-                String::from(format!("Unsupported number of channels: {}", ch)),
-            ))
+            return Err(PyErr::new::<pyo3::exceptions::PyValueError, _>(format!(
+                "Unsupported number of channels: {}",
+                ch
```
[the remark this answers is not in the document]
dhruvkjain: When I ran `clippy` I got a warning as follows:

```rust
warning: useless conversion to the same type: `std::string::String`
   --> src/io/jpeg.rs:120:17
    |
120 |                 String::from(format!("Unsupported number of channels: {}", ch)),
    |                 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ help: consider removing `String::from()`: `format!("Unsupported number of channels: {}", ch)`
    |
    = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#useless_conversion
    = note: `#[warn(clippy::useless_conversion)]` on by default
```

as `format` already convert to `String`

edgarriba (approved):

PR #456, opened 2025-08-22
camera distortion parameters for  kornia-pnp crate

### LINKED ISSUE #453: Add camera distortion parameters to `kornia-pnp` crate.

jeffin07:
Added camera distortion parameters for kornia-pnp create 
Fixes #453

### DIFF
--- a/crates/kornia-pnp/src/camera.rs
+++ b/crates/kornia-pnp/src/camera.rs
@@ -1,5 +1,4 @@
 //! Camera models and distortion handling for PnP solvers.
-
 use thiserror::Error;
 
 /// Error types for camera operations.
@@ -69,6 +68,7 @@ impl CameraIntrinsics {
 
 /// Represents polynomial distortion parameters using the Brown-Conrady model.
 #[derive(Debug, Clone, PartialEq)]
+#[allow(missing_docs)]
 pub struct PolynomialDistortion {
     /// Radial distortion coefficients
     pub k1: f32,
@@ -159,7 +159,7 @@ impl CameraModel {
     pub fn undistort_points(&self, points: &[[f32; 2]]) -> CameraResult<Vec<[f32; 2]>> {
         points
             .iter()
-            .map(|&[x, y]| self.undistort_point(x, y))
+            .map(|&[x, y]| self.undistort_point(x, y).map(|(ux, uy)| [ux, uy]))
             .collect()
     }
 
@@ -176,7 +176,7 @@ impl CameraModel {
     pub fn distort_points(&self, points: &[[f32; 2]]) -> CameraResult<Vec<[f32; 2]>> {
         points
             .iter()
-            .map(|&[x, y]| self.distort_point(x, y))
+            .map(|&[x, y]| self.distort_point(x, y).map(|(dx, dy)| [dx, dy]))
             .collect()
     }
 
@@ -360,3 +360,4 @@ mod tests {
         assert!((original_point[1] - undistorted.1).abs() < 1e-3);
     }
 }
+
--- a/crates/kornia-pnp/src/camera.rs
+++ b/crates/kornia-pnp/src/camera.rs
@@ -1,0 +1,362 @@
+//! Camera models and distortion handling for PnP solvers.
+
+use thiserror::Error;
+
+/// Error types for camera operations.
+#[derive(Debug, Error)]
+pub enum CameraError {
+    /// Invalid camera intrinsics matrix
+    #[error("Invalid camera intrinsics matrix: {0}")]
+    InvalidIntrinsics(String),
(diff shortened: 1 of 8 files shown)


crates/kornia-pnp/src/pnp.rs
```diff
@@ -1,5 +1,6 @@
 //! Common data types shared across Perspective-n-Point (PnP) solvers.
 
+use crate::camera::CameraModel;
```
cjpurackal, line 3: @jeffin07 I don't think we need have an explicit camera model here. You just need to import the distortion functions from the [kornia-imgproc](https://github.com/kornia/kornia-rs/blob/main/crates/kornia-imgproc/src/calibration/distortion.rs) crate and use it with the distortion coefficients from the `solve` function signature. Check opencv pnp here:
https://github.com/opencv/opencv/blob/7a4bd852997274c56b8e883016f69a4c955e3a35/modules/calib3d/src/solvepnp.cpp#L132

cjpurackal:
We also don't need a separate example, just need to update the current pnp example with the optional distortion coefficients.

crates/kornia-pnp/src/camera.rs
```diff
@@ -0,0 +1,392 @@
...
+    pub k5: f32,
+    pub k6: f32,
+    /// Tangential distortion coefficients
+    pub p1: f32,
+    pub p2: f32,
+}
+
+impl PolynomialDistortion {
+    /// Create distortion parameters with all coefficients set to zero (no distortion).
+    pub fn none() -> Self {
```
edgarriba, line 88:
```suggestion
    pub fn empty() -> Self {
```
or zeros ?

crates/kornia-pnp/src/camera.rs
```diff
@@ -0,0 +1,392 @@
...
+pub struct CameraModel {
+    /// Camera intrinsics
+    pub intrinsics: CameraIntrinsics,
+    /// Distortion parameters (None for no distortion)
+    pub distortion: Option<PolynomialDistortion>,
+}
+
+impl CameraModel {
+    /// Create a camera model without distortion.
+    pub fn pinhole(intrinsics: CameraIntrinsics) -> Self {
```
edgarriba, line 153: wondering if the type should be known at compile time instead to use full rust type safe checks semantics /cc @cjpurackal

crates/kornia-pnp/src/camera.rs
```diff
@@ -0,0 +1,392 @@
...
+    fn test_distort_undistort_roundtrip() {
+        let intrinsics = CameraIntrinsics::new(1000.0, 1000.0, 640.0, 480.0);
+        let distortion = PolynomialDistortion::radial(0.1, 0.01);
+        let camera = CameraModel::with_distortion(intrinsics, distortion);
+
+        let original_point = [100.0, 200.0];
+        let distorted = camera
+            .distort_point(original_point[0], original_point[1])
+            .unwrap();
+        let undistorted = camera.undistort_point(distorted.0, distorted.1).unwrap();
```
edgarriba, line 386: i'd suggest to test against numbers from known libraries

crates/kornia-pnp/src/epnp.rs
```diff
@@ -148,6 +151,40 @@ pub fn solve_epnp(
     })
 }
 
+// /// Solve Perspective-n-Point (EPnP) with camera model support.
```
edgarriba, line 154: no commented codes

crates/kornia-pnp/src/camera.rs
```diff
@@ -0,0 +1,392 @@
...
+    #[test]
+    fn test_distort_undistort_roundtrip() {
+        let intrinsics = CameraIntrinsics::new(1000.0, 1000.0, 640.0, 480.0);
+        let distortion = PolynomialDistortion::radial(0.1, 0.01);
+        let camera = CameraModel::with_distortion(intrinsics, distortion);
+
+        let original_point = [100.0, 200.0];
+        let distorted = camera
+            .distort_point(original_point[0], original_point[1])
+            .unwrap();
```
edgarriba, line 385: no unwraps

examples/pnp_demo/src/main.rs
```diff
@@ -95,7 +95,17 @@ fn main() -> Result<(), Box<dyn std::error::Error>> {
         .collect();
 
     // Run EPnP
-    let result = kpnp::solve_pnp(&world_pts, &image_pts, &k, kpnp::PnPMethod::EPnPDefault)?;
+    let distortion = kpnp::PolynomialDistortion {
```
edgarriba, line 98: how can we verify correctness here; can we provide  any image with that distiortion ?

PR #460, opened 2025-08-29
Implement newer FAST Corner Detector

AS1100K:
image

```bash
$ cargo run -p fast_detector --release -- --image-path "path/to/image.png" --min-distance 10 --threshold 0.15
```

## Benchmark

1. NVIDIA Orin
    ```
    Before:
    FastCornerDetect/fast_native_cpu/1920x1080
                            time:   [2.3486 ms 2.3577 ms 2.3676 ms]
    
    After:
    FastCornerDetect/fast_native_cpu/1920x1080
                            time:   [6.0860 ms 6.1104 ms 6.1371 ms]
    ``` 

2. AMD Ryzen 7
    ```
    Before:
    FastCornerDetect/fast_native_cpu/1920x1080
                            time:   [1.3061 ms 1.3108 ms 1.3156 ms]
    
    After:
    FastCornerDetect/fast_native_cpu/1920x1080
                        time:   [1.9339 ms 1.9433 ms 1.9527 ms]
    ```

## TODOs

- [x] Add Docs
- [x] Add Tests

### DIFF
--- a/.devcontainer/devcontainer.json
+++ b/.devcontainer/devcontainer.json
@@ -1,7 +1,6 @@
 {
     "name": "Rust",
     "image": "mcr.microsoft.com/devcontainers/rust:1-1",
-
     // Set customizations for the container.
     "customizations": {
         "vscode": {
--- a/.github/workflows/python_test.yml
+++ b/.github/workflows/python_test.yml
@@ -26,37 +26,34 @@ jobs:
       matrix:
         python-version: ["3.8", "3.9", "3.10", "3.11", "3.12", "3.13"]
     steps:
-    - run: sudo apt-get update
-    - uses: actions/checkout@v4
+      - run: sudo apt-get update
+      - uses: actions/checkout@v4
 
-    - uses: extractions/setup-just@v2
+      - uses: extractions/setup-just@v2
 
-    - name: Set up Python ${{ matrix.python-version }}
-      uses: astral-sh/setup-uv@v5
-      with:
-        python-version: ${{ matrix.python-version }}
-    - uses: actions-rs/toolchain@v1
-      with:
-        toolchain: stable
-        profile: minimal
+      - name: Set up Python ${{ matrix.python-version }}
+        uses: astral-sh/setup-uv@v5
+        with:
+          python-version: ${{ matrix.python-version }}
+      - uses: actions-rs/toolchain@v1
+        with:
+          toolchain: stable
+          profile: minimal
 
-    - name: Install sccache
-      uses: mozilla-actions/sccache-action@v0.0.8
+      - name: Install sccache
+        uses: mozilla-actions/sccache-action@v0.0.9
 
-    - name: Install system dependencies
-      run: sudo apt-get install -y nasm libunwind-dev
+      - name: Install system dependencies
+        run: sudo apt-get install -y nasm libunwind-dev
 
-    - name: Setup project environment
-      run: cd kornia-py/ && just install ${{ matrix.python-version }}
+      - name: Setup project environment
+        run: cd kornia-py/ && just install ${{ matrix.python-version }}
 
-    - name: Run tests
-      run: cd kornia-py/ && just test
+      - name: Run tests
+        run: cd kornia-py/ && just test
 
-
-    - name: Show sccache stats
-      run: sccache --show-stats
-
-
+      - name: Show sccache stats
+        run: sccache --show-stats
 
   # TODO: integrate with linux tests once uv bug is fixed
   # https://github.com/astral-sh/uv/issues/12445
@@ -71,44 +68,37 @@ jobs:
       SCCACHE_ERROR_LOG: /tmp/sccache_log.txt
       SCCACHE_LOG: info
 
-
     steps:
-    - run: sudo apt-get update
-    - uses: actions/checkout@v4
+      - run: sudo apt-get update
+      - uses: actions/checkout@v4
 
-    - uses: extractions/setup-just@v2
+      - uses: extractions/setup-just@v2
 
-    - uses: astral-sh/setup-uv@v5
+      - uses: astral-sh/setup-uv@v5
 
-    - uses: actions-rs/toolchain@v1
-      with:
-        toolchain: stable
-        profile: minimal
+      - uses: actions-rs/toolchain@v1
+        with:
+          toolchain: stable
+          profile: minimal
 
+      - name: Install sccache
+        uses: mozilla-actions/sccache-action@v0.0.9
 
-    - name: Install sccache
-      uses: mozilla-actions/sccache-action@v0.0.8
+      - name: Install system dependencies
+        run: sudo apt-get install -y nasm libunwind-dev
 
+      - name: Setup project environment
+        run: cd kornia-py/ && just install-dev 3.13t
 
+      - name: Run tests
+        run: |
+          cd kornia-py/
+          source .venv/bin/activate
+          maturin develop -m Cargo.toml
+          pytest --parallel-threads=8 --iterations=50
 
-    - name: Install system dependencies
-      run: sudo apt-get install -y nasm libunwind-dev
-
-    - name: Setup project environment
-      run: cd kornia-py/ && just install-dev 3.13t
-
-    - name: Run tests
-      run: |
-        cd kornia-py/
-        source .venv/bin/activate
-        maturin develop -m Cargo.toml
-        pytest --parallel-threads=8 --iterations=50
-
-
-    - name: Show sccache stats
-      run: sccache --show-stats
-
-
+      - name: Show sccache stats
+        run: sccache --show-stats
 
   # NOTE: there's a systematic fail with the ci on macos
   # test-python-macos:
--- a/.github/workflows/rust_lint.yml
+++ b/.github/workflows/rust_lint.yml
@@ -29,11 +29,9 @@ jobs:
     steps:
       - run: sudo apt-get update
       - uses: actions/checkout@v4
-      - uses: dtolnay/rust-toolchain@stable
-        with:
-          components: clippy
+      - run: rustup update
       - name: Install sccache
-        uses: mozilla-actions/sccache-action@v0.0.8
+        uses: mozilla-actions/sccache-action@v0.0.9
 
       - name: Install cross
         uses: taiki-e/install-action@cross
@@ -56,10 +54,10 @@ jobs:
     steps:
       - run: sudo apt-get update
       - uses: actions/checkout@v4
-      - run: rustup toolchain install stable
+      - run: rustup update
 
       - name: Install sccache
-        uses: mozilla-actions/sccache-action@v0.0.8
+        uses: mozilla-actions/sccache-action@v0.0.9
 
       - name: Install cross
         uses: taiki-e/install-action@cross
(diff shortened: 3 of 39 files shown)


AS1100K:
@edgarriba for the KdTree, should we use the [`kiddo::fixed::KdTree`](https://docs.rs/kiddo/latest/kiddo/fixed/kdtree/struct.KdTree.html) or [`ImmutableKdTree`](https://docs.rs/kiddo/latest/kiddo/type.ImmutableKdTree.html).

The problem with the fixed KdTree was the generic constant `B` i.e. bucket size which can cause panic depending on image, threshold, and arc_length, while the `ImmutableKdTree` uses floating points i.e. we need to use f32 for the keypoints pos.

[_Commit with fixed KdTree_](https://github.com/kornia/kornia-rs/pull/460/commits/42842f84e971c3d17db000c67b51aa5c704e287d)

CC: @cjpurackal

AS1100K:
Nvm, `corner_peaks` function is pretty much useless so we can avoid the concept of KdTree entirely

examples/fast_detector/src/main.rs
```diff
@@ -50,12 +66,15 @@ fn main() -> Result<(), Box<dyn std::error::Error>> {
     // log the keypoints
     let points = keypoints
         .iter()
-        .map(|k| (k[0] as f32, k[1] as f32))
+        .map(|k| (k[1] as f32, k[0] as f32))
```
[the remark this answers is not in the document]
AS1100K: We are returning the indices in `(y, x)` format while rerun accepts `(x, y)` format

crates/kornia-imgproc/src/features/fast.rs
```diff
@@ -1,162 +1,353 @@
...
+    ///
+    /// * `src` - The source grayscale image.
+    ///
+    /// # Returns
+    ///
+    /// Returns a reference to the image containing the corner response.
+    pub fn compute_corner_response<A: ImageAllocator>(
+        &mut self,
+        src: &Image<f32, 1, A>,
+    ) -> &Image<f32, 1, CpuAllocator> {
```
cjpurackal: add an Image size safety check with `self.corner_reponse.size()` here and return a result

cjpurackal (approved):

crates/kornia-imgproc/src/features/fast.rs
```diff
@@ -1,162 +1,362 @@
...
-        } else if pixel < lower_threshold {
-            consecutive_darker += 1;
-            consecutive_brighter = 0;
+    /// Extracts keypoints from the computed corner response.
+    ///
+    /// # Returns
+    ///
+    /// Returns a `Result` containing a vector of keypoint coordinates or an `ImageError`.
+    pub fn extract_keypoints(&mut self) -> Result<Vec<[usize; 2]>, ImageError> {
+        get_peak_mask(&self.corner_response, &mut self.mask, self.threshold);
```
edgarriba: can any of this loops be fused ?

crates/kornia-imgproc/src/features/fast.rs
```diff
@@ -1,162 +1,362 @@
...
+) -> Vec<[usize; 2]> {
+    let src_size = src.size();
+    let width = src_size.width;
+    let height = src_size.height;
+
+    let coords_intensity: Vec<[usize; 2]> = mask
+        .as_slice()
+        .par_iter()
+        .enumerate()
+        .filter(|&(_, &value)| value)
```
edgarriba: why not use directly the map?

crates/kornia-imgproc/src/features/fast.rs
```diff
@@ -1,162 +1,362 @@
...
+        .map(|(i, _)| {
+            let y = i / width;
+            let x = i % width;
+            [y, x]
+        })
+        .collect();
+
+    let mut result = Vec::new();
+
+    for coord in coords_intensity {
```
edgarriba:
```suggestion
    for (y, x) in coords_intensity {
```

crates/kornia-imgproc/src/features/fast.rs
```diff
@@ -1,162 +1,365 @@
...
+            taken: vec![false; image_size.height * image_size.width],
         })
-        .collect();
+    }
 
-    Ok(keypoints)
-}
+    /// Clears the internal state of the detector, marking it ready to detect again.
+    pub fn clear(&mut self) {
+        self.taken.par_iter_mut().for_each(|px| *px = false);
```
edgarriba: this is not compute intensive, i would use the non par iter

crates/kornia-imgproc/src/features/fast.rs
```diff
@@ -1,162 +1,365 @@
...
+        result.push([y, x]);
+
+        // Suppress all within min_distance
+        let y0 = y.saturating_sub(min_distance);
+        let y1 = (y + min_distance + 1).min(height);
+        let x0 = x.saturating_sub(min_distance);
+        let x1 = (x + min_distance + 1).min(width);
+
+        (y0..y1)
+            .flat_map(|yy| (x0..x1).map(move |xx| (yy, xx)))
```
edgarriba: is there a way to simplify the flat_map/filter/for_each ? looks a bit overcomplicated

crates/kornia-imgproc/src/features/fast.rs
```diff
@@ -1,162 +1,365 @@
...
-
-    let m0 = !(p5 >= lower_threshold || p1 >= lower_threshold && p9 >= lower_threshold);
-    let m1 = !(p5 <= upper_threshold || p1 <= upper_threshold && p9 <= upper_threshold);
-
-    if !m0 && !m1 {
-        return false;
+        const RP: [isize; 16] = [0, 1, 2, 3, 3, 3, 2, 1, 0, -1, -2, -3, -3, -3, -2, -1];
+        const CP: [isize; 16] = [3, 3, 2, 1, 0, -1, -2, -3, -3, -3, -2, -1, 0, 1, 2, 3];
+
+        corner_response[3 * width..(height - 3) * width]
```
edgarriba, line 84: add comment of what's happening here; i guess we are skipping the +-3 top/down borders (?)

crates/kornia-imgproc/src/features/fast.rs
```diff
@@ -1,162 +1,365 @@
...
+                Some([y, x])
+            } else {
+                None
+            }
+        })
+        .collect();
+
+    let mut result = Vec::new();
+
+    for [y, x] in coords_intensity {
```
edgarriba: wondering if this logic can be concatenated after the .filter_map with another builtin function

edgarriba:
any ideas what make the benchmark worst ? can we compare also with an existing implementation?

AS1100K:
> any ideas what make the benchmark worst ? can we compare also with an existing implementation?

I suspect that the previous implementation was in `u8` and this one is in `f32` _(because this is needed for ORB implementation which needs to be in `f32` as gaussian blur and harris response are in `f32`)_. `u8` us 4x smaller than `f32` which could result in more cache hit and uses faster CPU instructions for integer types.

I did tried to improve the performance ~15% (7ms to 6ms) but I don't think we have more room for improvements without SIMD


Benchmark against previous implementation on NVIDIA Orin
```
Before:
FastCornerDetect/fast_native_cpu/1920x1080
                        time:   [2.3486 ms 2.3577 ms 2.3676 ms]
   
After:
FastCornerDetect/fast_native_cpu/1920x1080
                        time:   [6.0860 ms 6.1104 ms 6.1371 ms]
```

crates/kornia-imgproc/src/features/fast.rs
```diff
@@ -1,162 +1,332 @@
...
+    /// * `image_size` - The size of the image to process.
+    /// * `threshold` - The intensity threshold for detecting corners.
+    /// * `arc_length` - The minimum arc length for a sequence of contiguous pixels to be considered a corner.
+    /// * `min_distance` - The minimum distance between detected keypoints.
+    ///
+    /// # Returns
+    ///
+    /// Returns a `Result` containing the new `FastDetector` or an `ImageError`.
+    pub fn new(
+        image_size: ImageSize,
```
edgarriba:
```suggestion
        size: ImageSize,
```

crates/kornia-imgproc/src/features/fast.rs
```diff
@@ -1,162 +1,332 @@
...
+            min_distance,
+            arc_length,
+            corner_response: Image::from_size_val(image_size, 0.0, CpuAllocator)?,
+            taken: vec![false; image_size.height * image_size.width],
+        })
+    }
+
+    /// Clears the internal state of the detector, marking it ready to detect again.
+    pub fn clear(&mut self) {
+        self.taken.iter_mut().for_each(|px| *px = false);
```
edgarriba, line 53: and what about the corner_response, no need to reset ?

crates/kornia-imgproc/src/features/fast.rs
```diff
@@ -1,162 +1,332 @@
...
+    ///
+    /// * `src` - The source grayscale image.
+    ///
+    /// # Returns
+    ///
+    /// Returns a reference to the image containing the corner response.
+    pub fn compute_corner_response<A: ImageAllocator>(
+        &mut self,
+        src: &Image<f32, 1, A>,
+    ) -> Result<&Image<f32, 1, CpuAllocator>, ImageError> {
```
edgarriba: why do you need to return the image ? other parts of the pipeline can just access that. If the user for some reason need it, we can provide a getter method

crates/kornia-imgproc/src/features/fast.rs
```diff
@@ -1,162 +1,332 @@
...
+        Ok(&self.corner_response)
+    }
+
+    /// Extracts keypoints using Non-Maximum Suppression on the corner response.
+    ///
+    /// # Returns
+    ///
+    /// Returns a vector of keypoint coordinates `[y, x]`.
+    pub fn extract_keypoints(&mut self) -> Vec<[usize; 2]> {
+        get_high_intensity_peaks(
```
edgarriba, line 167: why not directly calling the function extract_keypoints ? i'd avoid having an overloaded method and use directly the function where is needed

AS1100K: Do you mean having a single function instead of two `compute_corner_response` and `extract_keypoints`

edgarriba: to make `get_high_intensity_peaks` public or even rename that to `extract_keypoints`

crates/kornia-imgproc/src/features/fast.rs
```diff
@@ -1,162 +1,328 @@
...
+    /// * `image_size` - The size of the image to process.
+    /// * `threshold` - The intensity threshold for detecting corners.
+    /// * `arc_length` - The minimum arc length for a sequence of contiguous pixels to be considered a corner.
+    /// * `min_distance` - The minimum distance between detected keypoints.
+    ///
+    /// # Returns
+    ///
+    /// Returns a `Result` containing the new `FastDetector` or an `ImageError`.
+    pub fn new(
+        size: ImageSize,
```
edgarriba, line 37: do you think it makes sense here to pass instead a `config: FastDetectorConfig` -- or to have a builder pattern with presets ?

edgarriba:
Should we merge ?

edgarriba:
@cjpurackal @sidd-27 maybe to find someone else to work on this

PR #470, opened 2025-09-05
interpolation fix

andrew-shc:


crates/kornia-imgproc/src/interpolation/interpolate.rs
```diff
@@ -35,5 +37,8 @@ pub fn interpolate_pixel<const C: usize, A: ImageAllocator>(
     match interpolation {
         InterpolationMode::Bilinear => bilinear_interpolation(image, u, v, c),
         InterpolationMode::Nearest => nearest_neighbor_interpolation(image, u, v, c),
+        InterpolationMode::Lanczos => {
+            unimplemented!("Lanczos interpolation is not yet implemented")
```
edgarriba, line 41: ideally this should throw an error instead. You can submit a new PR for that

andrew-shc: do you want this to leave it as an issue to-be-implemented? wrapping this with an error type might incur performance cost since it's called per pixels for each images (which is a lot, even with parallelization)

PR #471, opened 2025-09-05
SmolVLM example + validations

andrew-shc:


andrew-shc:
@edgarriba do you want the validation code to be split on a separate repo? i also have our custom validation code that's not in the git but doesn't fit well with the repo

andrew-shc:
placed validation code here: https://github.com/kornia/VLMEvalKit/tree/main

Issue #461, opened 2025-08-30
Add Gaussian Blur support for u8 images


AS1100K:
Currently, the Gaussian Blur functionality is only implemented for `f32` images.

It would be beneficial to add support for `u8` images as well. This would enable broader usage scenarios and improve the usability of the Gaussian Blur functionality for typical image processing pipelines where `u8` images are common.

gulshan-123:
Hi @AS1100K, I am new to kornia rust. I have read contributing guidelines and want to work on this issue.

AS1100K:
Go ahead @gulshan-123

gulshan-123:
Hi @AS1100K, I have opened a PR, please review

afrdbaig7:
Hi I would like to take this issue
If there’s no one else currently assigned, could you please assign it to me
Thanks

afrdbaig7:
@edgarriba 
Update: Finalized Fix for Issue #461
I’ve pushed the final implementation to add u8 support for gaussianblur (and boxblur for consistency).

Changes:
Generic API:
Updated gaussianblur and boxblur to be generic over the pixel type, exposing the existing multi-type support already implemented in separablefilter via FloatConversion.

Cleanup:
Removed the now-redundant gaussian_blur_u8 wrapper and updated benchmarks to use the unified generic API.

Testing:
Added a regression test to verify correct u8 gaussian blur behavior, including proper clamping. All existing tests pass, and benchmarks compile successfully.

Nyx128:
Hi, i would like to try this issue if nobody is working on it. As i saw in the previous PRs, float conversion will only bottleneck performance so here is a list of things i will do:

- Add fixed-point Gaussian kernel generation .

- Implement a separable u8 convolution using widened accumulators (u32) and bit-shift normalization.

- Expose a new gaussian_blur_u8 API while keeping existing float paths unchanged.

- write tests to validate against the current f32 implementation within certain tolerances we can discuss upon.

I also plan to implement SIMD operations for this, after my initial working functions. I will try to benchmark and document improvements best i can.

Please let me know if my plan sounds reasonable.

RakshithaKowlikar:
I've completed the box_blur optimization (merge pending) as discussed. I will be moving on to gaussian_blur for u8 that uses this. Would it be possible to assign this issue to me?

AdityaShome:
[...]

PR #475, opened 2025-09-06
Added Support for u8 images for Gaussian Blur

gulshan-123:
Solves #461

I have added template programming for supporting u8 as well as f32 similar in structure what was given in separable_filter.rs
I have also added a test similar to a f32 gaussian blur test. The output for the u8 gaussian blur will be the image matrix rounded towards zero.

Please suggest if there is anything left to be done.

### DIFF
--- a/crates/kornia-imgproc/src/filter/ops.rs
+++ b/crates/kornia-imgproc/src/filter/ops.rs
@@ -589,4 +589,32 @@ mod tests {
 
         Ok(())
     }
+
+    #[test]
+    fn test_gaussian_blur_u8() -> Result<(), ImageError> {
+        let size = ImageSize {
+            width: 5,
+            height: 5,
+        };
+
+        let img = Image::new(size, (0..25).map(|x| x as u8).collect(), CpuAllocator)?;
+
+        let mut dst = Image::<_, 1, _>::from_size_val(size, 0u8, CpuAllocator)?;
+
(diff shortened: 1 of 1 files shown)


AS1100K (changes requested):
I am afraid this might not be that simple. Converting the integer type to floating seems correct on the surface but introduces bottlenecks on performance which we are trying to avoid. The goal of using `u8` and other integer types is to leverage faster integer instructions and a smaller memory footprint, but a direct conversion to `f32` neglects those advantages.

crates/kornia-imgproc/src/filter/ops.rs
```diff
@@ -585,4 +589,32 @@ mod tests {
...
     }
+
+    #[test]
+    fn test_gaussian_blur_u8() -> Result<(), ImageError> {
+        let size = ImageSize {
+            width: 5,
+            height: 5,
+        };
+
+        let img = Image::new(size, (0..25).map(|x| x as u8).collect(), CpuAllocator)?;
```
edgarriba, line 600: as @AS1100K mention this will overflow e.g in case of many clusters of pixels with high values

gulshan-123: I think, I had misunderstood the approach, so will search how other libraries implemented specifically for u8

Issue #478, opened 2025-09-07
Adding Support for Otsu Thresholding


gulshan-123:
Similar to manual binary thresholding, but the threshold value is computed by maximizing inter-class variance

Reference: https://en.wikipedia.org/wiki/Otsu%27s_method
https://www.ipol.im/pub/art/2016/158/article_lr.pdf
Purpose: Common in teachings

TODO: How other libraries implemented it for multi-channel (eg: RGB) thresholding

Found PR related to this: #357

PR #479, opened 2025-09-08
Supporting Otsu threshold

gulshan-123:
Added an initial version of otsu threshold. It only supports single channel u8 images.
Solves #478 
Implementation according to the algorithm given in the paper: https://www.ipol.im/pub/art/2016/158/article_lr.pdf

### DIFF
--- a/crates/kornia-imgproc/src/threshold.rs
+++ b/crates/kornia-imgproc/src/threshold.rs
@@ -378,7 +378,6 @@ where
 pub fn threshold_otsu<A1: ImageAllocator, A2: ImageAllocator>(
     src: &Image<u8, 1, A1>,
     dst: &mut Image<u8, 1, A2>,
-    max_value: u8,
 ) -> Result<(), ImageError>
 {
     if src.size() != dst.size() {
@@ -427,7 +426,7 @@ pub fn threshold_otsu<A1: ImageAllocator, A2: ImageAllocator>(
 
     parallel::par_iter_rows_val(src, dst, |src_pixel, dst_pixel| {
         *dst_pixel = if *src_pixel > thresh {
-            max_value
+            255
         } else {
             0u8
         };
--- a/crates/kornia-imgproc/src/threshold.rs
+++ b/crates/kornia-imgproc/src/threshold.rs
@@ -374,6 +374,68 @@ where
 
 // TODO: outsu, triangle
 
+/// Apply an otsu threshold to an image
+pub fn threshold_otsu<A1: ImageAllocator, A2: ImageAllocator>(
+    src: &Image<u8, 1, A1>,
+    dst: &mut Image<u8, 1, A2>,
+    max_value: u8,
+) -> Result<(), ImageError>
+{
+    if src.size() != dst.size() {
+        return Err(ImageError::InvalidImageSize(
+            src.cols(),
+            src.rows(),
+            dst.cols(),
+            dst.rows(),
+        ));
+    }
+    let mut hist: Vec<usize>= vec![0_usize; 256];
+    let _ = compute_histogram(src, &mut hist, 256).unwrap();
+
+    let total = (src.rows() * src.cols()) as f64;
+
+    let mut sum_all = 0f64;
+    for i in 0..256 {
+        sum_all += (i as f64) * (hist[i] as f64);
+    }
+
+    let mut sum_b = 0f64;
+    let mut w_b = 0f64;
+    let mut var_max = 0f64;
+    let mut thresh = 0u8;
+
+    for t in 0..256 {
+        w_b += hist[t] as f64;
+        if w_b == 0.0 { continue; }
+
+        let w_f = total - w_b;
+        if w_f == 0.0 { break; }
+
+        sum_b += (t as f64) * (hist[t] as f64);
+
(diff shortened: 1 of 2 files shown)


crates/kornia-imgproc/src/threshold.rs
```diff
@@ -374,6 +374,66 @@ where
 
 // TODO: outsu, triangle
 
+/// Apply an otsu threshold to an image
+pub fn threshold_otsu<A1: ImageAllocator, A2: ImageAllocator>(
```
edgarriba, line 378: provide a test comparing with an exsiting library to match compatibility

gulshan-123:
Hi @edgarriba, I have added one test for the otsu thresholding. I have added to return the otsu threshold value computed.
On my side, I have tested for all the images present on the [demo of paper you sent](https://ipolcore.ipol.im/demo/clientApp/demo.html?id=158&key=FD4EE592DAF214DFEE6AC32108F47310). All are successful, but included only 1 in the test.

The values were confirmed from python opencv implementation, the snippet for checking value is:
```python3
import cv2
image = cv2.imread('input_1.png', cv2.IMREAD_GRAYSCALE)
thresh_value, otsu_img = cv2.threshold(image, 0, 255, cv2.THRESH_BINARY + cv2.THRESH_OTSU)
print(f"Otsu Threshold: {thresh_value}")
```

EDIT: Earlier some files were mistakenly edited, so removed them.

crates/kornia-imgproc/src/threshold.rs
```diff
@@ -374,11 +374,72 @@ where
...
+    src: &Image<u8, 1, A1>,
+    dst: &mut Image<u8, 1, A2>,
+    threshold_out: &mut u8,
```
edgarriba, line 381: why do we need the threshold output ?

crates/kornia-imgproc/src/threshold.rs
```diff
@@ -374,11 +374,72 @@ where
...
+    if src.size() != dst.size() {
+        return Err(ImageError::InvalidImageSize(
+            src.cols(),
+            src.rows(),
+            dst.cols(),
+            dst.rows(),
+        ));
+    }
+    let mut hist: Vec<usize> = vec![0_usize; 256];
+    let _ = compute_histogram(src, &mut hist, 256).unwrap();
```
edgarriba, line 392: no unwraps

crates/kornia-imgproc/src/threshold.rs
```diff
@@ -374,11 +374,72 @@ where
...
+
+    let n = (src.rows() * src.cols()) as f64;
+
+    let mut sum = 0f64;
+    for i in 0..256 {
```
edgarriba, line 397:
```suggestion
    for i in 0..HIST_SIZE {
```
no magic numbers, use consts. same for the rest

crates/kornia-imgproc/src/threshold.rs
```diff
@@ -374,11 +374,72 @@ where
...
+        // maximize inter class variance
+        let var_between = q1 * q2 * (mu_1 - mu_2).powi(2);
+
+        if var_between > var_max {
+            thresh = t as u8;
+            var_max = var_between;
+        }
+    }
+    *threshold_out = thresh;
+    parallel::par_iter_rows_val(src, dst, |src_pixel, dst_pixel| {
```
edgarriba, line 428: there's a `threshold` set of reusable functions

tests/data/barabara.png
edgarriba, line 1: now new images to be added for testing for now, we want to keep low the git tree size

crates/kornia-imgproc/src/threshold.rs
```diff
@@ -534,4 +595,14 @@ mod tests {
 
         Ok(())
     }
+
+    #[test]
+    fn test_threshold_otsu() -> Result<(), ImageError> {
```
edgarriba, line 600: provide a numerical test comparing to any known library such opencv, or scikit-image

edgarriba (changes requested):

crates/kornia-imgproc/src/threshold.rs
gulshan-123: I thought people may use it for comparison with manual thresholding in learning purpose.. Also, I found that both opencv and scikit-images return this
Should we remove it?

edgarriba: I see, we can mimic those libraries then. I'm not very convinced that the given threshold is updated as it can induce errors if someone wants to reuse later the variable with the initial value. What about this

/cc @cjpurackal @AS1100K 

**Option1** 

```rust
pub fn threshold_otsu<A1: ImageAllocator, A2: ImageAllocator>(
   src: &Image<u8, 1, A1>,
   dst: &mut Image<u8, 1, A2>,
   threshold: u8,
   threshold_out: &mut u8,
) -> Result<(), ImageError> { ... }
```

**Option2**

```rust
pub fn threshold_otsu<A1: ImageAllocator, A2: ImageAllocator>(
   src: &Image<u8, 1, A1>,
   dst: &mut Image<u8, 1, A2>,
   threshold: u8,
) -> Result<u8, ImageError> { ... }
```

PR #490, opened 2025-09-21
Adding Bicubic interpolation

gulshan-123:
I have added code for bicubic interpolation. I am now in process to test the function on some standard images.

### DIFF
--- a/crates/kornia-imgproc/src/interpolation/bicubic.rs
+++ b/crates/kornia-imgproc/src/interpolation/bicubic.rs
@@ -15,8 +15,8 @@ pub(crate) fn bicubic_interpolation<const C: usize, A: ImageAllocator>(
     let y = v - iv as f32;
 
     let get = |row: usize, col: usize| -> f32 {
-        let row = row.clamp(0, rows-1);
-        let col = col.clamp(0, cols-1);
+        let row = row.clamp(0, rows - 1);
+        let col = col.clamp(0, cols - 1);
 
         *image.get_unchecked([row, col, c])
     };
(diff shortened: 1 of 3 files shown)


crates/kornia-imgproc/src/interpolation/bicubic.rs
```diff
@@ -0,0 +1,90 @@
...
+    let mut fy = [[0.0; 2]; 2]; // derivative in y dir
+    let mut fxy = [[0.0; 2]; 2]; // derivative in xy
+
+    // using central derivative approximative
+    for j in 0..2 {
+        let rr = iv + j as usize;
+        for i in 0..2 {
+            let cc = iu + i as usize;
+            f[j][i] = get(rr, cc);
+            fx[j][i] = (get(rr, cc + 1) - get(rr, cc - 1)) * 0.5;
```
edgarriba, line 35: the `get` is an expensive op as you compute each time the linear index to the underlaying array storage. For continuous postions (in the x direction) you can just add or substract positions.

crates/kornia-imgproc/src/interpolation/bicubic.rs
```diff
@@ -0,0 +1,90 @@
...
+    let mut matrix_a = [[0.0; 4]; 4];
+
+    // M*F = temp
+    for i in 0..4 {
+        for j in 0..4 {
+            for k in 0..4 {
+                temp[i][j] += matrix_m[i][k] * matrix_f[k][j];
+            }
+        }
+    }
```
edgarriba, line 68:
```suggestion
    }

```

edgarriba (changes requested):
add a numerical test to compare against a known library

PR #511, opened 2025-10-12
optimized compute_centroid by 250x

sidd-27:
removed the creation of a new faer column vector at each iteration, which saves unnecessary heap allocations, in my benchmarks, this gives a speedup of about 250x on inputs of size 10k points

for 100 runs
Original:  1.860s total (18.600ms/run)
Optimized: 8.232ms total (82.323µs/run)
🚀 Speedup: 225.94x faster

### DIFF
--- a/crates/kornia-icp/src/ops.rs
+++ b/crates/kornia-icp/src/ops.rs
@@ -66,17 +66,19 @@ pub(crate) fn compute_centroids(
     points1: &[[f64; 3]],
     points2: &[[f64; 3]],
 ) -> (faer::Col<f64>, faer::Col<f64>) {
-    let mut centroid1 = faer::Col::zeros(3);
-    let mut centroid2 = faer::Col::zeros(3);
-
+    let (mut c1_x, mut c1_y, mut c1_z) = (0.0, 0.0, 0.0);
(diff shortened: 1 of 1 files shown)


sidd-27:
floating point error on a test i think

cjpurackal:
> I'd suggest that for consistency we try to use glam-rs instead of faer-rs. And see what's the performance there @cjpurackal @AS1100K
> 
> i think also that this function should be moved to a shared crate like linalg in a future PR. @cjpurackal ?

Yes, this should be inside linalg. This func is duplicated in `crates/kornia-pnp/src/ops.rs` as well

sidd-27:
Should I move these two to the lin alg crate and remove the duplicate?

cjpurackal:
> Should I move these two to the lin alg crate and remove the duplicate?

go for it

sidd-27:
so one of the functions calculates for two point clouds at once and the other just does for one point cloud, i would've kept the one at a time version but the two at a time version is 1.3x faster than just calculating the centroids seperately

edgarriba:
can we return glam Vec3 ?

edgarriba:

Issue #513, opened 2025-10-14
Error in the readme example for loading and visualizing the image in kornia 0.19 version


Astik-2002:
lines with `Image::<x, C:y, _>` give the error `struct takes 2 generic arguments but 3 generic arguments were supplied`.

I think the correct to write would be `Image::<x, C:y>`

pr made with corrections #514

PR #514, opened 2025-10-14
fix readme examples (issue #513)

Astik-2002:
Updated the readme example for correct syntax for defining an Image variable for kornia 0.19.0


edgarriba:
main branch is pointing to >0.1.9 (0.10.*) which already includes this changes. If we change this it will mislead versions

Astik-2002:
@edgarriba Got it. I'll close this pr since the changes are already present

PR #515, opened 2025-10-14
refactor: optimized covariance matrix computation by 140x

dhruvkjain:
Optimized Covariance matrix calculation (~140x) by avoiding repeated heap allocation of `faer::col` which was causing overhead in each iteration of loop 

for 100 iterations of benchmark:
```shell
unoptimized covariance time: [2.3402 ms 2.3502 ms 2.3633 ms] 
Found 5 outliers among 100 measurements (5.00%) 
2 (2.00%) high mild 
3 (3.00%) high severe 

optimized covariance time: [16.595 µs 16.791 µs 17.060 µs] 
Found 9 outliers among 100 measurements (9.00%) 
1 (1.00%) low mild 
4 (4.00%) high mild 
4 (4.00%) high severe
```

### DIFF
--- a/.github/workflows/rust_lint.yml
+++ b/.github/workflows/rust_lint.yml
@@ -27,7 +27,7 @@ jobs:
       SCCACHE_ERROR_LOG: /tmp/sccache_log.txt
       SCCACHE_LOG: info
     steps:
-      - run: sudo apt-get update
+      - run: sudo apt-get update && sudo apt-get install -y protobuf-compiler
       - uses: actions/checkout@v4
       - run: rustup update
       - name: Install sccache
@@ -52,7 +52,7 @@ jobs:
       SCCACHE_ERROR_LOG: /tmp/sccache_log.txt
       SCCACHE_LOG: info
     steps:
-      - run: sudo apt-get update
+      - run: sudo apt-get update && sudo apt-get install -y protobuf-compiler
       - uses: actions/checkout@v4
       - run: rustup update
 
--- a/.github/workflows/rust_test.yml
+++ b/.github/workflows/rust_test.yml
@@ -61,7 +61,7 @@ jobs:
       SCCACHE_LOG: info
 
     steps:
-      - run: sudo apt-get update
+      - run: sudo apt-get update && sudo apt-get install -y protobuf-compiler
       - uses: actions/checkout@v4
         with:
           submodules: recursive
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -130,6 +130,17 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "320119579fcad9c21884f5c4861d16174d0e06250625266f50fe6898340abefa"
 
 [[package]]
+name = "aes"
+version = "0.8.4"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "b169f7a6d4742236a0a00c541b845991d0ac43e546831af1249753ab4c3aa3a0"
+dependencies = [
+ "cfg-if",
+ "cipher",
+ "cpufeatures",
+]
+
+[[package]]
 name = "ahash"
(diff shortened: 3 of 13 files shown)


dhruvkjain:
yeah glam has reduced 3-4µs
```shell
covariance_faer (unoptimized)
time:   [1.3234 ms 1.3270 ms 1.3312 ms]
Found 2 outliers among 100 measurements (2.00%)
  2 (2.00%) high severe

covariance_glam (vectorized)
time:   [35.594 µs 35.739 µs 35.909 µs]
Found 8 outliers among 100 measurements (8.00%)
  4 (4.00%) low mild
  2 (2.00%) high mild
  2 (2.00%) high severe

covariance_plain (scalar loop)
time:   [38.865 µs 39.057 µs 39.277 µs]
```

crates/kornia-icp/src/ops.rs
```diff
@@ -14,13 +15,27 @@ pub(crate) fn fit_transformation(
     let (src_centroid, dst_centroid) = compute_centroids(points_in_src, points_in_dst);
```
edgarriba, line 15: proabably here we want also to use glam

crates/kornia-icp/src/ops.rs
```diff
@@ -14,13 +15,27 @@ pub(crate) fn fit_transformation(
...
-        let p_src = faer::col![p_in_src[0], p_in_src[1], p_in_src[2]] - &src_centroid;
-        let p_dst = faer::col![p_in_dst[0], p_in_dst[1], p_in_dst[2]] - &dst_centroid;
-        hh += p_src * p_dst.transpose();
+        let ps = DVec3::from(*p_in_src) - src_c;
+        let pd = DVec3::from(*p_in_dst) - dst_c;
+
+        hh += DMat3::from_cols(ps * pd.x, ps * pd.y, ps * pd.z);
     }
 
+    let hh = faer::Mat::<f64>::from_fn(3, 3, |i, j| hh.col(j)[i]);
```
edgarriba, line 37: maybe we want to create a `MatRef` instead to not copy data

sidd-27:
I would also recommend replacing the svd with the SVD3 we have, but alas its just for F32 for now whereas this icp pipeline is in f64, so this will have to wait a bit

crates/kornia-icp/src/ops.rs
dhruvkjain: so I tried converting `compute_centroids` to use glam and used glam matrices in covariance calculation also 

results:
```txt
using faer for both:
covariance_faer_full    time:   [4.2304 ms 4.2335 ms 4.2369 ms]

using glam matrices in covariance calculation:
covariance_glam_matrix    time:   [71.518 µs 72.282 µs 73.660 µs]

using glam for both
covariance_glam_accumulate_glam_centroid    time:   [45.576 µs 45.840 µs 46.240 µs]

using faer and scalar accumulation
covariance_scalar_accumulate    time:   [71.624 µs 71.647 µs 71.683 µs]
```

so basically best is to use glam but this part results in an faer `rr` matrix as svd is of faer crate
```rust
// solve the linear system H * x = 0 to find the rotation
    let svd = hh.svd();
    let (u_t, v) = (svd.u().transpose(), svd.v());

    // compute rotation matrix R = V * U^T
    let mut rr = v * u_t;
```
so we would require to type cast again here.

I saw last few messages on discord kornia-rs channel and it seems we are going to do some unification regarding matrix algebra.
so what should I do now ??

crates/kornia-icp/src/ops.rs
```diff
@@ -1,3 +1,4 @@
+use glam::{DMat3, DVec3};
```
edgarriba, line 1: Use the new merged kornia_algebra crate types

edgarriba (changes requested):

edgarriba:

PR #520, opened 2025-10-20
Add support for Python 3.14

### LINKED ISSUE #508: Support Python 3.14
I think this needs to wait on PyTorch releasing support for Python 3.14.

When that's done, it should be possible to substantially simplify the testing configuration for free-threaded Python. uv [no longer considers](https://github.com/astral-sh/uv/pull/16142) 3.14t to be experimental, so the workarounds in the 3.13t tests can be dropped for 3.14t. I don't think it's particularly necessary to explicitly test both 3.13t and 3.14t if 3.13t is onerous to support. Dropping support for 3.13t in future versions of kornia-rs is also entirely reasonable.

ngoldbaum:
Fixes #508.

It looks like 3.14 is already set up in the release automation, so I guess this just unbreaks the 3.14 release builds? I fired off a build on my fork: https://github.com/ngoldbaum/kornia-rs/actions/runs/18658536851.

### DIFF
--- a/kornia-py/Cargo.toml
+++ b/kornia-py/Cargo.toml
@@ -23,5 +23,5 @@ kornia-icp = { path = "../crates/kornia-icp" }
 kornia-3d = { path = "../crates/kornia-3d" }
 
 # external
-pyo3 = { version = "0.24.0", features = ["extension-module"] }
-numpy = { version = "0.24.0" }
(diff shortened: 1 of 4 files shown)


ngoldbaum:
Maybe also time to drop Python 3.8? It's EOL as of last year.

edgarriba (approved):
LGTM. feel free to open a new one dropping support to 3.8

PR #521, opened 2025-10-23
optimized covariance calculations

sidd-27:
# perf: Optimize covariance matrix computation by exploiting symmetry

This PR optimizes the covariance matrix calculation by leveraging its inherent symmetric property. Instead of calculating the full 9-component outer product for each point, it now only computes the 6 unique elements, avoiding redundant work and yielding a **~35% performance improvement**.

---

### The Problem

The previous implementation calculated the full outer product `diff * diffᵀ` for each point. A 3x3 covariance matrix is symmetric, meaning the element at `(i, j)` is identical to the element at `(j, i)`. The original approach was therefore performing redundant work by implicitly calculating, for example, both `diff.x * diff.y` and `diff.y * diff.x` when these values are identical.

---

### The Solution

The new implementation takes advantage of this symmetry to reduce the computational load.

* **Before**: Computed all 9 components of the matrix for every point.
* **After**: Accumulates only the 6 unique components (the diagonal and the upper triangle) in scalar variables inside the hot loop. The full, symmetric matrix is constructed just once at the end.

This change reduces the number of multiplications inside the loop from 9 to 6 per point—a **33% reduction in floating-point operations**.

---

### Benchmark Results

The benchmark results confirm the effectiveness of this approach. The reduction in computational work leads directly to a significant runtime improvement.

| Metric | Original | Optimized | Change |
| :--- | :---: | :---: | :--- |
| **Mean Time** | 105.78 µs | 70.584 µs | **↓ 33.27% Faster** |
| **Speedup** | — | **1.50×** | — |

The **~35% speedup** is a direct result of eliminating redundant calculations. This is a more efficient and direct method for computing a symmetric matrix.

---

### Verification

You can use the following `criterion` benchmark script to validate the numerical equivalence of the two functions and reproduce the performance results.

```rust
use glam::{Mat3, Vec3};
use rand::Rng;
use std::hint::black_box;
use criterion::{criterion_group, criterion_main, Criterion};

fn covariance_original(points: &[Vec3]) -> Mat3 {
    let n = points.len() as f32;
    let c = points.iter().sum::<Vec3>() / n;

    let mut cov_mat = Mat3::ZERO;
    for &p in points {
        let diff = p - c;
        // Outer product diff * diffᵀ via column scaling
        let outer = Mat3::from_cols(diff * diff.x, diff * diff.y, diff * diff.z);
        cov_mat += outer;
    }
    cov_mat * (1.0 / n)
}

fn covariance_optimized(points: &[Vec3]) -> Mat3 {
    let n = points.len() as f32;
    let c = points.iter().sum::<Vec3>() / n;

    let mut c00 = 0.0;
    let mut c01 = 0.0;
    let mut c02 = 0.0;
    let mut c11 = 0.0;
    let mut c12 = 0.0;
    let mut c22 = 0.0;

    for &p in points {
        let dx = p.x - c.x;
        let dy = p.y - c.y;
        let dz = p.z - c.z;

        c00 += dx * dx;
        c01 += dx * dy;
        c02 += dx * dz;
        c11 += dy * dy;
        c12 += dy * dz;
        c22 += dz * dz;
    }

    let s = 1.0 / n;
    Mat3::from_cols_array(&[
        c00 * s, c01 * s, c02 * s,
        c01 * s, c11 * s, c12 * s,
        c02 * s, c12 * s, c22 * s,
    ])
}

fn random_points(n: usize) -> Vec<Vec3> {
    let mut rng = rand::thread_rng();
    (0..n)
        .map(|_| Vec3::new(
            rng.gen_range(-10.0..10.0),
            rng.gen_range(-10.0..10.0),
            rng.gen_range(-10.0..10.0),
        ))
        .collect()
}

fn validate_equivalence() {
    let points = random_points(10_000);
    let cov1 = covariance_original(&points);
    let cov2 = covariance_optimized(&points);

    let diff = cov1 - cov2;
    let max_err = diff.to_cols_array()
        .iter()
        .map(|x| x.abs())
        .fold(0.0, f32::max);

    println!("Max absolute difference: {:.8e}", max_err);
    assert!(max_err < 1e-5, "Covariance mismatch!");
}

fn benchmark_covariance(c: &mut Criterion) {
    let points = random_points(20000);
    validate_equivalence();

    c.bench_function("covariance_original", |b| {
        b.iter(|| {
            // Prevent optimizer from removing the computation
            black_box(covariance_original(black_box(&points)))
        })
    });

    c.bench_function("covariance_optimized", |b| {
        b.iter(|| {
            black_box(covariance_optimized(black_box(&points)))
        })
    });
}

criterion_group!(benches, benchmark_covariance);
criterion_main!(benches);

### DIFF
--- a/crates/kornia-pnp/src/epnp.rs
+++ b/crates/kornia-pnp/src/epnp.rs
@@ -290,9 +290,15 @@ fn select_control_points(points_world: &[[f32; 3]]) -> [[f32; 3]; 4] {
 
     let s = 1.0 / n;
     let cov_mat = Mat3::from_cols_array(&[
-        c00 * s, c01 * s, c02 * s,
-        c01 * s, c11 * s, c12 * s,
-        c02 * s, c12 * s, c22 * s,
+        c00 * s,
+        c01 * s,
+        c02 * s,
+        c01 * s,
+        c11 * s,
+        c12 * s,
+        c02 * s,
+        c12 * s,
+        c22 * s,
     ]);
 
     let svd = svd3(&cov_mat);

### CODE REVIEW

File: crates/kornia-pnp/src/epnp.rs [hunk 1]
```diff
@@ -265,18 +265,41 @@ fn rmse_px(
...
-        cov_mat += outer_product;
+    let mut c00 = 0.0;
+    let mut c01 = 0.0;
+    let mut c02 = 0.0;
+    let mut c11 = 0.0;
+    let mut c12 = 0.0;
+    let mut c22 = 0.0;
+
+    for &p in points_world {
+        let dx = p[0] - c[0];
```
edgarriba, line 279: why not using glam vecotrized syntax. `let diff = Vec3:from_Array(p) - c` and so on. I'd prioritize anything to go with glam function as it abstract many optimisations across architectures

crates/kornia-pnp/src/epnp.rs
```diff
@@ -265,18 +265,41 @@ fn rmse_px(
...
     let c = compute_centroid(points_world);
 
-    // Compute covariance using glam for consistency
-    let mut cov_mat = Mat3::ZERO;
-    for p in points_world {
-        let diff = Vec3::new(p[0] - c[0], p[1] - c[1], p[2] - c[2]);
-        // Outer product diff * diffᵀ via column scaling
-        let outer_product = Mat3::from_cols(diff * diff.x, diff * diff.y, diff * diff.z);
-        cov_mat += outer_product;
+    let mut c00 = 0.0;
```
edgarriba, line 271: Here we could initialise  mutable zeros mat already

edgarriba:

Issue #512, opened 2025-10-13
Potential issue with transform_point


sidd-27:
```
fn transform_point(x: f32, y: f32, m: &[f32; 9]) -> (f32, f32) {
    let w = m[6] * x + m[7] * y + m[8];
    let x = (m[0] * x + m[1] * y + m[2]) / w;
    let y = (m[3] * x + m[4] * y + m[5]) / w;
    (x, y)
}

```

Here, isn't the variable x being reassigned and modified?

PabloDArandaR:
As far as I am concerned, the function won't modify the inputs outside of the function. Using the following example:
```
{
    mut x_input: f32 = 10.0;
    mut y_input:: f32 = 10.0;
    mut m_input:: [f32; 9] = [1.0; 9];

   transform_point(x_input ,y_input ,m_input:);
}
```
The original x_input, y_input will not be modified.

If you refer to whether the input x,y are being shadowed, then yeah, I assume they are being shadowed and therefore the calculation is not correct.

PabloDArandaR:
Created the PR https://github.com/kornia/kornia-rs/pull/522 to solve it

PR #522, opened 2025-10-24
Change variable names of transform_point to avoid shadowing

PabloDArandaR:
Solving issue #512.

transform_point is shadowing the input x for the y calculation.


PabloDArandaR:
@edgarriba okay, but before the change the variable x was indeed being overwritten in the scope of this function, no?

edgarriba:
in the scope yes

PR #535, opened 2025-11-06
bug fix in so3 exp

sidd-27:
## **Bug Report: `SO3::exp` (Quaternion) Unstable and Logically Flawed for Small Rotations**

**Title:** `SO3::exp` is numerically unstable for small non-zero rotations and has an incorrect mathematical limit for zero rotations.

  * This bug causes `NaN` (Not a Number) outputs for valid, small rotation inputs.
  * The logic for the $\theta = 0$ case is mathematically incorrect, though this is currently masked by a second bug.

-----

### **Summary**

The `SO3::exp` function, which converts a 3D rotation vector (Lie algebra) into a quaternion SO(3), fails when the input vector `v` has a small magnitude. This is a common and critical code path, as it's responsible for handling near-identity rotations.

The function has two distinct problems:

1.  **Numerical Instability:** The main `if` block calculates `sin(theta_half) / theta`, which becomes a $0/0$ division for small `theta`, leading to `NaN`.
2.  **Mathematical Error:** The `else` block (for `theta = 0.0`) incorrectly sets the coefficient `b` to `0.0`. The correct mathematical limit is `0.5`.

### **Problematic Code**

The flawed logic is in this `if/else` block:

```rust
        let (w, b) = if theta != 0.0 {
            (
                theta_half.cos(),
                theta_half.sin() / theta // [!] BUG 1: Becomes 0/0 for small theta
            )
        } else {
            (1.0, 0.0) // [!] BUG 2: Mathematically incorrect limit
        };
```

-----

#### 1\. The Numerical Instability (The `if` block)

When `theta` is a very small number (e.g., `1e-10`):

  * `theta_half` is `0.5e-10`.
  * `theta_half.sin()` is also a very small number, approximately `0.5e-10`.
  * The calculation becomes `(small number) / (small number)`, which is a textbook case of catastrophic cancellation and leads to `NaN` or massive precision loss.
  * This is a classic $0/0$ indeterminate form that must be handled with a Taylor series.

#### 2\. The Mathematical Error (The `else` block)

This block is intended to handle the $\theta = 0$ case, so it must be the **mathematical limit** of the `if` block's coefficients as $\theta \to 0$.

  * **Limit for `w`:** $\lim_{\theta \to 0} \cos(\theta/2) = \cos(0) = \mathbf{1.0}$. This part is correct.
  * **Limit for `b`:** $\lim_{\theta \to 0} \frac{\sin(\theta/2)}{\theta}$.
      * This is another $0/0$ form. We can use L'Hôpital's rule (differentiate top and bottom with respect to $\theta$):
      * $\lim_{\theta \to 0} \frac{\cos(\theta/2) \cdot (1/2)}{1} = \frac{\cos(0) \cdot (1/2)}{1} = \mathbf{0.5}$.

The correct limit for `b` is **`0.5`**, not `0.0`.

This bug is currently *masked*. The `else` block only runs when `theta = 0.0`, which implies `v` is `[0,0,0]`. The final calculation `xyz = b * v` becomes `0.0 * [0,0,0]`, which is `[0,0,0]`. If the buggy `b=0.0` was used with a correct `b=0.5`, the result would be `0.5 * [0,0,0]`, which is *also* `[0,0,0]`. It's a "two wrongs make a right" situation that hides the logical flaw.

-----

### **The Fix**

The solution is to replace the `if theta != 0.0` check with a small-angle epsilon and use a **Taylor series approximation** for the small-angle case. This solves both problems at once.

The Taylor expansions are:

  * $w = \cos(\theta/2) \approx 1 - \frac{(\theta/2)^2}{2!} = 1 - \frac{\theta^2}{8}$
  * $b = \frac{\sin(\theta/2)}{\theta} \approx \frac{(\theta/2) - (\theta/2)^3/3!}{\theta} = \frac{1}{2} - \frac{\theta^2}{48}$

Here is the corrected, robust implementation:

```rust
const SMALL_ANGLE_EPSILON: f32 = 1.0e-8;

pub fn exp(v: Vec3A) -> Self {
    let theta_sq = v.dot(v);
    let theta = theta_sq.sqrt();
    let theta_half = 0.5 * theta;

    let (w, b) = if theta < SMALL_ANGLE_EPSILON {
        // --- FIX: Use Taylor series for small angles ---
        // This correctly handles theta = 0.0, giving w=1.0, b=0.5.
        // It also handles small non-zero theta without 0/0 division.
        (1.0 - theta_sq / 8.0, 0.5 - theta_sq / 48.0)
    } else {
        // Large-angle case (your original, correct math)
        (theta_half.cos(), theta_half.sin() / theta)
    };

    let xyz = b * v;

    Self {
        q: Quat::from_xyzw(xyz.x, xyz.y, xyz.z, w),
    }
}
```

### DIFF
--- a/crates/kornia-lie/src/so3.rs
+++ b/crates/kornia-lie/src/so3.rs
@@ -88,7 +88,7 @@ impl SO3 {
         let theta_half = 0.5 * theta;
 
         let (w, b) = if theta < SMALL_ANGLE_EPSILON {
-            //using the taylor series expansion of cos(x/2) and sin(x/2)/x around 0
+            // using the taylor series expansion of cos(x/2) and sin(x/2)/x around 0
             (1.0 - theta_sq / 8.0, 0.5 - theta_sq / 48.0)
         } else {
             (theta_half.cos(), theta_half.sin() / theta)


edgarriba (approved):
LGTM. Can you add as a comment where the math comes from?

crates/kornia-lie/src/so3.rs
```diff
@@ -81,15 +82,16 @@ impl SO3 {
...
     pub fn exp(v: Vec3A) -> Self {
-        let theta = v.dot(v).sqrt();
+        let theta_sq = v.dot(v);
+        let theta = theta_sq.sqrt();
         let theta_half = 0.5 * theta;
 
-        let (w, b) = if theta != 0.0 {
-            (theta_half.cos(), theta_half.sin() / theta)
+        let (w, b) = if theta < SMALL_ANGLE_EPSILON {
+            //using the taylor series expansion of cos(x/2) and sin(x/2)/x around 0
```
edgarriba:
```suggestion
            // using the taylor series expansion of cos(x/2) and sin(x/2)/x around 0
```

Issue #525, opened 2025-10-31
potential epnp optimizations


sidd-27:
here are some possible micro-optimizations I've spotted in the EPNP implementation. If anyone is interested in coding the changes, please feel free to try it out

1) Use Smatrix instead of DMatrix - considering the majority of the matrices used in the algorithm have fixed sizes, it's definitely faster to use stack-allocated matrices instead of dynamic heap-allocated ones.

2) not storing the first value in the barycentric co-ordinates - since the first element is simply 1 - others, it can be recalculated in the build_m function instead of the loop in the compute_barycentric function. This change also reduces the space required by the algorithm, which may be very useful in edge devices. From my rough benchmarks, this is also faster, due to more efficient cache utilization; however, more thorough benchmarking is needed. Taking this one step further, we could simply not store the barycentric coordinates at all and compute them directly in the build_m loop, which saves a lot of space and may even be faster; however, benchmarking is needed.

3) Using more efficient solvers in the estimate_beta function - considering the inputs are only about 6x5 size matrices, a full SVD decomposition may be overkill; something like QR decomposition could work, however, testing is required to make sure that it's just as accurate.

4) not computing the full mtm matrix - since we know that it'll always be symmetric, and also that the symmetric eigenfunction only looks at the lower diagonal half, we could theoretically cut the number of computations in half, but since the GEMM algorithm in Nalgebra is probably very optimized, benchmarking is required to make sure that it's worth making this change is actually faster

these are the 4 possible optimizations I've spotted, other than the pull requests I've already raised. If anyone wants to try to implement these, make sure to do only one optimization in one PR and also provide the necessary benchmarking and validation.

Additionally, if there are any optimizations that I've missed, please feel free to point them out and discuss them further here.

sidd-27:
also another thing to look into is the use of glam vec3 instead of points, theres already a bit of that in the unemeya algorithm but thats just switching to glam for one function and back, which i feel like adds more overhead. we should make glam usage more consistent

edgarriba:
i'd be in favaour that in kornia-linalg or similar we re-export Vec3, Mat3, etc so that can be consistently used across the other crates @cjpurackal

cjpurackal:
> i'd be in favaour that in kornia-linalg or similar we re-export Vec3, Mat3, etc so that can be consistently used across the other crates [@cjpurackal](https://github.com/cjpurackal)

[...]

PR #542, opened 2025-11-08
Faer safe matrices

thimotedupuch:
Fixed a typo in `src/pose/affine.rs`.

Changed the constructions of matrices for the optimization problems from using `faer::Mat::zeros` and then filling them with unsafe `write_unchecked` calls to `faer::Mat::from_fn`. All the changes should pass the tests.

`kornia-3d` should now contain only unsafe code in tests.

I also made the unrelated change to `kornia-imgproc/src/pyramid.rs` on a second commit where the function `get_pyramid_gaussian_kernel` was almost useless. For now I don't have the time to investigate a more optimized computation of the gaussian kernel.

---
Thank you very much,
Thimoté D.


sidd-27:
I would recommend dropping the faer dependency wherever we can, considering the compilation issues. Also, Faer only has heap-allocated matrices, whereas nalgebra's [Smatrix](https://docs.rs/nalgebra/latest/nalgebra/base/type.SMatrix.html) is stack-allocated, and since here the size of the matrix is fixed, this + QR decomposition to solve it may be faster too

thimotedupuch:
Sure. I wasn't aware of the compilation issues with faer. Do you know where else it is used in kornia-rs ? Since we are working with small matrices, `nalgebra` is probably much better but that would mean a change of dependency.

sidd-27:
i benchmarked it, and even tho Nalgebra doesn't have a specialised solver for least squares rn, even with a full SVD it is still about 20% faster than Faer , and just as accurate, and since in most of out code we're dealing with small matrices using Nalgebra might be faster and easier to maintain @edgarriba

edgarriba:
for the solvers, then let's move to nalgebra. However, for interfacing or vec/mat operations i'd like to keep glam-rs as its a very simple and light weighted crate. @sidd-27 if we can capture in a ticket this ?

sidd-27:
yupp, as soon as we have kornia-svd support f64, it should be pretty easy to remove all dependencies and switch to glam for the 3x3 and 3x1 vectors and nalgebra for the bigger ones

cjpurackal:
I think, for consistency let's keep all the types in `kornia-linalg` (as proposed by @edgarriba  here: https://github.com/kornia/kornia-rs/issues/525#issuecomment-3496137637). Let's include this in the ticket too for further discussion. @sidd-27

PR #556, opened 2025-11-11
feat(linalg): replace nalgebra SVD in umeyama with internal svd3

### LINKED ISSUE #452: Make `kornia-linalg::umeyama` use `kornia-linalg::svd3` instead of `naglebra`'s SVD.
One of the current limitations of `kornia-linalg::svd` is that it is not accurate enough for `umeyama`, and therefore cannot be used in the `kornia-pnp` crate for eg. The discrepancy between the results from `nalgebra’`s SVD and `kornia-linalg::svd3` is likely due to a trade-off of speed over accuracy.

It might be useful to have a mode that prioritizes higher accuracy for certain applications, even at the cost of performance. This could potentially be achieved by tuning parameters in `kornia-linalg::svd3`, such as `JACOBI_STEPS`.

An additional benefit of this improvement would be the ability to drop the `nalgebra` dependency from the kornia-linalg crate.

AnilKumarSingh9856:
Closes #452

This PR follows up on the conversation in issue #452 and the recent merge of the new `svd3` function.

The `umeyama` function in `kornia-linalg/src/rigid.rs` has been updated to remove the `nalgebra` dependency. It now uses the internal `kornia_linalg::svd3` function for the SVD calculation.

All tests for `kornia-linalg` pass, confirming the implementation is correct.

### DIFF
--- a/crates/kornia-linalg/benches/bench_linalg.rs
+++ b/crates/kornia-linalg/benches/bench_linalg.rs
@@ -1,21 +1,28 @@
 use criterion::{criterion_group, criterion_main, BenchmarkId, Criterion};
 use faer::mat;
 use glam::{Mat3, Vec3};
+use kornia_algebra::Mat3F32;
 use kornia_linalg::svd;
 
 fn bench_svd3(c: &mut Criterion) {
     let mut group = c.benchmark_group("svd3");
-    let a1 = Mat3 {
+
+    // 1. Setup Data (glam type)
+    let a1_glam = Mat3 {
         x_axis: Vec3::new(1.0, 0.0, 0.0),
         y_axis: Vec3::new(0.0, 2.0, 0.0),
         z_axis: Vec3::new(0.0, 0.0, 3.0),
     };
 
+    // 2. Convert to kornia type for svd3 (do this once, outside the loop)
+    let a1_kornia = Mat3F32::from(a1_glam);
+
     let a2 = mat![[1.0, 0.0, 0.0], [0.0, 2.0, 0.0], [0.0, 0.0, 3.0]];
 
     group.bench_function(BenchmarkId::new("svd3", ""), |b| {
         b.iter(|| {
-            svd::svd3(&a1);
+            // Pass the correct kornia wrapper type
+            svd::svd3(&a1_kornia);
             std::hint::black_box(());
         })
     });


sidd-27:
LGTM, could you add a few more tests for the unemeya function?

AnilKumarSingh9856:
@sidd-27 
Thanks for the LGTM! I've added the two new tests, as you suggested.

sidd-27:
There still seem to be numerical precision issues with the function. I'm not sure why the SVD is pretty accurate now, @cjpurackal and @AS1100K. Any ideas why?

cjpurackal:
The epnp tests are failing because `kornia::linalg::svd3` isn't accurate enough. It would be nice to see how accurate it is compared to nalebra's svd.

crates/kornia-linalg/src/rigid.rs
cjpurackal: Let's use `Vec3` here

cjpurackal: Use `Mat3` here and we can get rid of subsequent loops

crates/kornia-linalg/src/rigid.rs
```diff
@@ -69,27 +62,32 @@ pub fn umeyama(src: &[Vec3], dst: &[Vec3]) -> UmeyamaResult {
...
-    let Some(v_t) = svd.v_t else {
-        return Err(UmeyamaError::SvdVT);
-    };
-
-    let mut r_na = u * v_t;
-    if r_na.determinant() < 0.0 {
-        r_na.column_mut(2).scale_mut(-1.0);
+    // Convert H from [[f32; 3]; 3] (row-major) to glam::Mat3 (column-major)
+    // This is the corrected block
+    let h_glam = Mat3::from_cols(
```
cjpurackal: I think we can get rid of this too, once h is `Mat3`

cjpurackal:
It would also be nice if we can get some tests for more complex transformations rather than just identity ones.

AnilKumarSingh9856:
Hi @cjpurackal, thanks for the refactoring tips! I've updated the PR to use Vec3 and Mat3 directly, which really cleaned up the covariance calculation. I also added a new test with a 45-degree rotation as you suggested. The kornia-linalg tests are all passing locally. It looks like the core svd3 accuracy issue is the remaining blocker for the epnp tests. I'm happy to help in any other way I can!

crates/kornia-linalg/src/rigid.rs
```diff
@@ -33,77 +26,59 @@ pub fn umeyama(src: &[Vec3], dst: &[Vec3]) -> UmeyamaResult {
     }
     let n = src.len() as f32;
 
-    // Centroids
-    let mut mu_s = [0.0; 3];
-    let mut mu_d = [0.0; 3];
+    // --- Centroids (using Vec3 as requested) ---
```
edgarriba: no agentic comments, follow rust good practices

crates/kornia-linalg/src/rigid.rs
```diff
@@ -33,77 +26,59 @@ pub fn umeyama(src: &[Vec3], dst: &[Vec3]) -> UmeyamaResult {
...
-        let dc = [dst[i].x - mu_d[0], dst[i].y - mu_d[1], dst[i].z - mu_d[2]];
-        for (r, &dc_r) in dc.iter().enumerate() {
-            for (c, &sc_c) in sc.iter().enumerate() {
-                h[r][c] += dc_r * sc_c;
-            }
-        }
+        let sc = src[i] - mu_s;
+        let dc = dst[i] - mu_d;
+        // H = sum(dc * sc.transpose())
+        h += Mat3::from_cols(dc * sc.x, dc * sc.y, dc * sc.z);
```
edgarriba: it cannot be done first as dc * sc instead component by component?

crates/kornia-linalg/src/rigid.rs
```diff
@@ -33,77 +26,59 @@ pub fn umeyama(src: &[Vec3], dst: &[Vec3]) -> UmeyamaResult {
...
-        let sc = [src[i].x - mu_s[0], src[i].y - mu_s[1], src[i].z - mu_s[2]];
-        let dc = [dst[i].x - mu_d[0], dst[i].y - mu_d[1], dst[i].z - mu_d[2]];
-        for (r, &dc_r) in dc.iter().enumerate() {
-            for (c, &sc_c) in sc.iter().enumerate() {
-                h[r][c] += dc_r * sc_c;
-            }
-        }
+        let sc = src[i] - mu_s;
+        let dc = dst[i] - mu_d;
+        // H = sum(dc * sc.transpose())
```
edgarriba: add meaningful comments

AnilKumarSingh9856:
@sidd-27 @cjpurackal @edgarriba 
Thanks for the refactoring tips! I've updated the PR to use Vec3 and Mat3 directly, which really cleaned up the covariance calculation.
I also corrected the determinant-flip logic and added the complex rotation test you suggested.
I've cleaned up the code to pass the linter. It looks like the core svd3 accuracy issue is the remaining blocker for the epnp tests.

AnilKumarSingh9856:
Thanks for all the great feedback! I've pushed an update that addresses all the review comments:

  1.   Cleaned up all the extra comments as @edgarriba suggested.
  2.   Refactored the determinant logic to the canonical det(U)*det(V) method.
  3.   Added a new test_umeyama_reflection() to cover that case.

The kornia-linalg tests are all passing locally. I'll wait for the team to investigate the core svd3 accuracy bug that's failing the CI.

cjpurackal:
@AnilKumarSingh9856 Do you want to take a look into the accuracy issue of svd3? I don't think we can merge this until that is solved. For context, the 3d-2d correspondence points in the failing test are derived from opencv's epnp method.

crates/kornia-linalg/src/rigid.rs
```diff
@@ -34,76 +27,60 @@ pub fn umeyama(src: &[Vec3], dst: &[Vec3]) -> UmeyamaResult {
...
-
-    let mut r_na = u * v_t;
-    if r_na.determinant() < 0.0 {
-        r_na.column_mut(2).scale_mut(-1.0);
+    // Call the internal svd3 function
+    let svd_result = svd3(&h);
+
+    // Clone to get owned Mat3 regardless of whether svd3 returned &Mat3 or Mat3.
+    // If svd3 returns a Result or other wrapper you will need to adapt this.
+    let mut u_mat = svd_result.u().clone();
```
edgarriba: why clone ? same bleow

crates/kornia-linalg/src/rigid.rs
```diff
@@ -34,76 +27,60 @@ pub fn umeyama(src: &[Vec3], dst: &[Vec3]) -> UmeyamaResult {
...
+    let v_mat = svd_result.v().clone();
+
+    // Canonical Umeyama sign correction: if det(U)*det(V) < 0, flip the last column of U.
+    if (u_mat.determinant() * v_mat.determinant()) < 0.0 {
```
edgarriba: probably it needs a small epsilon instead of absolute zero

AnilKumarSingh9856:
Hi @edgarriba and @cjpurackal, thanks for the new review.
I've pushed a new commit that fixes all the style comments:
  1. Replaced .clone() with dereferencing (*svd_result.u()).
  2. Added an epsilon to the determinant check for robustness.

Regarding the main svd3 accuracy bug that's failing the epnp tests: I would like to try to look into it, but I'll admit that deep numerical benchmarking is new to me and it might be beyond my current skill set.
Do you have any pointers on how I could get started, or a specific test case I should try to reproduce? I'm happy to help in any way I can, even if it's just gathering data for the core team to analyze.

AnilKumarSingh9856:
Hi @edgarriba and @cjpurackal I took your suggestion and investigated the svd3 accuracy issue.

I've confirmed that the logic in my umeyama3d function is a correct 1:1 port of the old nalgebra version, so the bug is definitely inside the new svd3 implementation.
To debug it, I isolated the failing epnp test. I found that the svd3 algorithm is converging, but to the wrong value.
I tried increasing the precision of the jacobi_eigenanalysis by:
  1. Increasing MAX_SWEEPS from 4 to 10.
  2. Tightening the break tolerance from 1e-6 to 1e-10.

Neither change had any effect on the final result. The epnp test fails with the exact same incorrect value (0.65032107).
This suggests it's not a simple precision or convergence issue, but a deeper logic error in the svd3 math (maybe in qr_decomposition or sort_singular_values) that causes it to produce an incorrect U or V matrix for this specific case.

This bug seems more complex than I can solve right now. I'll focus on my other PR (fix-brightness-v2) for now, but I wanted to share my findings.

AnilKumarSingh9856:
Hi @cjpurackal and @sidd-27,
Good news: PR #580 (SVD f64 support) is now passing all CI checks (including aarch64 and benchmarks).
I have also verified locally that using the new svd3_f64 implementation from #580 inside this umeyama function completely resolves the accuracy issues. The synthetic tests pass perfectly with high precision.

Once #580 is merged, I will update this PR to use the new API.

crates/kornia-linalg/src/rigid.rs
```diff
@@ -1,8 +1,9 @@
 //! Rigid alignment utilities (Kabsch / Umeyama)
 
-// TODO: Make this work with kornia-linalg SVD(encountered some issues with precision)
-use glam::Vec3;
-use nalgebra::{Matrix3, SVD};
+// We use the f64 version of SVD for higher precision, then cast back to f32
+use crate::svd::svd3_f64;
+use glam::{DMat3, DVec3, Vec3};
```
edgarriba, line 4: Use the new merged kornia_algebra stuff

edgarriba:

PR #558, opened 2025-11-12
Implement kornia-algebra

### LINKED ISSUE #552: 3d operators structure unification
Based on different efforts and open discussion let's unify and reestructure all the 3d operators and types (drop basically faer) #525 #542

My proposal is the following:

- kornia-3d: functions for pose, homografies, ransac, low level solvers are here (sharing glma-rs interface)
- kornia-poincloud: spinoff from kornia-3d any pointcloud structs, io into a new crate
- kornia-linalg: fast and especialized svd, cholesky or whatever is needed
- kornia-lie: lie algebra and basic algebra types can live here (Vec3F32, VecF64, etc) [maybe we can rename this crate?]
- kornia-pnp: specific pnp solvers

cjpurackal:
fixes #552

### DIFF
--- a/crates/kornia-algebra/src/lib.rs
+++ b/crates/kornia-algebra/src/lib.rs
@@ -52,7 +52,7 @@
 //! // Create a rotation of 90 degrees around X axis
 //! let rotation = SO3F32::exp(Vec3AF32::new(std::f32::consts::FRAC_PI_2, 0.0, 0.0));
 //!
-//! let point = Vec3F32::new(0.0, 1.0, 0.0);
+//! let point = Vec3AF32::new(0.0, 1.0, 0.0);
 //! let rotated_point = rotation * point;
 //!
 //! // (0, 1, 0) rotated 90 deg around X becomes (0, 0, 1)


crates/kornia-algebra/src/lib.rs
```diff
@@ -0,0 +1,39 @@
+//! Lie groups and algebras for kornia-rs.
+//!
+//! This crate provides:
+//! - Unified algebraic types (`types` module) using newtype pattern over glam backend
+//! - Lie group implementations (SO2, SO3, SE2, SE3)
+
+// Algebraic types
+pub mod mat2;
```
edgarriba: if we re-export, i wouldn't make the submod public

crates/kornia-algebra/src/lib.rs
```diff
@@ -0,0 +1,39 @@
...
+pub use mat3a::Mat3A;
+pub use mat4::Mat4;
+pub use quat::Quat;
+pub use vec2::Vec2;
+pub use vec3::Vec3;
+pub use vec3a::Vec3A;
+pub use vec4::Vec4;
+
+// Re-export glam types that are used directly
+pub use glam::Affine3A;
```
edgarriba: can we have an `Isometry2F32` / `Isometry3F32` (and F64 respecteviley)

crates/kornia-algebra/src/lib.rs
```diff
@@ -0,0 +1,39 @@
...
+
+// Unified algebraic types module (re-exports)
+pub mod types;
+
+// Lie groups
+pub mod se2;
```
edgarriba: i'd keep this still under a `kalgebra::lie` mod

crates/kornia-algebra/src/matrix/mat2.rs
```diff
@@ -0,0 +1,78 @@
+//! 2x2 matrix (single precision).
+
+use crate::vec2::Vec2;
+use glam;
+use std::ops::{Deref, DerefMut};
+
+/// 2x2 matrix (single precision).
+#[derive(Debug, Clone, Copy, PartialEq)]
+#[repr(transparent)]
+pub struct Mat2(pub glam::Mat2);
```
edgarriba: I'd like that we follow a `Mat2F32` / `Mat2F64` convention
we can explore with macros to support modalities

crates/kornia-algebra/src/quat.rs
```diff
@@ -0,0 +1,71 @@
+//! Quaternion (single precision).
+
+use glam;
+
+/// Quaternion (single precision).
+#[derive(Debug, Clone, Copy, PartialEq)]
+#[repr(transparent)]
+pub struct Quat(pub glam::Quat);
```
edgarriba: we should support F32/F64 i guess here too

crates/kornia-algebra/src/lib.rs
edgarriba: i think we have to decide what pattern we want to follow

```rust
use kornia_algebra::prelude::*;

VS

use kornia_algebra::vec2::{Vec2F32, Vec2F64};

VS

use kornia_algebra::vec::{Vec2F32, Vec2F64, Vec3F32, Vec3F64};
```
@sidd-27 @AS1100K

crates/kornia-algebra/src/types.rs
edgarriba, line 1: types.rs is too generic -- I'd avoid that

edgarriba:
I'd say also, let's plan python/c++ for this

edgarriba:
@gandhigauri maybe you are interested in this

crates/kornia-algebra/src/lib.rs
sidd-27: the last one seems best to me, to both use and maintain

cjpurackal: I'd be in favour of having the types exposed at the crate root level to lessen the cognitive complexity of the api.
```
use kornia_algebra::{Vec2F32, Vec3F32, Vec3AF32};
// and for lie groups
use kornia_algebra::lie::{SO2, SE3};
```

edgarriba: Sounds but let's have the implementation more grouped as @sidd-27 suggested

cjpurackal:
@edgarriba can you do a quick review? Any thoughts on how we want to support multiple dtypes for the lie module?

crates/kornia-algebra/src/matrix/mat2.rs
```diff
@@ -0,0 +1,43 @@
...
+use crate::{Vec2F32, Vec2F64};
+
+// 2x2 matrix (single and double precision).
+define_matrix_type!(
+    Mat2F32,
+    glam::Mat2,
+    [f32; 4],
+    Vec2F32,
+    glam::Vec2,
+    [x_axis, y_axis]
```
edgarriba: why is this needed ?

crates/kornia-algebra/src/matrix/mat3a.rs
```diff
@@ -0,0 +1,92 @@
...
+    Mat3AF32,
+    glam::Mat3A,
+    [f32; 9],
+    Vec3AF32,
+    glam::Vec3A,
+    [x_axis, y_axis, z_axis]
+);
+
+// Scalar multiplication
+impl std::ops::Mul<f32> for Mat3AF32 {
```
edgarriba: why all this is not part of the macro?

crates/kornia-algebra/src/lib.rs
```diff
@@ -0,0 +1,34 @@
...
+
+// Type aliases for explicit precision (single precision / f32)
+pub type Vec4 = Vec4F32;
+
+// Isometry type aliases
+pub type Isometry2F32 = SE2;
+pub type Isometry3F32 = SE3;
+
+// Re-export glam types that are used directly
+pub use glam::Affine3A;
```
edgarriba: is not this the same as isometry in practice ? i'd keep only one

crates/kornia-algebra/src/quat.rs
```diff
@@ -0,0 +1,113 @@
...
+// Quaternion (single precision, `f32`).
+define_quat_type!(QuatF32, glam::Quat, f32);
+
+// Quaternion (double precision, `f64`).
+define_quat_type!(QuatF64, glam::DQuat, f64);
+
+impl QuatF32 {
+    /// Create a quaternion from a `Mat3AF32` matrix.
+    #[inline]
+    pub fn from_mat3a(mat: &crate::Mat3AF32) -> Self {
```
edgarriba: why not part of the macro?

crates/kornia-algebra/src/matrix/mat.rs
edgarriba, line 1: i'd keep all the class defined in the mat.rs (and same for vector) to have less file sparsity and easy to find/maintain. we don't have a lot

crates/kornia-algebra/src/matrix/mat3a.rs
```diff
@@ -0,0 +1,92 @@
+//! 3x3 matrix (aligned, single precision).
+
+use crate::Vec3AF32;
```
edgarriba: why don't just only support the "A" aka SIMD optimized version ?

edgarriba (commented):
to support lie for F32/F64 why we cannot do with macros ?

cjpurackal:
> to support lie for F32/F64 why we cannot do with macros ?

We can, just a little concerned with code readability.

edgarriba:
we can try later this

crates/kornia-algebra/src/lib.rs
cjpurackal: Affine will include shear and scaling, different to Isometry. The most general type would be affine.

crates/kornia-algebra/src/matrix/mat.rs
cjpurackal: done

crates/kornia-algebra/src/matrix/mat3a.rs
cjpurackal: moved

crates/kornia-algebra/src/matrix/mat2.rs
cjpurackal: these are function args for the `from_cols` method.

crates/kornia-algebra/src/matrix/mat3a.rs
cjpurackal: The lie types by default uses the aligned types and I will make it default for all the other dependencies as well. I think it's still good that we support non-aligned types just for non-simd cases perhaps?

crates/kornia-algebra/src/quat.rs
cjpurackal: now moved, although the macro is now a little ugly.

crates/kornia-algebra/src/matrix/mat3a.rs
edgarriba: would be great to claim that our lib is default simd accel

edgarriba:
@cjpurackal can you fix the copilot suggestions that makes more sense and we merge a first version of this to start integrating over the rest of the crates ?

cjpurackal:
@edgarriba it should be good now.

crates/kornia-algebra/src/mat.rs
```diff
@@ -0,0 +1,900 @@
...
+mod tests {
+    use super::*;
+
+    // Mat2 tests
+    #[test]
+    fn test_mat2f32_from_cols() {
+        let col1 = Vec2F32::new(1.0, 2.0);
+        let col2 = Vec2F32::new(3.0, 4.0);
+        let m = Mat2F32::from_cols(col1, col2);
+        assert_eq!(m.x_axis, glam::Vec2::new(1.0, 2.0));
```
edgarriba: Wondering if we should return Vec2F32 ?

crates/kornia-algebra/src/vec.rs
```diff
@@ -0,0 +1,385 @@
...
+                let v: $glam_type = self.into();
+                v.length()
+            }
+
+            /// Dot product between two vectors.
+            #[inline]
+            pub fn dot(self, rhs: Self) -> $scalar {
+                let a: $glam_type = self.into();
+                let b: $glam_type = rhs.into();
+                a.dot(b)
```
edgarriba, line 56: Can’t we do self.0.dot(rhs.0) ? Wondering if all the into will create an overhead

crates/kornia-algebra/README.md
```diff
@@ -0,0 +1,148 @@
+# kornia-algebra
```
edgarriba, line 1: This should go as docstring in the lib.rs and the blocks of code being compiled to not get deprecated

crates/kornia-algebra/README.md
```diff
@@ -0,0 +1,148 @@
...
+
+```toml
+[dependencies]
+kornia-algebra = "0.1"
+```
+
+### Basic Linear Algebra
+
+```rust
+use kornia_algebra::{Vec3, Mat3};
```
edgarriba: I advocate to teach users to use Vec2F32 flavour

crates/kornia-algebra/src/lib.rs
```diff
@@ -0,0 +1,39 @@
...
+mod vec;
+
+// Re-export types at crate root for convenience
+pub use lie::{SE2, SE3, SO2, SO3};
+pub use mat::{Mat2F32, Mat2F64, Mat3AF32, Mat3F32, Mat3F64, Mat4F32, Mat4F64};
+pub use quat::{QuatF32, QuatF64};
+pub use vec::{Vec2F32, Vec2F64, Vec3AF32, Vec3F32, Vec3F64, Vec4F32, Vec4F64};
+
+// Type aliases for default single-precision types.
+pub type Vec2 = Vec2F32;
```
edgarriba: After reading the whole thing this is confusing for users. I would stick to one single convention and avoid maintenance burden

edgarriba: Having defaults is also not good I we want to have strict typings

cjpurackal: okay, let's go for more strict typing.

PR #580, opened 2025-11-28
feat(linalg): add f64 support to svd3

### LINKED ISSUE #451: Add fp64 support to `kornia-linalg::svd3`
The current `svd` method in `kornia-linalg` only support fp32,  which is limiting for crates such as `konira-icp` and `kornia-pnp` that requires fp64 support.

AnilKumarSingh9856:
Closes #451

This PR adds `f64` (double precision) support to the SVD implementation, as requested in issue #451.

**Changes:**
- Added `f64` versions of all internal structures (`Symmetric3x3F64`, `GivensF64`, `QR3F64`).
- Implemented `svd3_f64` which accepts `glam::DMat3` and returns `SVD3SetF64` containing `DMat3` results.
- Added comprehensive tests for `f64` decomposition to ensure parity with the `f32` implementation.
- All tests pass via `cargo test -p kornia-linalg`.

This enables high-precision rigid alignment (Umeyama/Kabsch) in downstream tasks.


crates/kornia-linalg/src/svd.rs
```diff
@@ -1,35 +1,36 @@
...
+//  f32 Implementation
+// =================================================================================
+
 #[derive(Debug, Clone)]
 /// A simple symmetric 3x3 Matrix class (contains no storage for (0, 1) (0, 2) and (1, 2)
 struct Symmetric3x3 {
-    /// The element at row 0, column 0, first diagonal element.
     m_00: f32,
-
-    /// The element at row 1, column 0. equivalent to `m_01`.
```
siddharth-rambhia, line 15: why have all the comments been removed?

AnilKumarSingh9856:
Apologies for that oversight. I likely removed them accidentally during the refactoring process.
I have restored the documentation for Symmetric3x3 and also ensured the new Symmetric3x3F64 struct is fully documented to match.

cjpurackal:
Hey, now that https://github.com/kornia/kornia-rs/pull/558 is merged, this is probably the first place where we can use `kornia-algebra`. Could you look into adapting SVD to use `kornia-algebra` ?

Additionally, I''m also considering whether we should integrate `kornia-linalg` to `kornia-algebra` sometime later. @edgarriba

crates/kornia-linalg/src/svd.rs
```diff
@@ -436,22 +393,396 @@ pub fn svd3(a: &Mat3) -> SVD3Set {
...
     SVD3Set { u, s, v }
 }
 
+// =================================================================================
+//  f64 Implementation (Duplicated for Double Precision)
+// =================================================================================
+
+#[derive(Debug, Clone)]
+/// A simple symmetric 3x3 Matrix class (f64)
+struct Symmetric3x3F64 {
```
cjpurackal: I don't think we should be keeping two different implementation for the same functionality. Probably best to use macro_rules. What do you think ? @edgarriba

edgarriba: agree

AnilKumarSingh9856: Thanks for the feedback! I agree that macro_rules is the right way to handle this duplication. I will refactor the code to use a macro to generate both the f32 and f64 implementations.

Regarding kornia-algebra: To keep this PR focused on fixing the fp64 support, would you prefer I stick to glam types inside the macro for now? We could migrate SVD to kornia-algebra in a follow-up PR to avoid making this one too complex.

cjpurackal: The plan is to drop third-party linalg crates and use `kornia-algebra` instead. In this PR, replace glam with kornia-algebra’s vector, matrix, and quaternion types (with f32/f64 support).
Yes, let's think about merging kornia-linalg later, no need to address here.

AnilKumarSingh9856: I have updated the implementation to use kornia-algebra types (Mat3F32, Mat3F64, etc.) for the public API as requested.
However, regarding the internal implementation, I found that kornia-algebra is currently missing several helper methods and trait implementations that are necessary for the algorithm and tests.

Findings from verification:
Missing Methods: max_element (for matrices) and min_element (for vectors) do not exist in kornia-algebra. I verified this by checking src/mat.rs and src/vec.rs.
Trait Limitations: The Sub trait is currently only implemented for Value - Value (e.g., impl Sub<$name> for $name), which makes operations involving references (e.g., &Mat3 - Mat3) verbose or impossible without cloning/dereferencing.

Solution: To keep this PR focused on adding f64 support without blocking on a massive refactor of the algebra crate, I opted to:
Accept kornia-algebra types in the function signature (clean public API).
Convert to glam types internally to leverage its complete feature set for the SVD logic.
Return kornia-algebra types.

This fulfills the requirement of using the new types in the API while ensuring the math remains correct and testable today.

cjpurackal:
> @cjpurackal after seeing this, it feel it's too convoluted and over complicated to maintain. Should we just support f64 ?

I agree that it's getting ugly with macros. I'd still want an fp32 svd though because of performance reasons.

AnilKumarSingh9856:
@edgarriba, @cjpurackal  Understood. Since we need to maintain both f32 (for performance) and f64 (for precision), but want to avoid the complexity of macros, we effectively have two options:

**1. Generics:** I can rewrite svd3 to be generic over T (e.g., fn svd3<T: Float>(...)). This is the cleanest usage but requires kornia-algebra types to implement common traits.
**2. Explicit Duplication:** I can revert to having two separate, explicit implementations (in f32 and f64 modules) without macros. This involves code duplication but is very easy to read and debug.

Which approach do you prefer for this codebase?

edgarriba:
How does it look like algebra to have generic traits in this case ?

AnilKumarSingh9856:
> How does it look like algebra to have generic traits in this case ?

To make the SVD implementation truly generic, kornia-algebra would need to expose common traits (e.g., a Matrix trait) that abstract over the underlying scalar type (so both Mat3F32 and Mat3F64 implement it).

Currently, Mat3F32 and Mat3F64 are distinct structs without a shared interface for operations like .transpose() or column access. Implementing those traits would require a significant refactor of kornia-algebra first.

My Proposal: Since we want to avoid the complexity of macros but keep performance (f32) and precision (f64), I suggest I explicitly verify the implementation by creating two clean, separate modules in this file (mod f32_impl and mod f64_impl) without using macros.

This introduces some code duplication but makes the code readable, debuggable, and simple. We can refactor to generics later when kornia-algebra adds the necessary traits.

Shall I proceed with removing the macros and splitting the implementation?

edgarriba:
alright, then let's keep separated for now into separate file/modules. @cjpurackal could be interesting to explore having a couple a Vector and Matrix traits

AnilKumarSingh9856:
> alright, then let's keep separated for now into separate file/modules. @cjpurackal could be interesting to explore having a couple a Vector and Matrix traits

Agreed. I have completely removed the macros and split the implementation into explicit impl_f32 and impl_f64 modules for better readability and maintenance.

I also implemented the kornia-algebra wrapper types for the public API as requested by @cjpurackal.

AnilKumarSingh9856:
The build is failing on aarch64, but the error is in crates/kornia-algebra/src/lie/se3.rs (test: test_se3_lplus_lminus_consistency), which I haven't touched.

-     left  = -0.599999
-     right = -0.6

This appears to be a precision issue in the existing kornia-algebra tests on ARM runners.
Since this is unrelated to my f64 SVD changes, would you like me to add a fix for this epsilon in this PR to get the CI green, or should we ignore the failure?

AnilKumarSingh9856:
Hi @edgarriba, @cjpurackal , I have verified locally that this new svd3_f64 implementation solves the precision issues reported in the Umeyama PR (#556).

When I plugged this logic into umeyama, the synthetic tests (including rotation and reflection) passed perfectly. This f64 support is exactly what is needed to unblock that PR.

edgarriba:

PR #585, opened 2025-12-03
imgproc: optimise threshold::in_range by removing iterator overhead

VyasGuru:
optimised threshold_in_range by manually unrolling over channels instead of using iter to avoid overhead.

### DIFF
--- a/crates/kornia-imgproc/benches/bench_threshold_in_range.rs
+++ b/crates/kornia-imgproc/benches/bench_threshold_in_range.rs
@@ -1,0 +1,94 @@
+use criterion::{criterion_group, criterion_main, BenchmarkId, Criterion};
+use kornia_image::{Image, ImageSize};
+use kornia_imgproc::parallel;
+use kornia_tensor::CpuAllocator;
+
+//old version using iter
+#[inline]
+fn in_range_old(
+    src: &Image<u8, 3, CpuAllocator>,
+    dst: &mut Image<u8, 1, CpuAllocator>,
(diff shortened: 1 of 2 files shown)


sidd-27:
```
pub fn in_range<T, const C: usize, A1: ImageAllocator, A2: ImageAllocator>(
    src: &Image<T, C, A1>,
    dst: &mut Image<u8, 1, A2>,
    lower_bound: &[T; C],
    upper_bound: &[T; C],
) -> Result<(), ImageError>
where
    T: Copy + PartialOrd + Send + Sync,
{
    if src.size() != dst.size() {
        return Err(ImageError::InvalidImageSize(
            src.cols(), src.rows(), dst.cols(), dst.rows(),
        ));
    }

    parallel::par_iter_rows(src, dst, |src_pixel, dst_pixel| {
        unsafe {
            let mut ok = true;

            let sp = src_pixel.as_ptr();
            let lp = lower_bound.as_ptr();
            let up = upper_bound.as_ptr();

            for c in 0..C {
                let v = *sp.add(c);
                let lo = *lp.add(c);
                let hi = *up.add(c);
                ok = ok & (v >= lo) & (v <= hi);
            }

            *dst_pixel.get_unchecked_mut(0) = ok as u8 * 255;
        }
    });

    Ok(())
}
```

This code gives up to a 2x optimization, mainly due to changes to the traits and the unsafe pixel access. 

old - 888μs
new - 443μs

do try the changes made here too, and provide more comprehensive benchmarks

PR #587, opened 2025-12-04
Optimize resize_native by removing meshgrid allocation 

VyasGuru:
# Optimize `resize_native` by removing meshgrid allocation

## Summary
This PR optimizes `resize_native` by **eliminating the allocation of intermediate meshgrids**
(`map_x`, `map_y`) used for coordinate calculation. Instead, source coordinates are computed
**on the fly inside the resize loop**, significantly reducing memory overhead and improving
cache locality.

---

## What Changed
- **Removed meshgrid allocation**
  - The previous implementation precomputed `map_x` and `map_y` using
    `meshgrid_from_fn`

- **On-the-fly coordinate computation**
  - Source coordinates (`x`, `y`) are now derived incrementally using precomputed
    `step_x` / `step_y`
  - Row-level invariants (e.g. `y`, vertical weights) are hoisted outside the inner loop

- **Tightened hot loops**
  - Reduced per-pixel overhead
  - Interpolation weights are computed once per pixel and reused across channels
  - Direct pointer arithmetic replaces per-channel interpolation calls

---

## Performance Results
Benchmarked using Criterion against the previous meshgrid-based implementation:

| Output Size | Old Time | New Time | Speedup |
|------------|---------:|---------:|--------:|
| 256×256    | 75.96 µs | 31.82 µs | ~2.4× |
| 512×512    | 119.23 µs | 46.59 µs | ~2.6× |
| 1024×1024  | 280.11 µs | 102.86 µs | ~2.7× |

---

## Supported Interpolation Modes
- Nearest
- Bilinear
- Bicubic (unchanged, not yet implemented)
- Lanczos (unchanged, not yet implemented)

---

## Benchmark Code

<details>
<summary><strong>Click to expand Criterion benchmark</strong></summary>

```rust
use criterion::{criterion_group, criterion_main, BenchmarkId, Criterion};
use std::hint::black_box;

use kornia_image::{allocator::ImageAllocator, Image, ImageSize};
use kornia_tensor::CpuAllocator;

use kornia_imgproc::{
    interpolation::{grid::meshgrid_from_fn, interpolate_pixel, InterpolationMode},
    parallel,
    resize::resize_native,
};

/// Old implementation
pub fn resize_native_old<const C: usize, A1: ImageAllocator, A2: ImageAllocator>(
    src: &Image<f32, C, A1>,
    dst: &mut Image<f32, C, A2>,
    interpolation: InterpolationMode,
) -> Result<(), kornia_image::ImageError> {
    if src.size() == dst.size() {
        dst.as_slice_mut().copy_from_slice(src.as_slice());
        return Ok(());
    }

    let (dst_rows, dst_cols) = (dst.rows(), dst.cols());

    let step_x = if dst_cols > 1 {
        (src.cols() - 1) as f32 / (dst_cols - 1) as f32
    } else {
        0.0
    };

    let step_y = if dst_rows > 1 {
        (src.rows() - 1) as f32 / (dst_rows - 1) as f32
    } else {
        0.0
    };

    let (map_x, map_y) = meshgrid_from_fn(dst_cols, dst_rows, |x, y| {
        Ok((x as f32 * step_x, y as f32 * step_y))
    })?;

    parallel::par_iter_rows_resample(dst, &map_x, &map_y, |&x, &y, dst_pixel| {
        dst_pixel.iter_mut().enumerate().for_each(|(k, pixel)| {
            *pixel = interpolate_pixel(src, x, y, k, interpolation);
        });
    });

    Ok(())
}

fn bench_resize_native(c: &mut Criterion) {
    let mut group = c.benchmark_group("resize_native_compare");

    let sizes = [(256, 256), (512, 512), (1024, 1024)];
    let modes = [InterpolationMode::Nearest, InterpolationMode::Bilinear];

    for (w, h) in sizes {
        let src = Image::<f32, 3, _>::from_size_val(
            ImageSize {
                width: w,
                height: h,
            },
            0.5,
            CpuAllocator,
        )
        .unwrap();

        for &mode in &modes {
            let mut dst_old = Image::<f32, 3, _>::from_size_val(
                ImageSize {
                    width: w / 2,
                    height: h / 2,
                },
                0.0,
                CpuAllocator,
            )
            .unwrap();

            let mut dst_new = dst_old.clone();

            let label = format!("{w}x{h}_{mode:?}");

            group.bench_with_input(BenchmarkId::new("old", &label), &src, |b, src| {
                b.iter(|| {
                    resize_native_old(
                        black_box(src),
                        black_box(&mut dst_old),
                        black_box(mode),
                    )
                    .unwrap();
                })
            });

            group.bench_with_input(BenchmarkId::new("new", &label), &src, |b, src| {
                b.iter(|| {
                    resize_native(
                        black_box(src),
                        black_box(&mut dst_new),
                        black_box(mode),
                    )
                    .unwrap();
                })
            });
        }
    }

    group.finish();
}

criterion_group!(benches, bench_resize_native);
criterion_main!(benches);
</details> ```

### CODE REVIEW

**Comment by sidd-27** [comment]
run `cargo fmt` to fix the lint tests

File: crates/kornia-imgproc/src/resize.rs [hunk 1]
```diff
@@ -70,26 +68,29 @@ where
...
-    // and interpolate the values from the input image.
     let (dst_rows, dst_cols) = (dst.rows(), dst.cols());
-    let step_x = (src.cols() - 1) as f32 / (dst.cols() - 1) as f32;
-    let step_y = (src.rows() - 1) as f32 / (dst.rows() - 1) as f32;
-    let (map_x, map_y) = meshgrid_from_fn(dst_cols, dst_rows, |x, y| {
-        Ok((x as f32 * step_x, y as f32 * step_y))
-    })?;
-
-    // iterate over the output image and interpolate the pixel values
-    parallel::par_iter_rows_resample(dst, &map_x, &map_y, |&x, &y, dst_pixel| {
```
edgarriba, line 83: in that case maybe we can just remove this function ? /cc @sidd-27

crates/kornia-imgproc/src/resize.rs
```diff
@@ -70,26 +68,29 @@ where
...
-        // interpolate the pixel values for each channel
-        dst_pixel.iter_mut().enumerate().for_each(|(k, pixel)| {
-            *pixel = interpolate_pixel(src, x, y, k, interpolation);
-        });
+
+    let step_x = (src.cols() - 1) as f32 / (dst_cols - 1) as f32;
+    let step_y = (src.rows() - 1) as f32 / (dst_rows - 1) as f32;
+
+    //going for otf calculation instead of allocating mesh
```
edgarriba: keep meaningful comments

crates/kornia-imgproc/src/resize.rs
```diff
@@ -60,8 +59,7 @@ pub fn resize_native<const C: usize, A1: ImageAllocator, A2: ImageAllocator>(
     src: &Image<f32, C, A1>,
     dst: &mut Image<f32, C, A2>,
     interpolation: InterpolationMode,
-) -> Result<(), ImageError>
-where
+) -> Result<(), kornia_image::ImageError>
```
edgarriba:
```suggestion
) -> Result<(), ImageError>
```
use proper imports

sidd-27:
Maybe instead of re calculating the X_src every iteration with a fma operation, why not precaculate the X values before hand with a addition and just look up that value?

this gives me about a 2-3x speedup further, do benchmark this further and try it out

```
pub fn resize_native_new<const C: usize, A1: ImageAllocator, A2: ImageAllocator>(
    src: &Image<f32, C, A1>,
    dst: &mut Image<f32, C, A2>,
    interpolation: InterpolationMode,
) -> Result<(), kornia_image::ImageError>
{
    if src.size() == dst.size() {
        dst.as_slice_mut().copy_from_slice(src.as_slice());
        return Ok(());
    }

    let (dst_rows, dst_cols) = (dst.rows(), dst.cols());

    let step_x = (src.cols() - 1) as f32 / (dst_cols - 1) as f32;
    let step_y = (src.rows() - 1) as f32 / (dst_rows - 1) as f32;

    kornia_imgproc::parallel::par_iter_rows_indexed_mut(dst, |row_idx, row| {
        let y_src = row_idx as f32 * step_y;

        let mut x_cache = Vec::with_capacity(dst_cols);
        let mut x_val = 0.0f32;
        for _ in 0..dst_cols {
            x_cache.push(x_val);
            x_val += step_x;
        }

        for (col_idx, pix) in row.chunks_exact_mut(C).enumerate() {
            let x_src = x_cache[col_idx];

            pix.iter_mut().enumerate().for_each(|(k, p)| {
                *p = interpolate_pixel(src, x_src, y_src, k, interpolation);
            });
        }
    });

    Ok(())
}
```

VyasGuru:
> Maybe instead of re calculating the X_src every iteration with a fma operation, why not precaculate the X values before hand with a addition and just look up that value?
> 
> this gives me about a 2-3x speedup further, do benchmark this further and try it out

 tried it myself but it's coming out to be slower actually. i think allocating the cache vector is much heavier than the on-the-fly calculation.

crates/kornia-imgproc/src/resize.rs
sidd-27: yhea makes sense

sidd-27:
him and I are getting different results for the same benchmark test run over two machines, not just in time taken to run, which is obvious, but in which function is faster, his is a MacBook 3 i think, and mine is mid end HP laptop, in my device the cached version is 2x faster than the non cached version, and in his the non cached version is slightly faster than the cached version, how to decide which one is faster? @edgarriba

crates/kornia-imgproc/src/resize.rs
VyasGuru: yes, will sort out the optimisation @sidd-27 proposed with him and then make the necessary changes

edgarriba:
Benchmarks should be done in a Linux machine and eventually in a jetson device

crates/kornia-imgproc/src/resize.rs
edgarriba:
```suggestion
```

edgarriba (changes requested):
Follow copilot suggestions

sidd-27:
Considering that interpolation geometry is guaranteed to be identical for all channels, it is possible to remove the innermost loop over all channels and use a function that calculates the geometry for all channels at once. This gives me about a 2-3x speedup on my machine, definitely worth looking into

```
pub(crate) fn bilinear_interpolation_shared<const C: usize, A: ImageAllocator>(
    image: &Image<f32, C, A>,
    u: f32,
    v: f32,
    out: &mut [f32],
) {
    debug_assert!(out.len() == C);

    let rows = image.rows();
    let cols = image.cols();

    let iu = u as usize;
    let iv = v as usize;

    let iu1 = (iu + 1).min(cols - 1);
    let iv1 = (iv + 1).min(rows - 1);

    let fu = u - iu as f32;
    let fv = v - iv as f32;

    let wu0 = 1.0 - fu;
    let wv0 = 1.0 - fv;

    unsafe {
        // Convert &f32 → *const f32
        let p00 = image.get_unchecked([iv,  iu,  0]) as *const f32;
        let p01 = image.get_unchecked([iv,  iu1, 0]) as *const f32;
        let p10 = image.get_unchecked([iv1, iu,  0]) as *const f32;
        let p11 = image.get_unchecked([iv1, iu1, 0]) as *const f32;

        for c in 0..C {
            let v00 = *p00.add(c);
            let v01 = *p01.add(c);
            let v10 = *p10.add(c);
            let v11 = *p11.add(c);

            out[c] =
                v00 * wu0 * wv0 +
                v01 * fu  * wv0 +
                v10 * wu0 * fv  +
                v11 * fu  * fv;
        }
    }
}


// New code that does on-the-fly calculation for coordinates
pub fn resize_native_new<const C: usize, A1: ImageAllocator, A2: ImageAllocator>(
    src: &Image<f32, C, A1>,
    dst: &mut Image<f32, C, A2>,
    interpolation: InterpolationMode,
) -> Result<(), kornia_image::ImageError> {
    if src.size() == dst.size() {
        dst.as_slice_mut().copy_from_slice(src.as_slice());
        return Ok(());
    }

    let (dst_rows, dst_cols) = (dst.rows(), dst.cols());

    let step_x = if dst_cols > 1 {
        (src.cols() - 1) as f32 / (dst_cols - 1) as f32
    } else {
        0.0
    };

    let step_y = if dst_rows > 1 {
        (src.rows() - 1) as f32 / (dst_rows - 1) as f32
    } else {
        0.0
    };

    kornia_imgproc::parallel::par_iter_rows_indexed_mut(dst, |row_idx, row| {
        let y_src = row_idx as f32 * step_y;
        let mut x_src = 0.0f32;

        for pix in row.chunks_exact_mut(C) {
            bilinear_interpolation_shared::<C, _>(
                src,
                x_src,
                y_src,
                pix,
            );
            x_src += step_x;
        }
    });

    Ok(())
}
```

Here is the code I used to test the optimization, we may need to write new functions for all other interpolation modes tho, which work with this shared channel approach

current approach (with the new opti) - 420.59 µs
this approach - 203.81 µs

these were the benchmark results on 512x512 images, with a 0.0 max abs diff

edgarriba:
I think we should leverage the fact that channels pixels are contiguous in memory and access to the next channel by adding just positions and avoid recomputing the linear index each time

edgarriba:
What’s the decision here ? Please, rebase to main and address review comments if still interested in merging this. Otherwise we’ll close

VyasGuru:
> What’s the decision here ? Please, rebase to main and address review comments if still interested in merging this. Otherwise we’ll close

have made the necessary fixes for benchmarking and testing.

VyasGuru:
@edgarriba have implemented the changes suggested above, it is showing a 2x-3x speedup on all sizes now.

crates/kornia-imgproc/src/resize.rs
```diff
@@ -57,42 +58,99 @@ pub fn resize_native<const C: usize, A1: ImageAllocator, A2: ImageAllocator>(
...
+                    }
+                    x += step_x;
+                }
+            });
+        }
 
-    crate::parallel::par_iter_rows_indexed_mut(dst, |row_idx, row| {
-        let y_src = row_idx as f32 * step_y;
+        InterpolationMode::Bilinear => {
+            parallel::par_iter_rows_indexed_mut(dst, |row_idx, row| {
```
edgarriba, line 106: why not using or improving https://github.com/kornia/kornia-rs/blob/main/crates/kornia-imgproc/src/interpolation/interpolate.rs#L32

edgarriba: same for bilinear

VyasGuru: discussed this with @sidd-27 , the current bilinear and nn functions in interpolate return the data at the interpolated coord not the coords themselves, so we would need to call it for each channel even tho the coords are the same.

edgarriba: i believe those are just used here, i'd update to make the lib better instead of duplicating code

crates/kornia-imgproc/src/resize.rs
```diff
@@ -60,32 +58,99 @@ pub fn resize_native<const C: usize, A1: ImageAllocator, A2: ImageAllocator>(
...
+        InterpolationMode::Nearest => {
+            parallel::par_iter_rows_indexed_mut(dst, |row_idx, row| {
+                let iy = ((row_idx as f32 * step_y).round() as usize).min(src_rows - 1);
+                let mut x = 0.0f32;
+
+                for pix in row.chunks_exact_mut(C) {
+                    let ix = (x.round() as usize).min(src_cols - 1);
+
+                    unsafe {
+                        let src_ptr = src.get_unchecked([iy, ix, 0]) as *const f32;
```
edgarriba, line 95: to make this fast, i would access to the src once as get_unchecked has the extra cost of computer the linear index each time and pixels in the row are just in consecutive positions

crates/kornia-imgproc/src/resize.rs
```diff
@@ -60,32 +58,99 @@ pub fn resize_native<const C: usize, A1: ImageAllocator, A2: ImageAllocator>(
...
+
+                    // Pre-calculating weights once for all channels
+                    let w00 = (1.0 - fu) * wv0;
+                    let w01 = fu * wv0;
+                    let w10 = (1.0 - fu) * wv1;
+                    let w11 = fu * wv1;
+
+                    unsafe {
+                        // pointer math as pixel data is stored contiguously
+                        let p00 = src.get_unchecked([iv, iu, 0]) as *const f32;
```
edgarriba, line 129: proably similar as above

edgarriba:
what was missing here?

sidd-27:
i think we should rewrite the resize in general, following the lookup table approach that openCV and fast-image-resize use

edgarriba:
sounds good. Can we prioritize this function as seems quite a foundational one ? to match/improve existing implementations. It should be reusable for resample/warps too

Issue #538, opened 2025-11-07
Parallelize code using rayon


sidd-27:
A lot of the code in the library is still single-threaded, despite being "embarrassingly parallelizable" using rayons par_iter and other iterators. There are a lot of easy speedup wins to be had. this issue is also good for beginners as it won't have massive changes to the code and helps them learn a useful rust crate

edgarriba:
some comments here:

- we should improve the `parallel` apis to support better per row or pixel parallelization
- expose at runtime the num of threads to use
- give the option to not parallelize at all and use plane iterators (e.g for embedded devices)

5usu:
this will be a lot of work? should I give this a shot?

sidd-27:
I think the first step should be identifying all the functions which can be easily parallelized, and then like edgar mentioned, give options to not use parallelization at all etc

utk-avi:
> A lot of the code in the library is still single-threaded, despite being "embarrassingly parallelizable" using rayons par_iter and other iterators. There are a lot of easy speedup wins to be had. this issue is also good for beginners as it won't have massive changes to the code and helps them learn a useful rust crate

Hi sidd-27, I am new to open source contributions and I would like to contribute to this issue. Should I start with identifying functions in the chat itself or somewhere else, basically how do I start?

edgarriba:
We should check one by one in the code base

nilaysrivastava:
Hi @edgarriba and @sidd-27 ,

I would like to contribute to this.

Based on the discussion, I understand the requirement is not just to add `par_iter`, but to ensure we expose an option to control the thread count or disable parallelization entirely (for embedded contexts).

I plan to start by auditing the color or imgproc modules to identify a single candidate function. I will benchmark the current serial implementation, propose a parallelized version with the required "toggle" logic, and submit a PR for that specific function first.

Please let me know if there is a specific module you would prefer I prioritise.

edgarriba:
I'd say let's discuss a good api here. What's your suggestion ? there are possible some learnings to take from other PRs e.g @sidd-27 here https://github.com/kornia/kornia-rs/pull/596/changes#diff-aa12647f784bf331bb81fde95090a8c5d0a7fc103d2061195196ae596e027a9fR62

nilaysrivastava:
I was looking through the optimisations in #596 to see how we can standardise that logic.

[...]

PR #588, opened 2025-12-04
parallelize stdmean

VyasGuru:
# `std_mean` Benchmark Results

| Image Size | Sequential | Parallel | Speedup |
|-----------:|-----------:|----------:|--------:|
| **128×128** | 12.55 µs | 52.01 µs | **–76% (slower)** |
| **256×256** | 50.76 µs | 72.77 µs | **–30% (slower)** |
| **512×512** | 203.14 µs | 108.59 µs | **+87% faster** |
| **1024×1024** | 812.60 µs | 232.86 µs | **+248% faster** |
| **4096×4096** | 13.009 ms | 2.725 ms | **+377% faster** |

As expected, for larger sizes, parallelisation proves to be much faster. could implement a version that switches based on this "cutoff". However, that would require knowledge of the device core count. personally tested on two different systems, and both of them have a different  " cutoff". Do lmk on how to proceed with this. 

and also renamed variance to stddev

### DIFF
--- a/crates/kornia-imgproc/src/core.rs
+++ b/crates/kornia-imgproc/src/core.rs
@@ -42,30 +42,47 @@ use rayon::{
 /// assert_eq!(mean, [111.25, 112.25, 113.25]);
 /// ```
 pub fn std_mean<A: ImageAllocator>(image: &Image<u8, 3, A>) -> ([f64; 3], [f64; 3]) {
-    let (sum, sq_sum) = image.as_slice().chunks_exact(3).fold(
-        ([0f64; 3], [0f64; 3]),
-        |(mut sum, mut sq_sum), pixel| {
-            sum.iter_mut()
-                .zip(pixel.iter())
-                .for_each(|(s, &p)| *s += p as f64);
-            sq_sum
-                .iter_mut()
-                .zip(pixel.iter())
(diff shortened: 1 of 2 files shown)


crates/kornia-imgproc/src/core.rs
```diff
@@ -42,30 +42,47 @@ use rayon::{
...
-        },
-    );
+    //parallel version for the same(welford's algorithm)
+    let (sum, sq_sum) = image
+        .as_slice()
+        .par_chunks_exact(3)
+        .fold(
+            || ([0f64; 3], [0f64; 3]),
+            |(mut sum, mut sq_sum), pixel| {
+                sum[0] += pixel[0] as f64;
```
sidd-27: what if we accumulate in u64 and only convert to float at the end?

VyasGuru:
hi @edgarriba, could you please tell me how i should have proceeded here and why the pr was closed? Considering the significant improvement, i believe we should look into this further and would be more than happy to consider furthering this optimisation by including switches for different cases.

sidd-27:
someone's working on improving the parallel api to let the user more control over the parallelism, after that work is over that will probably be used to parallelize the function

edgarriba:
For ref #538

PR #590, opened 2025-12-12
ci: implement python docstring verification via doctest

### LINKED ISSUE #586: ci: implement docstring checking for python bindings
**Is your feature request related to a problem? Please describe.**
Currently, the Python bindings in `kornia-py` lack automated testing for documentation examples. Because of this, we have omitted code snippets from the docstrings to prevent them from becoming outdated or broken without notice (see discussion in PR #576).

**Describe the solution you'd like**
I would like to implement a CI check that verifies Python docstrings. This could involve:
1. Using tools like **`pytest --doctest-modules`** (Preferred approach per maintainer feedback).
2. Configuring a GitHub Actions workflow to run these checks on PRs.

**Additional context**
Once this CI check is in place, we can safely add rich code examples back into the Python binding documentation.

AnilKumarSingh9856:
### Description
This PR implements automated verification for Python docstrings using `pytest --doctest-modules`, Fixes #586.

### Solution
- Added a wrapper script `test_docs.py` that triggers `doctest` on the compiled `kornia_rs` module.
- Added a GitHub Actions workflow `.github/workflows/python-docs.yml` to run this check on every push and PR.
- This setup ensures that code examples in Python docstrings (like those removed in previous PRs) can be safely added back without fear of becoming outdated.

### Verification
- Validated locally by running `maturin develop` followed by `python test_docs.py`.
- The script correctly loads the compiled extension and reports 0 failures (attempted 0 for now as no examples exist yet, but the mechanism is functional).

### DIFF
--- a/.github/workflows/python-docs.yml
+++ b/.github/workflows/python-docs.yml
@@ -18,10 +18,10 @@ jobs:
         with:
           pixi-version: v0.40.1
           cache: true
-          environments: default
+          environments: dev
 
       - name: Build kornia_rs
-        run: pixi run --environment default py-build
+        run: pixi run --environment dev py-build
 
       - name: Run Doctests
-        run: pixi run pytest tests/test_doctests.py
+        run: pixi run --environment dev pytest tests/test_doctests.py
--- a/pixi.toml
+++ b/pixi.toml
@@ -193,6 +193,3 @@ toml-fmt-check = { cmd = "tombi format --check", description = "Check TOML forma
 clean-all = { depends-on = ["rust-clean", "py-clean", "cpp-clean"], description = "Clean all build artifacts" }
 test-all = { depends-on = ["rust-test", "py-test", "cpp-test"], description = "Run all tests" }
 fmt-all = { depends-on = ["rust-fmt", "toml-fmt", "cpp-fmt"], description = "Format all code" }
-
-[dependencies]
-pip = ">=24.3.1,<26"
--- a/pixi.lock
+++ b/pixi.lock
@@ -196,7 +196,6 @@ environments:
       - conda: https://conda.anaconda.org/conda-forge/noarch/packaging-25.0-pyh29332c3_1.conda
       - conda: https://conda.anaconda.org/conda-forge/linux-64/pango-1.56.4-hadf4263_0.conda
       - conda: https://conda.anaconda.org/conda-forge/linux-64/pcre2-10.47-haa7fec5_0.conda
-      - conda: https://conda.anaconda.org/conda-forge/noarch/pip-25.3-pyh145f28c_0.conda
       - conda: https://conda.anaconda.org/conda-forge/linux-64/pixman-0.46.4-h54a6638_1.conda
       - conda: https://conda.anaconda.org/conda-forge/linux-64/pkg-config-0.29.2-h4bc722e_1009.conda
       - conda: https://conda.anaconda.org/conda-forge/noarch/pluggy-1.6.0-pyhf9edf01_1.conda
@@ -425,7 +424,6 @@ environments:
       - conda: https://conda.anaconda.org/conda-forge/noarch/packaging-25.0-pyh29332c3_1.conda
       - conda: https://conda.anaconda.org/conda-forge/linux-aarch64/pango-1.56.4-he55ef5b_0.conda
       - conda: https://conda.anaconda.org/conda-forge/linux-aarch64/pcre2-10.47-hf841c20_0.conda
-      - conda: https://conda.anaconda.org/conda-forge/noarch/pip-25.3-pyh145f28c_0.conda
       - conda: https://conda.anaconda.org/conda-forge/linux-aarch64/pixman-0.46.4-h7ac5ae9_1.conda
       - conda: https://conda.anaconda.org/conda-forge/linux-aarch64/pkg-config-0.29.2-hce167ba_1009.conda
       - conda: https://conda.anaconda.org/conda-forge/noarch/pluggy-1.6.0-pyhf9edf01_1.conda
@@ -542,7 +540,6 @@ environments:
       - conda: https://conda.anaconda.org/conda-forge/noarch/packaging-25.0-pyh29332c3_1.conda
       - conda: https://conda.anaconda.org/conda-forge/osx-64/pango-1.56.4-h6ef8af8_0.conda
       - conda: https://conda.anaconda.org/conda-forge/osx-64/pcre2-10.47-h13923f0_0.conda
-      - conda: https://conda.anaconda.org/conda-forge/noarch/pip-25.3-pyh145f28c_0.conda
       - conda: https://conda.anaconda.org/conda-forge/osx-64/pixman-0.46.4-ha059160_1.conda
       - conda: https://conda.anaconda.org/conda-forge/osx-64/pkg-config-0.29.2-hf7e621a_1009.conda
       - conda: https://conda.anaconda.org/conda-forge/noarch/pluggy-1.6.0-pyhf9edf01_1.conda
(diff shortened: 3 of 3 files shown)


AnilKumarSingh9856:
@edgarriba, @AS1100K,
Update: Finalized Test Runner Logic
I have pushed a refactor of test_docs.py to ensure the CI is robust and future-proof.

Changes:
Recursive Discovery: The script now uses inspect to recursively find classes inside submodules (like apriltag) which doctest misses on compiled extensions.

Deduplication: Added a seen_ids check to prevent testing the same object twice (e.g., preventing double-testing of kornia_rs.resize due to internal self-references).

Verification: Verified locally that it correctly discovers 45+ objects (including submodules and main functions) and executes Python examples correctly.
The CI should now be green and ready for a final review.

edgarriba (changes requested):
I don’t see where is pytest --doctest-modules used

AnilKumarSingh9856:
@edgarriba I have pushed the fixes addressing your review.

**1. Regarding pytest (your question):** You are correct that pytest is not used. I removed the dependency. I initially considered it, but standard pytest collection struggles with compiled PyO3 extensions (it often fails to import or inspect binary modules correctly). Instead, I wrote test_docs.py to perform runtime introspection. This script explicitly imports the compiled kornia_rs, recursively finds all submodules, and runs doctest on them. This ensures we don't miss any corner cases in the compiled bindings.

**2. CI Fixes:**
Logic: Fixed the double-counting bug in test_docs.py (flagged by Copilot).
Toolchain: Switched to dtolnay/rust-toolchain and removed the redundant branch trigger.
Build: Fixed the maturin manifest path so it correctly builds the kornia-py crate in CI.

Verified locally with both positive and negative controls. CI should be green now. Ready for re-review! 🚀

AnilKumarSingh9856:
> Running python -m doctest -v‘ in ci doesn’t work ?

I verified this locally to be sure.

The Issue: Running python -m doctest -v kornia_rs fails with FileNotFoundError. The standard doctest CLI expects a file path or a python source file that it can parse. It cannot natively import and inspect a compiled .so / .pyd extension module to find docstrings hidden inside Rust structs.

The Solution (test_docs.py): My script works by importing the binary as a module in memory (import kornia_rs). It then uses dir() and getattr() to recursively walk through the module's contents (functions, classes, submodules) and manually feeds their __doc__ strings into the doctest runner.

This allows us to test examples inside nested compiled structs (e.g., kornia_rs.Image) that standard tools completely miss.

edgarriba:
if locally works, there must be a way that in CI too. The script approach it's too much to maintain

AnilKumarSingh9856:
> if locally works, there must be a way that in CI too. The script approach it's too much to maintain

I verified this locally to be sure.

The Issue: Running pytest --doctest-modules --pyargs kornia_rs results in collected 0 items. The standard test collector inspects .py source files but cannot recurse into the compiled .so extension to find docstrings attached to Rust structs.

The Solution (Integrated): I understand the concern about maintaining a standalone script. I have moved the introspection logic into a standard test file (tests/test_doctests.py).

Zero Config: It runs automatically with the standard pytest command.
No Extra Script: It lives inside the existing test suite.
Same Logic: It performs the necessary runtime recursion to find and execute the examples embedded in the binary.

I have pushed this change. It is now just another test file in the suite.

edgarriba (commented):
can you try something more like

```python
import doctest
import pytest
import kornia_rs
import inspect

def get_testable_objects(mod):
    """Recursively discover all objects with docstrings in the compiled module."""
    objects = []
    seen = set()

    def recurse(obj, name_prefix):
        if id(obj) in seen:
            return
        seen.add(id(obj))

        # 1. Add object if it has a docstring
        if hasattr(obj, "__doc__") and obj.__doc__:
            objects.append((obj, name_prefix))

        # 2. Inspect members
        # Using dir() is safer for extensions than inspect.getmembers
        for name in dir(obj):
            if name.startswith("_"): continue
            
            try:
                member = getattr(obj, name)
            except Exception:
                continue

            # Check if member belongs to kornia_rs (prevents recursion into stdlib)
            # Some builtins don't have __module__, handle gracefully
            mod_name = getattr(member, "__module__", None)
            if mod_name and "kornia_rs" in mod_name:
                if inspect.ismodule(member) or inspect.isclass(member):
                     recurse(member, f"{name_prefix}.{name}")

    recurse(mod, "kornia_rs")
    return objects

# Collect objects once
TEST_OBJECTS = get_testable_objects(kornia_rs)

@pytest.mark.parametrize("obj, name", TEST_OBJECTS)
def test_docstrings(obj, name):
    """
    Run doctest for a specific object.
    """
    # Create the context so 'kornia_rs' is available in the docstring examples
    globs = {"kornia_rs": kornia_rs}
    
    # Parser and Runner
    parser = doctest.DocTestParser()
    runner = doctest.DocTestRunner(verbose=False)
    
    # Create the test
    test = parser.get_doctest(obj.__doc__, globs, name, name, 0)
    
    # Run it
    runner.run(test)
    
    # Assert
    assert runner.failures == 0, f"Doctest failed for {name}. See output above."
```

AnilKumarSingh9856:
Thanks for the suggestion and the code snippet @edgarriba!
I agree that this approach using pytest.mark.parametrize is much cleaner. I will rebase against main and update the PR to implement this logic. I'll ping you again once the changes are pushed.

AnilKumarSingh9856:
Thanks for the guidance and the code snippet @edgarriba!

I have rebased against main and updated the implementation to use the @pytest.mark.parametrize approach you suggested. You are right—this is much cleaner than the previous script because it integrates natively with pytest's reporting.
I also verified locally that it runs correctly using the new pixi infrastructure (it correctly skips when no docstrings are found and fails when I inject a broken example). Ready for review!

AnilKumarSingh9856:
@edgarriba The PR Validation workflow is failing with a 403 error because it's trying to comment on this PR from a fork (which is restricted).
The bot is trying to warn that Issue #586 is unassigned.

Please assign Issue #586 to me. That will clear the warning, stop the bot from trying to comment, and make the CI pass. I'd rather not modify your workflow files in this PR.

.github/workflows/python-docs.yml
```diff
@@ -0,0 +1,43 @@
...
+
+    steps:
+      - uses: actions/checkout@v4
+
+      - name: Install Rust toolchain
+        uses: dtolnay/rust-toolchain@stable
+        with:
+          toolchain: stable
+
+      - name: Set up Python
```
edgarriba: this needs to use pixi to not redo the wheel

AnilKumarSingh9856:
@edgarriba The build is now passing with the Pixi setup. The only remaining failure is the 'PR Validation' check because the issue is unassigned. Could you please assign #586 to me?

.github/workflows/python-docs.yml
```diff
@@ -13,31 +13,15 @@ jobs:
...
+          pixi-version: v0.40.1
+          cache: true
+          environments: default
 
-      - name: Build and Install kornia_rs
-        run: |
-          maturin build --release --out dist -m kornia-py/Cargo.toml
-          pip install dist/*.whl
+      - name: Build kornia_rs
+        run: pixi run maturin develop -m kornia-py/Cargo.toml
```
edgarriba: there's a pixi task already ...

AnilKumarSingh9856:
Done changes as you suggested.

pixi.toml
```diff
@@ -193,3 +193,6 @@ toml-fmt-check = { cmd = "tombi format --check", description = "Check TOML forma
 clean-all = { depends-on = ["rust-clean", "py-clean", "cpp-clean"], description = "Clean all build artifacts" }
 test-all = { depends-on = ["rust-test", "py-test", "cpp-test"], description = "Run all tests" }
 fmt-all = { depends-on = ["rust-fmt", "toml-fmt", "cpp-fmt"], description = "Format all code" }
+
+[dependencies]
+pip = ">=24.3.1,<26"
```
edgarriba: why is this needed ?

AnilKumarSingh9856: I added pip previously to resolve a maturin error in CI, but I realize now that using the dev environment is the proper way to handle this, as it already provides uv. I've removed pip and updated the workflow to use the dev environment.

PR #595, opened 2025-12-20
fix(io): handle RGBA to RGB conversion in PNG reading

Meggison:
## Description

Fixes a critical bug where PNG images with alpha channels (RGBA/GrayscaleAlpha) caused a `Data length does not match the image size` error when read using typed readers that don't expect alpha channels.

**Note**: This PR has been rebased with main and now works alongside PR #571. While PR #571 introduced the new `read_image()` function with auto-detection (recommended API), this PR fixes the existing typed readers (`read_image_png_rgb8`, etc.) to properly validate color types.

## Problem

The CI was failing with this error in the Python library when loading `cells_binary.png`:

```python
FileExistsError: Data length (1449616) does not match the image size (1087212)
```

This occurred at:
```python
File /opt/hostedtoolcache/Python/3.12.12/x64/lib/python3.12/site-packages/kornia/io/io.py:61
img = kornia_rs.read_image_any(str(path_file))
```

### Root Cause Analysis

The error was caused by a mismatch between PNG color channels:
- **Data length**: 1,449,616 bytes (RGBA decoded from PNG)
- **Expected size**: 1,087,212 bytes (RGB image we tried to create)
- **Ratio**: 1,449,616 ÷ 1,087,212 = **1.333 = 4/3**

This proves the PNG contained **4 channels (RGBA)** but the code tried to create a **3 channel (RGB)** image.

## Solution

The fix modifies `read_png_impl()` in `crates/kornia-io/src/png.rs` to:

1. **Accept expected color type parameter** - Each typed reader specifies what color type it expects
2. **Validate PNG color type** before decoding - Detect the actual PNG color type and compare with expected
3. **Return clear error on mismatch** - Returns `IoError::PngColorTypeMismatch` instead of confusing buffer size errors
4. **Truncate buffer** to actual decoded size (since `output_buffer_size()` returns max possible size)
5. **Return ImageLayout** instead of size array - Provides complete image metadata
6. **Use IDENTITY transformations** - Keeps PNG data untouched for accurate color type detection

## Changes

- Added new error variant `PngColorTypeMismatch` in `crates/kornia-io/src/error.rs`
- Modified `read_png_impl()` to accept `expected_color_type` parameter and validate color types
- Updated `read_png_impl()` to return `ImageLayout` instead of `[usize; 2]`
- Updated all typed PNG readers to pass their expected color type:
  - `read_image_png_rgb8` expects `ColorType::Rgb`
  - `read_image_png_rgba8` expects `ColorType::Rgba`  
  - `read_image_png_mono8` expects `ColorType::Grayscale`
- Added color type validation that throws descriptive errors on mismatch
- Added buffer truncation to handle decoder's max buffer size allocation
- Added proper error documentation to all public reader functions

## Testing

✅ All PNG tests pass  
✅ Color type validation prevents the original CI error  
✅ Clear error messages guide users to use the correct reader or `read_image()` for auto-detection

### Verification

**Before (confusing buffer size error):**
```
Error: Data length (1454436) does not match the image size (1090827)
test result: FAILED
```

**After (clear color type error):**
```
Error: Invalid PNG color type (expected Rgb, found Rgba)
```

## Relationship with PR #571

- **PR #571**: Introduced `read_image()` with auto-detection (recommended API for handling PNGs with unknown color types)
- **This PR**: Adds proper validation to typed readers to fail fast with clear errors instead of cryptic buffer size mismatches
- **Combined**: Users can choose between auto-detection (`read_image()`) or explicit validation (typed readers)

## Impact

This prevents the CI failure in the kornia Python library by ensuring PNG color type mismatches are caught early with clear error messages. Users can either use the correct typed reader for their PNG format or use `read_image()` for automatic detection.


Meggison:
@edgarriba Still testing it, but it looks like it does work using a separate function to detect the channels and calls an appropriate reader depending on the image size using the read_image() function. My fix focused on stripping alpha channels in the case you got RGBA or more channels so it might be good to keep until the current function is deprecated. Would need to update the function calls in the Python library after but that shouldn't take too long

edgarriba:
@Meggison can you rebase with the new io changes and check if that's still needed?

Meggison:
@edgarriba changes are only needed for the deprecated functions

crates/kornia-io/src/png.rs
```diff
@@ -519,4 +580,37 @@ mod tests {
...
+        let file_path = tmp_dir.path().join("rgba_test.png");
+        write_image_png_rgba8(&file_path, &rgba_image)?;
+        let result = read_image_png_rgb8(&file_path);
+
+        // Debug: print the error
+        if let Err(ref e) = result {
+            println!("Error: {}", e);
+        }
+
+        assert!(
```
edgarriba: the assert should be comparing resulted and original bytes

edgarriba (changes requested):
the DS_STORE artifects shouldn't be here

crates/kornia-io/src/png.rs
Meggison: Noted. Is an additional test for the read_image fn needed here as well?

crates/kornia-io/src/png.rs
```diff
@@ -17,6 +17,11 @@ use std::{
 
 /// Read a PNG image as grayscale (Gray8).
 ///
+/// # Deprecation Note
+///
+/// For Python users: prefer using `kornia_rs.read_image()` which auto-detects format and type.
```
edgarriba: i wouldn't refer to any python usage in the Rust library

crates/kornia-io/src/png.rs
```diff
@@ -258,19 +289,49 @@ fn read_png_impl(file_path: impl AsRef<Path>) -> Result<(Vec<u8>, [usize; 2]), I
 
     let file = fs::File::open(file_path)?;
     let reader = BufReader::new(file);
-    let mut reader = Decoder::new(reader)
+    let mut decoder = Decoder::new(reader);
+    decoder.set_transformations(png::Transformations::IDENTITY);
```
edgarriba, line 301: why is this used ?

crates/kornia-io/src/png.rs
```diff
@@ -258,19 +289,49 @@ fn read_png_impl(file_path: impl AsRef<Path>) -> Result<(Vec<u8>, [usize; 2]), I
...
         .output_buffer_size()
         .ok_or_else(|| IoError::PngDecodeError("PNG output buffer size overflowed".into()))?;
     let mut buf = vec![0; buffer_size];
-    let info = reader
+    let frame_info = reader
         .next_frame(&mut buf)
         .map_err(|e| IoError::PngDecodeError(e.to_string()))?;
 
-    Ok((buf, [info.width as usize, info.height as usize]))
+    buf.truncate(frame_info.buffer_size());
```
edgarriba, line 327: why is this needed if we allocate the buffer based on the read output_buffer_size()?

crates/kornia-io/src/png.rs
```diff
@@ -25,12 +30,17 @@ use std::{
 ///
 /// A grayscale image (Gray8).
 pub fn read_image_png_mono8(file_path: impl AsRef<Path>) -> Result<Gray8<CpuAllocator>, IoError> {
-    let (buf, size) = read_png_impl(file_path)?;
+    let (buf, size) = read_png_impl(file_path, true)?;
```
edgarriba: ok, wondering if during the decode check that what color type is expected, and in this case if not grayscale just raise an error. Dropping channels could lead eventually to bugs

crates/kornia-io/src/png.rs
Meggison: keeps the PNG data untouched so we can detect the alpha channel and handle stripping manually. Without IDENTITY, the decoder may automatically convert RGBA to RGB

crates/kornia-io/src/png.rs
Meggison: I believe `output_buffer_size()` gives the max possible size, and might have extra zeros at the end that mess up the image size calculations.

crates/kornia-io/src/png.rs
Meggison: That's a good point, I had the same concern later on when I went through the `read_image()` implementation. Might be better to simply throw an error rather than silently strip the band. Also in the tests, seems there's a test using this function for RGBA images so this might only be valid as a CI fix

edgarriba: alirght, then let's do it in that other direction. Also make sure you rebase from main to use the new pixi infra

crates/kornia-io/src/error.rs
```diff
@@ -48,6 +48,10 @@ pub enum IoError {
     #[error("Failed to decode the png image. {0}")]
     PngDecodeError(String),
 
+    /// Error when PNG color type doesn't match expected type.
+    #[error("PNG color type mismatch. Expected {0}, but found {1}")]
+    PngColorTypeMismatch(String, String),
```
edgarriba: we could introduce a more advanced struct eg

```rust
#[error("Invalid PNG color type (expected {expected}, found {found})")]
PngColorTypeMismatch {
    expected: String,
    found: String,
},
```

crates/kornia-io/src/png.rs
```diff
@@ -17,89 +17,135 @@ use std::{
 
 /// Read a PNG image as grayscale (Gray8).
 ///
+/// # Deprecation warning: prefer using `kornia_rs.read_image()` which auto-detects format and type.
```
edgarriba: the rust crated do not have anything to do with the python lib

crates/kornia-io/src/png.rs
```diff
@@ -240,7 +286,10 @@ pub fn decode_image_png_layout(src: &[u8]) -> Result<ImageLayout, IoError> {
 }
 
 // utility function to read the png file
-fn read_png_impl(file_path: impl AsRef<Path>) -> Result<(Vec<u8>, [usize; 2]), IoError> {
+fn read_png_impl(
+    file_path: impl AsRef<Path>,
+    expected_color_type: ColorType,
+) -> Result<(Vec<u8>, [usize; 2]), IoError> {
```
edgarriba: consider returning ImageLayout https://github.com/kornia/kornia-rs/blob/main/crates/kornia-image/src/image.rs#L67

crates/kornia-io/src/png.rs
Meggison: Removed the python reference but I forgot to take them out for all the read functions. should I take out the deprecation warning completely?

edgarriba: yes, we are modifying here the downstream rust library. Which is totally unware of any python thing

PR #596, opened 2025-12-21
perf: optimized compute_histogram by 3.5x

sidd-27:
### Summary
This PR significantly optimizes the `compute_histogram` function, achieving a **3.88x to 5.1x** speedup across standard resolutions. The improvements focus on removing floating-point arithmetic in the hot loop, reducing memory allocation overhead, and improving parallel task granularity.

### Key Optimizations

**1. Removed Floating-Point Arithmetic (LUT + Integer Math)**
* **Before:** Calculated bin indices using floating-point division (`(pixel as f32 / scale).floor()`) for every pixel.
* **After:** Implemented a precomputed Look-Up Table (LUT) using integer arithmetic (`(i * num_bins) >> 8`). This eliminates costly float conversions and division operations inside the hot loop.

**2. Reduced Allocation Overhead (Rayon `map` vs `fold`)**
* **Before:** Iterated row-by-row using `.map()`. This forced the allocation of a new `Vec<usize>` buffer for *every single row* of the image (e.g., 1080 allocations for a 1080p image).
* **After:** Switched to `.fold()`. This reuses the accumulator vector across chunks, drastically reducing heap allocations to roughly one per thread rather than one per row.

**3. Improved Cache Locality & Task Granularity**
* **Before:** Chunked by `width` (one row). For small images or wide aspect ratios, this created uneven workloads.
* **After:** Uses `par_chunks(chunk_size) to ensure that each cpu thread gets the optimal workload`. 

### Benchmarks
Benchmarks were run using Criterion. The optimized implementation shows consistent scaling improvements, performing better as image resolution increases.

| Resolution | Original Time (Mean) | Optimized Time (Mean) | Speedup |
| :--- | :--- | :--- | :--- |
| **640x480 (VGA)** | 254.37 µs | 54.59 µs | **4.66x** |
| **1280x720 (HD)** | 626.24 µs | 161.53 µs | **3.88x** |
| **1920x1080 (FHD)** | 1.30 ms | 251.53 µs | **5.18x** |

```
use criterion::{criterion_group, criterion_main, Criterion, BenchmarkId};
use kornia_image::{allocator::{ImageAllocator, CpuAllocator}, Image, ImageError, ImageSize};
use rand::Rng;
use rayon::prelude::*;
use std::hint::black_box;

// ==========================================
// 1. ORIGINAL IMPLEMENTATION
// ==========================================
pub fn compute_histogram_original<A: ImageAllocator>(
    src: &Image<u8, 1, A>,
    hist: &mut [usize],
    num_bins: usize,
) -> Result<(), ImageError> {
    if num_bins == 0 || num_bins > 256 { return Err(ImageError::InvalidHistogramBins(num_bins)); }
    if hist.len() != num_bins { return Err(ImageError::InvalidHistogramBins(num_bins)); }

    let scale = 256.0 / num_bins as f32;
    let width = src.width();
    let src_slice = src.as_slice();

    let partial_hist = src_slice
        .par_chunks_exact(width)
        .map(|row| {
            let mut local_hist = vec![0_usize; num_bins];
            for &pixel in row {
                let bin = (pixel as f32 / scale).floor() as usize;
                local_hist[bin] += 1;
            }
            local_hist
        })
        .reduce(
            || vec![0; num_bins],
            |mut a, b| {
                for (i, val) in b.into_iter().enumerate() { a[i] += val; }
                a
            },
        );

    for (i, val) in partial_hist.into_iter().enumerate() { hist[i] += val; }
    Ok(())
}

// ==========================================
// 2. OPTIMIZED IMPLEMENTATION
// ==========================================
pub fn compute_histogram_optimized<A: ImageAllocator>(
    src: &Image<u8, 1, A>,
    hist: &mut [usize],
    num_bins: usize,
) -> Result<(), ImageError> {
    if num_bins == 0 || num_bins > 256 {
        return Err(ImageError::InvalidHistogramBins(num_bins));
    }

    if hist.len() != num_bins {
        return Err(ImageError::InvalidHistogramBins(num_bins));
    }

    let mut bin_lut = [0usize; 256];
    for (i, val) in bin_lut.iter_mut().enumerate() {
        *val = (i * num_bins) >> 8;
    }

    let total_pixels = src.width() * src.height();
    let num_threads = rayon::current_num_threads();
    let chunk_size = total_pixels / num_threads;

    let counts = src.as_slice()
        .par_chunks(chunk_size) 
        .fold(
            || vec![0usize; num_bins],
            |mut local_bin, chunk| {
                for &px in chunk {
                    unsafe {
                        let idx = *bin_lut.get_unchecked(px as usize);
                        *local_bin.get_unchecked_mut(idx) += 1;
                    }
                }
                local_bin
            },
        )
        .reduce(
            || vec![0usize; num_bins],
            |mut a, b| {
                for (i, val) in b.iter().enumerate() {
                    a[i] += val;
                }
                a
            },
        );

    for (i, count) in counts.iter().enumerate() {
        hist[i] += count;
    }

    Ok(())
}

fn create_test_image(width: usize, height: usize) -> Image<u8, 1, CpuAllocator> {
    let mut rng = rand::rng();
    let data: Vec<u8> = (0..(width * height)).map(|_| rng.random()).collect();
    let size = ImageSize { width, height };
    
    Image::new(size, data, CpuAllocator).expect("Failed to create image")
}

fn bench_histogram_compare(c: &mut Criterion) {
    let mut group = c.benchmark_group("HistogramCompare");
    
    let sizes = [(640, 480), (1280, 720), (1920, 1080)];
    let bins = 256;

    for (w, h) in sizes {
        let label = format!("{w}x{h}");
        
        let src = create_test_image(w, h);

        group.bench_with_input(
            BenchmarkId::new("original", &label),
            &src,
            |b, src| {
                let mut hist = vec![0; bins];
                b.iter(|| {
                    compute_histogram_original(
                        black_box(src),
                        black_box(&mut hist),
                        black_box(bins)
                    ).unwrap();
                })
            },
        );

        group.bench_with_input(
            BenchmarkId::new("optimized", &label),
            &src,
            |b, src| {
                let mut hist = vec![0; bins];
                b.iter(|| {
                    compute_histogram_optimized(
                        black_box(src),
                        black_box(&mut hist),
                        black_box(bins)
                    ).unwrap();
                })
            },
        );
    }

    group.finish();
}

criterion_group!(benches, bench_histogram_compare);
criterion_main!(benches);
```

### DIFF
--- a/crates/kornia-imgproc/src/histogram.rs
+++ b/crates/kornia-imgproc/src/histogram.rs
@@ -58,15 +58,22 @@ pub fn compute_histogram<A: ImageAllocator>(
         *val = (i * num_bins) >> 8;
     }
 
-    let counts = src
-        .as_slice()
-        .par_chunks(4096)
+    let total_pixels = src.width() * src.height();
+    let num_threads = rayon::current_num_threads();
+    
+    // max with 4*width to ensure divide by 0 error does not occur for very small images
+    let chunk_size = (total_pixels / num_threads).max(src.width() * 4);
+
+    let counts = src.as_slice()
+        .par_chunks(chunk_size) 
         .fold(
             || vec![0usize; num_bins],
             |mut local_bin, chunk| {
                 for &px in chunk {
-                    let idx = bin_lut[px as usize];
-                    local_bin[idx] += 1;
+                    unsafe {
+                        let idx = *bin_lut.get_unchecked(px as usize);
+                        *local_bin.get_unchecked_mut(idx) += 1;
+                    }
                 }
                 local_bin
             },
@@ -81,10 +88,10 @@ pub fn compute_histogram<A: ImageAllocator>(
             },
         );
 
-    for (bin, val) in hist.iter_mut().zip(counts.iter()) {
-        *bin += val;
+    for (i, count) in counts.iter().enumerate() {
+        hist[i] += count;
     }
-    
+
     Ok(())
 }
 


crates/kornia-imgproc/src/histogram.rs
```diff
@@ -53,35 +53,36 @@ pub fn compute_histogram<A: ImageAllocator>(
...
-            let mut local_hist = vec![0_usize; num_bins];
-            for &pixel in row {
-                let bin = (pixel as f32 / scale).floor() as usize;
-                local_hist[bin] += 1;
-            }
-            local_hist
-        })
+    let counts = src
+        .as_slice()
+        .par_chunks(4096)
```
edgarriba: why this magic number ?

crates/kornia-imgproc/src/histogram.rs
```diff
@@ -53,35 +53,36 @@ pub fn compute_histogram<A: ImageAllocator>(
...
+        .fold(
+            || vec![0usize; num_bins],
+            |mut local, chunk| {
```
edgarriba:
```suggestion
            |mut local_bin, chunk| {
```
to keep consistency with previous implementation

crates/kornia-imgproc/src/histogram.rs
```diff
@@ -53,35 +53,36 @@ pub fn compute_histogram<A: ImageAllocator>(
...
                     a[i] += val;
                 }
                 a
             },
         );
 
-    for (i, val) in partial_hist.into_iter().enumerate() {
-        hist[i] += val;
+    for (h, c) in hist.iter_mut().zip(counts.iter()) {
+        *h += c;
```
edgarriba:
```suggestion
        *hist += val;
```

crates/kornia-imgproc/src/histogram.rs
sidd-27: 4096 bytes is 4kb, which is the size of virtual memory pages, just to assure that the data remains in the cache in case of large but narrow images, I tried multiple different methods but this one was the fastest, on my machine atleast, ill try it on others too tho and change it if something else works better

edgarriba: Please try different bit I'm not quite sure this is generic enough across machines and architectures

sidd-27: changed it to make it so each thread gets equal workload, updated the benchmarks accordingly

crates/kornia-imgproc/src/histogram.rs
```diff
@@ -53,35 +53,43 @@ pub fn compute_histogram<A: ImageAllocator>(
...
+        *val = (i * num_bins) >> 8;
+    }
 
-    let width = src.width();
-    let src_slice = src.as_slice();
+    let total_pixels = src.width() * src.height();
+    let num_threads = rayon::current_num_threads();
+    
+    // max with 4*width to ensure divide by 0 error does not occur for very small images
+    let chunk_size = (total_pixels / num_threads).max(src.width() * 4);
```
edgarriba, line 65: @sidd-27 how we could formalize this as an api ?

edgarriba:

sidd-27:
ill open it after the parallelization api is done

PR #600, opened 2025-12-25
feat(imgproc): introduce ExecutionStrategy and optimize threshold (~30x speedup)

nilaysrivastava:
Partially addresses #538

### Summary
This PR introduces the `ExecutionStrategy` API to give users runtime control over parallelization. As discussed in #538, not all operations benefit from Rayon's default parallelism.

To validate this, I implemented the API in `threshold.rs` as a pilot. Benchmarks show that **forcing serial execution** for this memory-bound operation results in a **~30x speedup** on standard resolutions (1080p), proving the need for this granularity.

### Changes
1.  **Added `ExecutionStrategy` Enum** in `parallel.rs`:
    * `Auto`: Defaults to Rayon global pool (existing behavior).
    * `Serial`: Forces execution on the current thread (no overhead).
    * `Fixed(n)`: Creates a local thread pool with `n` threads.
2.  **Updated `threshold_binary`**:
    * Updated signature to accept `strategy: ExecutionStrategy`.
    * Implemented logic to switch between `par_iter` and standard `iter`.
3.  **Updated Benchmarks**:
    * Updated `bench_threshold.rs` to support the new signature and compare strategies.

### Performance Analysis
Benchmarked on Apple M-series (1920x1080 image).

| Strategy | Time (Mean) | Notes |
| :--- | :--- | :--- |
| **Auto (Parallel)** | **987.48 µs** | High overhead for simple ops |
| **Serial** | **31.42 µs** | **~30x Faster** (Zero overhead) |

Benchmarks show that for memory-bound operations like thresholding, the overhead of parallelism outweighs the benefits. By switching to Serial execution, we bypass this overhead, achieving a ~30x speedup (reaching memory bandwidth limits) on standard 1080p images compared to the default parallel implementation.

This confirms that for lightweight, memory-bound operations like thresholding, the overhead of managing threads significantly outweighs the benefits. The `Serial` strategy allows users (and future default optimizations) to bypass this overhead.

### Checklist
- [x] Implementation of `ExecutionStrategy`
- [x] Applied to `threshold.rs`
- [x] Verified with `cargo test`
- [x] Verified with `cargo bench`

### DIFF
--- a/crates/kornia-imgproc/benches/bench_threshold.rs
+++ b/crates/kornia-imgproc/benches/bench_threshold.rs
@@ -21,25 +21,25 @@ fn bench_threshold(c: &mut Criterion) {
     
     // 1. Benchmark Serial Execution
     group.bench_with_input(BenchmarkId::new("binary_serial", format!("{}x{}", w, h)), &src, |b, src| {
+        // Allocate outside to measure only algorithm performance
+        let mut dst = Image::from_size_val(src.size(), 0, CpuAllocator).unwrap();
         b.iter(|| {
-            // Allocate dst inside loop to ensure correctness (avoiding cache cheating)
-            let mut dst = Image::from_size_val(src.size(), 0, CpuAllocator).unwrap();
             threshold_binary(src, &mut dst, 127, 255, ExecutionStrategy::Serial).unwrap();
         })
     });
 
-    // 2. Benchmark AutoFull (Parallel)
-    group.bench_with_input(BenchmarkId::new("binary_auto_full", format!("{}x{}", w, h)), &src, |b, src| {
+    // 2. Benchmark Parallel Elements
+    group.bench_with_input(BenchmarkId::new("binary_parallel_elements", format!("{}x{}", w, h)), &src, |b, src| {
+        let mut dst = Image::from_size_val(src.size(), 0, CpuAllocator).unwrap();
         b.iter(|| {
-            let mut dst = Image::from_size_val(src.size(), 0, CpuAllocator).unwrap();
-            threshold_binary(src, &mut dst, 127, 255, ExecutionStrategy::AutoFull).unwrap();
+            threshold_binary(src, &mut dst, 127, 255, ExecutionStrategy::ParallelElements).unwrap();
         })
     });
 
     // 3. Benchmark AutoRows (Parallel Rows)
     group.bench_with_input(BenchmarkId::new("binary_auto_rows", format!("{}x{}", w, h)), &src, |b, src| {
+        let mut dst = Image::from_size_val(src.size(), 0, CpuAllocator).unwrap();
         b.iter(|| {
-            let mut dst = Image::from_size_val(src.size(), 0, CpuAllocator).unwrap();
             // stride = width * channels (1)
             threshold_binary(src, &mut dst, 127, 255, ExecutionStrategy::AutoRows(src.width())).unwrap();
         })
@@ -47,8 +47,8 @@ fn bench_threshold(c: &mut Criterion) {
 
     // 4. Benchmark Fixed (Custom Pool)
     group.bench_with_input(BenchmarkId::new("binary_fixed_4", format!("{}x{}", w, h)), &src, |b, src| {
+        let mut dst = Image::from_size_val(src.size(), 0, CpuAllocator).unwrap();
         b.iter(|| {
-            let mut dst = Image::from_size_val(src.size(), 0, CpuAllocator).unwrap();
             threshold_binary(src, &mut dst, 127, 255, ExecutionStrategy::Fixed(4)).unwrap();
         })
     });
--- a/crates/kornia-imgproc/src/parallel.rs
+++ b/crates/kornia-imgproc/src/parallel.rs
@@ -5,11 +5,11 @@ use kornia_image::{allocator::ImageAllocator, Image};
 use kornia_tensor::{CpuAllocator, Tensor2};
 
 /// Errors that can occur during parallel execution.
-#[derive(Error, Debug)]
+#[derive(Error, Debug, PartialEq)]
 pub enum ParallelError {
     /// The thread pool failed to build.
-    #[error("failed to build thread pool")]
-    BuildError(#[from] rayon::ThreadPoolBuildError),
+    #[error("failed to build thread pool: {0}")]
+    BuildError(String),
 
     /// The requested thread count is invalid.
     #[error("thread count must be > 0, got {0}")]
@@ -17,7 +17,11 @@ pub enum ParallelError {
 
     /// The row stride for AutoRows must be valid.
     #[error("row stride must be > 0 for AutoRows strategy")]
-    InvalidRowStride,
+    InvalidRowStride(usize),
+
+    /// Input and output sizes do not match.
+    #[error("source and destination slices must have the same length")]
+    SizeMismatch,
 }
 
 /// Controls how parallel operations are executed.
@@ -27,12 +31,12 @@ pub enum ExecutionStrategy {
     ///
     /// This maximizes parallelism but may have overhead for small operations.
     #[default]
-    AutoFull,
+    ParallelElements,
 
     /// Use the global Rayon thread pool to process rows (chunks) in parallel.
     ///
     /// You must provide the row stride (width * channels).
-    /// This is often more cache-friendly than [`ExecutionStrategy::AutoFull`].
+    /// This is often more cache-friendly than [`ExecutionStrategy::ParallelElements`].
     AutoRows(usize),
 
     /// Run sequentially on the current thread.
@@ -48,8 +52,6 @@ pub enum ExecutionStrategy {
     /// Use this primarily for benchmarking or specific isolation needs.
     Fixed(usize),
 }
-
-// --- LEGACY HELPERS (Restored to fix compilation errors) ---
 
 /// Apply a function to each pixel in the image in parallel.
 pub fn par_iter_rows<
@@ -171,8 +173,6 @@ pub fn par_iter_rows_resample<const C: usize, A: ImageAllocator>(
         });
 }
 
-// --- NEW ABSTRACTION ---
-
 /// Trait to execute operations on a slice with a given strategy.
 pub trait ExecuteExt<T> {
     /// Execute an operation on the slice with the given strategy.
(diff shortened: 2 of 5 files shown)


nilaysrivastava:
My apologies! I realised I staged the API definition but missed adding the implementation and benchmark files to the commit. I've just updated the PR with the full implementation.

nilaysrivastava:
That's a really nice abstraction. It definitely cleans up the call site and makes it way easier to reuse this logic for other operations.

I'll go ahead and refactor `parallel.rs` to add the `ExecuteExt` trait and then update `threshold.rs` to use it.

I'll push the updates shortly!

nilaysrivastava:
I added the `ExecuteExt` trait to `parallel.rs` and updated `threshold.rs` to use the new interface.

edgarriba:
One more comment on the design, is that in future this dispatcher should be able to launch cuda or other optimised kernels. Not be added now, but to keep in mind

crates/kornia-imgproc/src/parallel.rs
```diff
@@ -128,3 +140,41 @@ pub fn par_iter_rows_resample<const C: usize, A: ImageAllocator>(
...
+    /// * `strategy` - The execution strategy.
+    /// * `dst` - The destination slice.
+    /// * `op` - The operation to perform on each pixel.
+    fn execute_with<F>(self, strategy: ExecutionStrategy, dst: &mut [T], op: F)
+    where
+        F: Fn((&T, &mut T)) + Sync + Send;
+}
+
+impl<T: Sync + Send> ExecuteExt<T> for &[T] {
+    fn execute_with<F>(self, strategy: ExecutionStrategy, dst: &mut [T], op: F)
```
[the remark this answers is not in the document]
nilaysrivastava: Since &[T] implements Copy, passing self by value is standard and efficient here. Passing &self would add unnecessary double indirection.

crates/kornia-imgproc/src/parallel.rs
```diff
@@ -128,3 +140,41 @@ pub fn par_iter_rows_resample<const C: usize, A: ImageAllocator>(
...
+                self.iter().zip(dst.iter_mut()).for_each(op);
+            }
+            ExecutionStrategy::Auto => {
+                self.par_iter().zip(dst.par_iter_mut()).for_each(op);
+            }
+            ExecutionStrategy::Fixed(n) => {
+                let pool = rayon::ThreadPoolBuilder::new()
+                    .num_threads(n)
+                    .build()
+                    .expect("Failed to create thread pool");
```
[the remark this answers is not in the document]
nilaysrivastava: I kept the panic! here to keep the API ergonomic. Since Fixed(n) is primarily for benchmarking/testing, forcing the user to unwrap a Result every time would be cumbersome. I did add a descriptive error message to the panic as suggested.

nilaysrivastava:
Refactor pushed! I've addressed the review comments:

- Added `ExecuteExt` trait abstraction to `parallel.rs`.
- Added validation for `Fixed(0)` and clearer panic messages.
- Added docs warning about thread pool overhead.
- Added benchmarks for `Fixed` strategy to demonstrate overhead.
- Updated PR description to 'Partially addresses #538'.

crates/kornia-imgproc/src/parallel.rs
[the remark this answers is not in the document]
edgarriba: we probably want reusablility @sidd-27

crates/kornia-imgproc/src/parallel.rs
```diff
@@ -128,3 +151,99 @@ pub fn par_iter_rows_resample<const C: usize, A: ImageAllocator>(
...
+        match strategy {
+            ExecutionStrategy::Serial => {
+                self.iter().zip(dst.iter_mut()).for_each(op);
+            }
+            ExecutionStrategy::Auto => {
+                self.par_iter().zip(dst.par_iter_mut()).for_each(op);
+            }
+            ExecutionStrategy::Fixed(n) => {
+                if n == 0 {
+                    panic!("ExecutionStrategy::Fixed(n) requires n > 0");
```
edgarriba: no panics, return proper error via thiserror

crates/kornia-imgproc/src/parallel.rs
```diff
@@ -128,3 +151,99 @@ pub fn par_iter_rows_resample<const C: usize, A: ImageAllocator>(
...
+                }
+                let pool = rayon::ThreadPoolBuilder::new()
+                    .num_threads(n)
+                    .build()
+                    .unwrap_or_else(|e| {
+                        panic!("Failed to create thread pool with {} threads: {}", n, e);
```
edgarriba: propagate error

crates/kornia-imgproc/src/parallel.rs
```diff
@@ -128,3 +151,99 @@ pub fn par_iter_rows_resample<const C: usize, A: ImageAllocator>(
...
+                    });
+
+                pool.install(|| self.par_iter().zip(dst.par_iter_mut()).for_each(op));
```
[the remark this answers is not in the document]
edgarriba: return result

crates/kornia-imgproc/src/parallel.rs
```diff
@@ -128,3 +151,99 @@ pub fn par_iter_rows_resample<const C: usize, A: ImageAllocator>(
...
+impl<T: Sync + Send> ExecuteExt<T> for &[T] {
+    fn execute_with<F>(self, strategy: ExecutionStrategy, dst: &mut [T], op: F)
+    where
+        F: Fn((&T, &mut T)) + Sync + Send,
+    {
+        match strategy {
+            ExecutionStrategy::Serial => {
+                self.iter().zip(dst.iter_mut()).for_each(op);
+            }
+            ExecutionStrategy::Auto => {
```
edgarriba: i believe this should be named `AutoFull` and we should have a `AutoRows` that parallelize by rows /cc @sidd-27

nilaysrivastava:
Refactor pushed! I've addressed all the review comments and resolved the merge conflicts.

- Renamed `Auto` to `ExecutionStrategy::AutoFull` and added `ExecutionStrategy::AutoRows(stride)` to support row-based parallelism as requested.

- Removed all `panic!` calls in `parallel.rs`. I introduced a proper `ParallelError` enum using `thiserror`, and `execute_with` now returns a `Result`.

- Fixed the critical bug in `bench_threshold.rs` where `dst` was being reused (allocation is now inside the loop to ensure correctness) and registered the benchmark in `Cargo.toml`.

- Restored the `par_iter_rows` helper functions in `parallel.rs` so that existing modules (like `color`) continue to compile without changes.

Note on Error Propagation: Currently, `threshold_binary` uses `.expect()` at the very end because the return type is fixed to `Result<(), kornia_image::ImageError>`, and `ImageError` (in the separate `kornia-image` crate) does not yet have a variant to wrap `ParallelError`. I have added a `FIXME `comment there, we will need to extend `ImageError` in a future PR to allow proper propagation across the crate boundary.

crates/kornia-imgproc/benches/bench_threshold.rs
```diff
@@ -18,26 +18,37 @@ fn bench_threshold(c: &mut Criterion) {
     // We will test on Full HD only for clear results
     let (w, h) = (1920, 1080);
     let src = create_test_image(w, h);
-    let mut dst = Image::from_size_val(src.size(), 0, CpuAllocator).unwrap();
     
     // 1. Benchmark Serial Execution
     group.bench_with_input(BenchmarkId::new("binary_serial", format!("{}x{}", w, h)), &src, |b, src| {
         b.iter(|| {
+            // Allocate dst inside loop to ensure correctness (avoiding cache cheating)
+            let mut dst = Image::from_size_val(src.size(), 0, CpuAllocator).unwrap();
```
edgarriba: Benchmark should be only in the algorithm

crates/kornia-imgproc/src/parallel.rs
```diff
@@ -152,6 +171,8 @@ pub fn par_iter_rows_resample<const C: usize, A: ImageAllocator>(
         });
 }
 
+// --- NEW ABSTRACTION ---
```
edgarriba: No agentic comments

crates/kornia-imgproc/src/parallel.rs
```diff
@@ -160,36 +181,65 @@ pub trait ExecuteExt<T> {
...
+        &self,
+        strategy: ExecutionStrategy,
+        dst: &mut [T],
+        op: F,
+    ) -> Result<(), ParallelError>
     where
         F: Fn((&T, &mut T)) + Sync + Send,
     {
+        // Validation: slices must be same length.
+        debug_assert_eq!(
```
edgarriba: Raise error

crates/kornia-imgproc/src/parallel.rs
```diff
@@ -160,36 +181,65 @@ pub trait ExecuteExt<T> {
...
-                    panic!("ExecutionStrategy::Fixed(n) requires n > 0");
+                    return Err(ParallelError::InvalidThreadCount(n));
                 }
                 let pool = rayon::ThreadPoolBuilder::new()
                     .num_threads(n)
-                    .build()
-                    .unwrap_or_else(|e| {
-                        panic!("Failed to create thread pool with {} threads: {}", n, e);
-                    });
+                    .build()?; // Error propagates automatically via `?` thanks to thiserror
```
edgarriba: No redundant comments

crates/kornia-imgproc/src/threshold.rs
```diff
@@ -57,17 +58,21 @@ where
...
+                *dst_pixel = if *src_pixel > threshold {
+                    max_value
+                } else {
+                    T::zero()
+                };
+            },
+        )
+        // FIXME: kornia_image::ImageError does not yet have a variant for ParallelError.
+        // We expect() here to allow compilation until ImageError is extended.
+        .expect("Parallel execution failed");
```
edgarriba: Propagate error

nilaysrivastava:
Updates:

- Renamed `AutoFull` to `ParallelElements` in `ExecutionStrategy` for clarity.
- Extended `ImageError` in `kornia-image` with a `Parallel` variant to properly propagate errors, removing the `expect()` panic.
- Fixed Benchmarks by moving the destination image allocation outside the benchmark loop to measure only the algorithm's performance.
- Added Tests for edge cases (invalid stride, zero thread count).
- Fixed the `rayon` version mismatch to use the workspace default.
- Removed all internal/agentic comments.

sidd-27:
LGTM, but would be good to test atleast one other function to see how this works, thresholding is relatively simple, something more compute intensive perhaps

nilaysrivastava:
That's a fair point, thresholding is largely limited by memory bandwidth rather than CPU cores.

I noticed @edgarriba referenced PR #648 (Interpolation). My plan was to get this abstraction merged as a foundation, and then use that PR to validate the performance on a complex operation like `resize`.

I'm happy to add a synthetic benchmark here if you'd prefer to verify the scaling immediately, but integrating with #648 seems like the better test.

PR #611, opened 2025-12-31
feat: implement find_contours (#169)

mathurojus:
Implements find_contours function to detect contours in binary images.

Follows OpenCV API semantics with support for different retrieval modes and approximation methods. Updated from the development guidelines and all tests passing locally.


edgarriba (changes requested):
this is direct agent output which is not allowed. Please, read and understand the algorithm and use the agents to help you write. We cannot allow this PR in this library

sidd-27:
closed because this is just plain AI agent output, check out the updated AI policy

Issue #466, opened 2025-09-03
Implement `undistort_points`


edgarriba:
mimic https://docs.opencv.org/3.4/da/d54/group__imgproc__transform.html#ga55c716492470bfe86b0ee9bf3a1f0f7e

wtfashwin:
I have just opened a Pull Request to implement this feature. The PR includes a vectorized Rust implementation of undistort_points, Python bindings, and corresponding tests for both . #484
Looking forward to any feedback!

PR #623, opened 2026-01-07
feat: implemented undistort_points in kornia-imgproc (issue #466)

dhruvkjain:
This PR introduces a `undistort_points` function in `kornia-imgproc` using references from OpenCV's [undistortPoints()](https://docs.opencv.org/3.4/da/d54/group__imgproc__transform.html#ga55c716492470bfe86b0ee9bf3a1f0f7e),

Fixes: #466

The `undistort_points` function implements the Brown-Conrady polynomial distortion model (radial + tangential; no tilt or prism terms) and applies rectification and projection transformation optionally .

The `undistort_normalized_point_iter` function is the iterative solver implementation. The iterative solver uses a fixed-point method with a damping factor (alpha) of 1.0 and for each iteration it is reduced to half if update increases the error.

The `apply_r_and_p` function applies the rectification and projection transformation optionally.

Currently I have added #[allow(clippy::too_many_arguments)] for `undistort_points` as it has 8 arguments as I was following the OpenCV's implementation

### Benchmark Result:
```text
undistort_rust          time:   [393.24 µs 514.82 µs 669.10 µs]
                        change: [−39.419% −21.606% +3.3645%] (p = 0.10 > 0.05)
                        No change in performance detected.
Found 5 outliers among 100 measurements (5.00%)
  1 (1.00%) high mild
  4 (4.00%) high severe

Benchmarking undistort_opencv: Warming up for 3.0000 s
undistort_opencv        time:   [993.34 µs 997.71 µs 1.0066 ms]
                        change: [+65.072% +73.791% +81.767%] (p = 0.00 < 0.05)
                        Performance has regressed.
Found 3 outliers among 100 measurements (3.00%)
  2 (2.00%) high mild
  1 (1.00%) high severe
```

### DIFF
--- a/crates/kornia-imgproc/src/flip.rs
+++ b/crates/kornia-imgproc/src/flip.rs
@@ -120,16 +120,13 @@ where
         ));
     }
 
+    let row_len = src.cols() * C;
+
     dst.as_slice_mut()
-        .par_chunks_exact_mut(src.cols() * C)
-        .zip_eq(src.as_slice().par_chunks_exact(src.cols() * C).rev())
+        .par_chunks_exact_mut(row_len)
+        .zip(src.as_slice().par_chunks_exact(row_len).rev())
         .for_each(|(dst_row, src_row)| {
-            dst_row
-                .chunks_exact_mut(C)
-                .zip(src_row.chunks_exact(C))
-                .for_each(|(dst_pixel, src_pixel)| {
-                    dst_pixel.copy_from_slice(src_pixel);
-                })
+            dst_row.copy_from_slice(src_row);
         });
 
     Ok(())


dhruvkjain:
> can you provide an example showing that the image is well undistorted ?

`undistort_points` is for undistorting sparse point coordinates
for image processing example (that I have added in last commit), I have used bilinear splatting that pushes each distorted pixel onto the undistorted image because `undistort_points` provides an inverse mapping i.e. distorted -> undistorted (where a distorted image pixel is on undistorted image).

I have used these parameters:
```rust
    let intrinsic = CameraIntrinsic {
        fx: w as f64,
        fy: w as f64,
        cx: w as f64 / 2.0,
        cy: h as f64 / 2.0,
    };
    let distortion = PolynomialDistortion {
        k1: -0.2,
        k2: 0.05,
        k3: 0.0,
        k4: 0.0,
        k5: 0.0,
        k6: 0.0,
        p1: 0.0,
        p2: 0.0,
    };
    let criteria = TermCriteria {
        max_iter: 20,
        eps: 1e-6,
    };
```
 
f730e316-ffef-4de2-9c83-f607d681367a

examples/undistort_points_image/README.md
edgarriba, line 1: can you drag and drop a resulting image here

cjpurackal:
> looking at the output image, it's not very well rectified. Is it because of the parameters or some bug ? Is this from a known dataset with GT ? /cc @cjpurackal

+1 for providing a example with known dataset or an example cross checking with opencv. Also would be nice to have f32 support.

dhruvkjain:
> > looking at the output image, it's not very well rectified. Is it because of the parameters or some bug ? Is this from a known dataset with GT ? /cc @cjpurackal
> 
> +1 for providing a example with known dataset or an example cross checking with opencv. Also would be nice to have f32 support.

I took example image from here: https://opencv24-python-tutorials.readthedocs.io/en/latest/py_tutorials/py_calib3d/py_calibration/py_calibration.html

same can be found here in opencv's sample dataset: https://github.com/opencv/opencv/blob/master/samples/data/left12.jpg 

about f32 support should I add generics ?

edgarriba:
> same can be found here in opencv's sample dataset: https://github.com/opencv/opencv/blob/master/samples/data/left12.jpg

can you add the reference somewhere in the code ? how is the output of opencv close to ours ?

dhruvkjain:
> > same can be found here in opencv's sample dataset: https://github.com/opencv/opencv/blob/master/samples/data/left12.jpg
> 
> can you add the reference somewhere in the code ? how is the output of opencv close to ours ?

I have added the reference for opencv's dataset in readme, it already contains an image of output

cjpurackal:
LGTM, resolve the copilot issues.

cjpurackal:
> > > looking at the output image, it's not very well rectified. Is it because of the parameters or some bug ? Is this from a known dataset with GT ? /cc @cjpurackal
> > 
> > 
> > +1 for providing a example with known dataset or an example cross checking with opencv. Also would be nice to have f32 support.
> 
> I took example image from here: https://opencv24-python-tutorials.readthedocs.io/en/latest/py_tutorials/py_calib3d/py_calibration/py_calibration.html
> 
> same can be found here in opencv's sample dataset: https://github.com/opencv/opencv/blob/master/samples/data/left12.jpg
> 
> about f32 support should I add generics ?

Can you create a new issue for the f32 support? Let's discuss more there.

PR #626, opened 2026-01-07
reduced quickdecode memory usage by 770x

sidd-27:
# Pigeonhole-principle based quick decode table

### Problem Statement
The previous AprilTag decoding implementation relied on exhaustive precomputation of error permutations to achieve O(1) lookups. While effective for small tag families (e.g., 16h5), this approach is computationally and spatially prohibitive for larger families such as **52h13**.

For the 52h13 family, precomputing all variations with up to 2-bit errors results in approximately 67 million hash map entries. which amounts to approximately 400 MB of RAM being used for this simple check. If we were to extend this to 3-bit errors, the amount of space needed would increase exponentially to approximately 6 GB. Dedicating this many resources to a single check is not possible in many small robotics applications

### Proposed Solution
This PR replaces the combinatorial precomputation strategy with a search-based algorithm utilizing the **Pigeonhole Principle**.

Instead of storing every possible error permutation, the new implementation indexes the valid tags by splitting the code into 4 discrete chunks (for eg 13 bits each for 52h13). Given a maximum tolerance of 3-bit errors, the Pigeonhole Principle guarantees that at least one of the four chunks in an observed tag must match the valid tag perfectly.

The decoding process is updated to:
1. Perform lookups on the 4 chunks to identify candidate tags.
2. Compute the Hamming distance between the observed code and the candidate's perfect code.
3. Return the match if the distance is within the specified threshold.

### Technical Implementation Details
* **Data Structure:** Replaced the **open-addressing hash table** (linear probing `Vec<u16>` storing only Tag IDs) with a **chunk-based pigeonhole index** (`[Vec<Vec<u16>>; 4]`).
* **Memory Complexity:** The legacy implementation required storing every error permutation ($1 + N + \binom{N}{2}$) multiplied by a **load factor of 3** to resolve collisions. For the 52h13 family, this necessitated allocating over **4,100 slots per tag** (approx. 200 million total slots). The new implementation eliminates this combinatorial explosion, storing exactly **4 references per tag**. **52h13 Memory Footprint went from ~378 MB to ~500 KB which is nearly a 770x Reduction** 

### In summary - 
* **memory requirements:** 770x less memory than even the current optimized version of kornia, about 6700x smaller than the official apriltag library
* **initialization speed:** more than 6-700x faster than the original implementation.
* **Error correction:** now supports up to 3-bit error correction, the earlier only did 2, which is orders of magnitude better than the old implementation.
* **Regression:** Existing tests pass without modification.
* **performance:** The performance of the decode function has dropped a bit, but given that the decode function takes less than half a percent of the actual time taken in Apriltag, this is very worth it

### DIFF
--- a/crates/kornia-apriltag/benches/bench_decoding.rs
+++ b/crates/kornia-apriltag/benches/bench_decoding.rs
@@ -15,7 +15,7 @@ fn bench_decoding(c: &mut Criterion) {
     let mut gray_img = Image::from_size_val(img.size(), 0, CpuAllocator).unwrap();
     gray_from_rgb_u8(&img, &mut gray_img).unwrap();
 
-    let kornia_detector_config = DecodeTagsConfig::new(vec![TagFamilyKind::Tag36H11]);
+    let kornia_detector_config = DecodeTagsConfig::new(vec![TagFamilyKind::Tag36H11]).unwrap();
     let mut kornia_detector =
         AprilTagDecoder::new(kornia_detector_config, gray_img.size()).unwrap();
 
--- a/crates/kornia-apriltag/benches/bench_tagfamily.rs
+++ b/crates/kornia-apriltag/benches/bench_tagfamily.rs
@@ -15,7 +15,7 @@ fn bench_tagfamily(c: &mut Criterion) {
         let mut group = c.benchmark_group("Tag16H5");
         group.bench_function("kornia-apriltag", |b| {
             b.iter(|| {
-                let config = DecodeTagsConfig::new(vec![TagFamilyKind::Tag16H5]);
+                let config = DecodeTagsConfig::new(vec![TagFamilyKind::Tag16H5]).unwrap();
                 std::hint::black_box(AprilTagDecoder::new(config, IMG_SIZE).unwrap());
             });
         });
@@ -35,7 +35,7 @@ fn bench_tagfamily(c: &mut Criterion) {
         let mut group = c.benchmark_group("Tag36H11");
         group.bench_function("kornia-apriltag", |b| {
             b.iter(|| {
-                let config = DecodeTagsConfig::new(vec![TagFamilyKind::Tag36H11]);
+                let config = DecodeTagsConfig::new(vec![TagFamilyKind::Tag36H11]).unwrap();
                 std::hint::black_box(AprilTagDecoder::new(config, IMG_SIZE).unwrap());
             });
         });
@@ -57,7 +57,7 @@ fn bench_tagfamily(c: &mut Criterion) {
         let mut group = c.benchmark_group("Tag25H9");
         group.bench_function("kornia-apriltag", |b| {
             b.iter(|| {
-                let config = DecodeTagsConfig::new(vec![TagFamilyKind::Tag25H9]);
+                let config = DecodeTagsConfig::new(vec![TagFamilyKind::Tag25H9]).unwrap();
                 std::hint::black_box(AprilTagDecoder::new(config, IMG_SIZE).unwrap());
             });
         });
@@ -77,7 +77,7 @@ fn bench_tagfamily(c: &mut Criterion) {
         let mut group = c.benchmark_group("TagCircle21H7");
         group.bench_function("kornia-apriltag", |b| {
             b.iter(|| {
-                let config = DecodeTagsConfig::new(vec![TagFamilyKind::TagCircle21H7]);
+                let config = DecodeTagsConfig::new(vec![TagFamilyKind::TagCircle21H7]).unwrap();
                 std::hint::black_box(AprilTagDecoder::new(config, IMG_SIZE).unwrap());
             });
         });
@@ -97,7 +97,7 @@ fn bench_tagfamily(c: &mut Criterion) {
         let mut group = c.benchmark_group("TagCircle49H12");
         group.bench_function("kornia-apriltag", |b| {
             b.iter(|| {
-                let config = DecodeTagsConfig::new(vec![TagFamilyKind::TagCircle49H12]);
+                let config = DecodeTagsConfig::new(vec![TagFamilyKind::TagCircle49H12]).unwrap();
                 std::hint::black_box(AprilTagDecoder::new(config, IMG_SIZE).unwrap());
             });
         });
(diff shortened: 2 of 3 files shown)


rossng:
nice. I didn't know about [locality-sensitive hashing](https://en.wikipedia.org/wiki/Locality-sensitive_hashing), which I think is basically what this is.

this is just straight out of talking to claude but I think we can win more space and lookup time by turning the whole lookup table into one contiguous block rather than having lots of smaller heap allocations: https://github.com/rossng/kornia-rs/commit/44f331912f80ceebd7181a3a0884d5635236e3c7

The tagStandard52h13 memory usage also goes down to like 500kB. ~~There's surely some good SIMD potential in here too.~~

Edit: maybe not, given how small each bucket generally is


Tag Family | Codes | Chunk Size | Total Buckets | Non-Empty | Avg Size | Buckets ≥2 | Buckets ≥4
-- | -- | -- | -- | -- | -- | -- | --
Tag16H5 | 30 | 4 | 64 | 85.9% | 2.18 | 57.8% | 10.9%
Tag25H9 | 35 | 7 | 512 | 21.5% | 1.27 | 3.5% | 0.6%
Tag36H10 | 2,320 | 9 | 2,048 | 98.8% | 4.58 | 93.3% | 66.6%
Tag36H11 | 587 | 9 | 2,048 | 69.7% | 1.64 | 32.4% | 2.1%
TagCircle21H7 | 38 | 6 | 256 | 39.5% | 1.50 | 11.3% | 2.7%
TagCircle49H12 | 65,535 | 13 | 32,768 | 78.0% | 10.25 | 77.6% | 73.9%
TagCustom48H12 | 42,211 | 12 | 16,384 | 100.0% | 10.31 | 99.9% | 98.6%
TagStandard41H12 | 2,115 | 11 | 8,192 | 51.9% | 1.99 | 24.0% | 4.3%
TagStandard52H13 | 48,714 | 13 | 32,768 | 98.1% | 6.06 | 93.8% | 76.7%

rossng:
I also measured the impact of inlining the valid_ids into the contiguous array. For the large families it speeds things up maybe 15-25% (though memory cost goes back up to 2MB or so for tagStandard52h13).

Could be worthwhile for some use cases.

sidd-27:
> nice. I didn't know about [locality-sensitive hashing](https://en.wikipedia.org/wiki/Locality-sensitive_hashing), which I think is basically what this is.

Looking at the wiki, I don't think this is the same as locality sensitive hashing. That's a probabilistic algorithm, while this is deterministic

> this is just straight out of talking to claude but I think we can win more space and lookup time by turning the whole lookup table into one contiguous block rather than having lots of smaller heap allocations

ooh, ill look into it then and refactor this if needed

> I also measured the impact of inlining the valid_ids into the contiguous array. For the large families it speeds things up maybe 15-25% (though memory cost goes back up to 2MB or so for tagStandard52h13).

I don't think that small of an increase in lookup time would even be noticeable throughout the entire decoder, and that also blows up the memory needed to more than the average l2 cache size. so may actually slow things gown too

crates/kornia-apriltag/src/decoder.rs
```diff
@@ -94,82 +94,79 @@ impl GrayModelPair {
...
-        let entries_needed = ncodes // Hamming 0
-            + nbits * ncodes // Hamming 1
-            + ncodes * nbits * (nbits - 1) / 2; // Hamming 2
+    /// * `nbits` - The total number of bits in the tag (e.g., 36 for 36h11).
+    /// * `valid_codes` - The list of valid tag codes (the index is used as the ID).
+    /// * `allowed_errors` - Max bit errors to correct.
+    ///   - **WARNING:** Must be `< 4`. This implementation guarantees finding up to 3 errors.
+    ///   - set allowed_errors as 2 for 16h5 and other smaller families.
+    pub fn new(nbits: usize, valid_codes: &[usize], allowed_errors: u8) -> Self {
+        assert!(
```
edgarriba: propgate error

kornia-py/src/apriltag.rs
```diff
@@ -388,8 +388,9 @@ pub mod family {
     #[pymethods]
     impl PyQuickDecode {
         #[new]
-        pub fn new(nbits: usize, code_data: Vec<usize>) -> Self {
-            Self(QuickDecode::new(nbits, &code_data))
+        #[pyo3(signature = (nbits, code_data, allowed_errors=2))]
```
edgarriba: why the allowed errors?

edgarriba (approved):
LGTM

kornia-py/src/apriltag.rs
sidd-27: for some families, like 16h5, using 3-bit error correction could lead to wrong outputs etc, as the Hamming spheres would then intersect, so this gives an option to set the minimum Hamming error correction, default to two and then increase it if needed, but perhaps we could just set it to 2 for 16h5 and 3 for everything else

rossng: IMO the TagFamily can be responsible for setting up the QuickDecode with the correct values, and the QuickDecode itself can be relatively unopinionated

sidd-27: I don't have much experience using Apriltags, but @rossng  would always use 3-bit error correction whenever possible, ever be a problem? then we can keep the earlier structure, but for families like 16h5, just use 2 bit

rossng: I do think the user should be able to configure it, because some applications are more sensitive to false positives than others. But:

- it's locked to 2-bit in the current implementation anyway
- ~nobody depends on this code in production yet
 
I think making it configurable can be done separately. But we should perhaps default to 2-bit everywhere for now.

sidd-27: Because of how the algorithm is designed, 2 and 3-bit errors would take up the same space, but yes, ill hard set it to 2 for now, someone else can make it configurable later

sidd-27:
the memory can be reduced by another 25% if the codewords list is sorted first, but this, to me, seems like overkill, at least for now, we're now at about 500kbs for even the largest tag families. And I'm pretty happy with that. I've currently hardcoded the value of  2 for the decoder, but it can be changed to be 3-bit errors too, without any extra memory overhead. So @rossng, if you want to implement the configurability, then go ahead.

rossng:
yeah, I'll strip down my previous PR to just the configurability. All the stuff for creating different size lookup tables is redundant now 😄

sidd-27:
@edgarriba  i think this can be merged, @rossng  will then update this to be configurable i think

crates/kornia-apriltag/src/family/tag16h5.rs
```diff
@@ -12,8 +12,8 @@ impl TagFamily {
             bit_x: vec![1, 2, 3, 2, 4, 4, 4, 3, 4, 3, 2, 3, 1, 1, 1, 2],
             bit_y: vec![1, 1, 1, 2, 1, 2, 3, 2, 4, 4, 4, 3, 4, 3, 2, 3],
             code_data: CODE_DATA.into(),
-            quick_decode: QuickDecode::new(16, &CODE_DATA),
-            sharpening_buffer: SharpeningBuffer::new(64),
+            quick_decode: QuickDecode::new(16, &CODE_DATA, 2).unwrap(),
```
edgarriba: no unwraps

edgarriba:
feel free to merge

PR #646, opened 2026-01-14
perf: optimize box_blur with incremental columnwise sat

RakshithaKowlikar:
## 📝 Description
What this PR does

Replaces the separable filter implementation in box_blur() with an incremental column sums algorithm that achieves constant-time performance per row regardless of kernel size.

Why this change is needed
The previous separable filter became slower as kernel size increased, requiring more operations per pixel for larger kernels. This PR:
- Implements `columnar_sat()` using incremental column sums
- Uses minimal memory overhead
- Achieves constant performance regardless of kernel size
- Updates `box_blur()` to use the optimized implementation

Benchmark results
vs. `separable_filter`:

410 µs → 272 µs: 3×3 kernel (1.5x faster)
21.0 ms → 4.34 ms: 1024×896, 11×11 kernel (4.8x faster)

vs. box_blur_fast(`fast_horizontal_filter`):

2.3 ms → 272 µs: 256×224, small kernels (8.5x faster)
42 ms → 4.34 ms: 1024×896, 11×11 kernel (9.7x faster)

**Relates to:** # 461

---

## 🕵️ AI Usage Disclosure
*Check one of the following:*
- [ ] 🟢 **No AI used.**
- [ ] 🟡 **AI-assisted:** I used AI for boilerplate/refactoring but have manually reviewed and tested every line.

---

## 🚦 Checklist
- [ ] This PR strictly implements what the linked issue describes (no scope creep)
- [ ] I have performed a **self-review** of my code (no "ghost" variables or hallucinations).
- [ ] My code follows the existing style guidelines of this project.
- [ ] I have commented my code, particularly in hard-to-understand areas.

---

## 💭 Additional Context
Add any other context or screenshots about the pull request here.


sidd-27:
rebase is needed i think

crates/kornia-imgproc/src/filter/separable_filter.rs
```diff
@@ -256,6 +256,109 @@ pub(crate) fn fast_horizontal_filter<const C: usize, A1: ImageAllocator, A2: Ima
...
+                        row_acc[ch] += col_sums[(x_end - 1) * C + ch];
+                    }
+                }
+            }
+
+            let area = ((x_end - x_start) * (y_end - y_start)) as f32;
+            let out_idx = row_offset + c * C;
+            for ch in 0..C {
+                unsafe {
+                    *dst_data.get_unchecked_mut(out_idx + ch) = row_acc[ch] / area;
```
sidd-27: It would be good to calculate the inverse of the area and multiply instead of divide operations, compilers usually do that but just in case

crates/kornia-imgproc/src/filter/separable_filter.rs
```diff
@@ -256,6 +256,109 @@ pub(crate) fn fast_horizontal_filter<const C: usize, A1: ImageAllocator, A2: Ima
...
+                    }
+                }
+            }
+        }
+
+        for c in 0..src.cols() {
+            let x_start = c.saturating_sub(half_x);
+            let x_end = (c + half_x + 1).min(src.cols());
+
+            if c == 0 {
```
sidd-27: removing this and the r==0 cases outside their loops would be cleaner and avoids a if else statement in the loop

RakshithaKowlikar:
i'll rebase and make the changes

RakshithaKowlikar:
i'll look into the failing checks and make the necessary changes

sidd-27:
Could you provide a comparison against OpenCV or another similar library?

RakshithaKowlikar:
On my machine (1024×896, k=17):

opencv `cv::blur`: ~3.5 ms

`columnar_sat`: ~4.2 ms (≈ 3.7 Gelem/s)

Implementation-wise, `columnar_sat` follows the same incremental `RowSum` + `ColumnSum` sliding-window strategy used by opencv’s box filter (`modules/imgproc/src/box_filter.simd.hpp`). Border handling is more configurable in opencv.

sidd-27:
hmmm, i mean this is pretty good, could be merged after review by @edgarriba , but could you also look into why this is still a little slower than opencv?

RakshithaKowlikar:
i guess it is because opencv’s box blur uses hand-written SIMD kernels with runtime cpu dispatch

edgarriba (commented):
interesting approach using incremental column sums (SAT variant) for box blur.

concerns:

1. **naming**: `__dst__`, `__dst_data__`, `__col_sums__`, `__row_acc__` — the double underscore naming is non-idiomatic rust. use normal names like `col_sums`, `row_acc`.

2. **`wide` crate dependency**: you're pulling in `wide` for SIMD `f32x4`. this adds a dependency for a single function. have you benchmarked without SIMD? the compiler auto-vectorizes many of these patterns. if the `wide` version isn't significantly faster, prefer letting the compiler handle it.

3. **unsafe**: `load4`/`store4`/`update_row` are full of unsafe pointer arithmetic. this needs `// SAFETY:` comments at minimum, and ideally bounds checking with debug assertions.

4. **the approach**: incremental column sums is O(1) per pixel for the vertical direction, which is great. but the horizontal pass is still O(kernel_width) per pixel. for a true O(1) box blur you need a running sum in both directions (sliding window).

5. **benchmarks**: can you share numbers comparing this to the current separable filter implementation for different kernel sizes? the win should be bigger for larger kernels.

6. this replaces the separable filter path for `box_blur` entirely — make sure all existing tests pass.

@sidd-27 can you cross-review the SIMD parts?

sidd-27:
why not use f32x8 instead of 4?

RakshithaKowlikar:
I have addressed points 1, 2 and 3
I compared the SIMD version and created a scalar version for benchmark comparison
| Image Size | SIMD Time | Scalar Time | Improvement |
|-----------:|----------:|------------:|------------:|
| 256 × 224  | ~125 µs   | ~149 µs     | ~19% faster |
| 512 × 448  | ~508 µs   | ~591 µs     | ~16% faster |
| 1024 × 896 | ~2.02 ms  | ~2.36 ms    | ~17% faster |

SIMD is consistently ~17% faster. Kernel size 3, 7, 17, 51 all give nearly identical times.
Will focus on other points in consequent commits

RakshithaKowlikar:
f32x8 is not universally compatible. I guess it will create problems in few devices like ARM

edgarriba:
@RakshithaKowlikar thanks for the benchmark numbers and the clarification on `f32x8` — you are right that AVX2 (`f32x8`) is not universally available and would break on ARM (Neon only supports up to 128-bit = `f32x4`). Sticking with `f32x4` with the current approach is the right call for portability.

The ~17% improvement is solid for constant-time complexity regardless of kernel size. A few things to resolve from my earlier review before this can move forward:

1. **Naming**: please rename `__dst__`, `__col_sums__`, `__row_acc__` etc. to idiomatic Rust (`col_sums`, `row_acc`) — double underscores are non-idiomatic.
2. **`wide` crate**: was this dependency concern addressed? We prefer not to pull in a heavy SIMD crate if we can use `std::simd` (nightly) or a simpler approach.
3. **OpenCV comparison**: the ~17% over scalar is shown, but how does this compare to OpenCV box blur at the same sizes? That context is what reviewers and users will care about.

Once those are addressed, @sidd-27 can give the final pass.

sidd-27:
@RakshithaKowlikar wide does automatic fall back to lower size registers, pretty sure f32x8 would work

RakshithaKowlikar:
@sidd-27  it does fall back but i suppose it performs worse than f32x4 on ARM devices, but yes it would give better results than f32x4 in other devices
i thought of maintaining uniformity while choosing f32x4, please let me know if i should still change it to f32x8

sidd-27:
i think the fall back would be equally fast in arm, while being faster elsewhere, unless you have benchmarks to prove otherwise, i feel like f32x8 would be faster

RakshithaKowlikar:
consistently columnar_sat gives 2.4× speed up for kernel size 3 to 11.4× speed up for kernel size 17 across all images. 1024×896 Images take under 3ms and all other images in few hundred µs.

crates/kornia-imgproc/src/filter/separable_filter.rs
```diff
@@ -256,6 +257,232 @@ pub(crate) fn fast_horizontal_filter<const C: usize, A1: ImageAllocator, A2: Ima
...
+        *p,
+        *p.add(1),
+        *p.add(2),
+        *p.add(3),
+        *p.add(4),
+        *p.add(5),
+        *p.add(6),
+        *p.add(7),
+    ])
+}
```
sidd-27: `let sv = f32x8::from(&src_data[idx..idx + 8]);` this function can be replaced with this line, this is more readable

crates/kornia-imgproc/src/filter/separable_filter.rs
```diff
@@ -256,6 +257,232 @@ pub(crate) fn fast_horizontal_filter<const C: usize, A1: ImageAllocator, A2: Ima
...
+    let a = v.to_array();
+    *p = a[0];
+    *p.add(1) = a[1];
+    *p.add(2) = a[2];
+    *p.add(3) = a[3];
+    *p.add(4) = a[4];
+    *p.add(5) = a[5];
+    *p.add(6) = a[6];
+    *p.add(7) = a[7];
+}
```
sidd-27: `col_sums[out_idx..out_idx + 8].copy_from_slice(&r.to_array());` something like this here

sidd-27 (changes requested):

crates/kornia-imgproc/src/filter/separable_filter.rs
RakshithaKowlikar: I have made the necessary changes for all 3 reviews

crates/kornia-imgproc/src/filter/separable_filter.rs
```diff
@@ -256,6 +257,382 @@ pub(crate) fn fast_horizontal_filter<const C: usize, A1: ImageAllocator, A2: Ima
...
+                    *dst_data.get_unchecked_mut(out_idx + 1) = row_acc[1] * inv_area;
+                }
+            } else {
+                for (ch, &acc) in row_acc.iter().enumerate() {
+                    unsafe {
+                        *dst_data.get_unchecked_mut(out_idx + ch) = acc * inv_area;
+                    }
+                }
+            }
+        }
```
sidd-27, line 429: Does this unrolling actually speed things up?

RakshithaKowlikar: unrolling gives around 5 percent improvement across all configs of images. 
one of the examples:
256x224x7 
194.78 µs with unrolling
205.26 µs otherwise

RakshithaKowlikar: I can remove the unrolling if this improvement is negligible to be even considered.

sidd-27: sure then keep the unrolled version

sidd-27 (approved):

RakshithaKowlikar:
This pr has been approved but has been marked stale. Is there something else to be done?

edgarriba:
Check latest commits as this function has been optimised. Please provide benchmarks

Varun-sai-500:
@RakshithaKowlikar  Tests are failing

RakshithaKowlikar:
Yes. I will be working on it.

PR #648, opened 2026-01-14
Optimise interpolation utilities 

VyasGuru:
## Summary

This PR introduces an optimized implementation of `resize_native`, for both **nearest-neighbor** and **bilinear** interpolation. Across common image resolutions, the new implementation achieves **~2×–3× speedups**.

The optimizations focus on **eliminating per-channel overhead**, **exploiting contiguous memory layouts**, while preserving exact numerical semantics and edge behavior.

---

### 1. Vectorized Per-Pixel Interpolation 

- Compute **all `C` channels at once**, returning `[f32; C]`.
- Geometry and weights are computed **once per pixel**, not once per channel.

---

### 2. Direct Slice-Based Memory Access
- Computes linear base offsets once per pixel.
- Loads neighboring pixels as **contiguous slices** from `image.as_slice()`.

---

## Benchmarks

Benchmarks were run using **Criterion** on representative workloads.  
Results show consistent and scalable improvements, especially at higher resolutions.

| Benchmark | Original Time (Mean) | Optimized Time (Mean) | Speedup |
|---------|----------------------|------------------------|---------|
| Nearest Upscale RGB 512×512 | ~230 µs | ~84 µs | **~2.7×** |
| Bilinear Downscale RGB 256×256 | ~156 µs | ~70 µs | **~2.2×** |
| Bilinear Resize 1080×1920 (4K-scale) | ~3.38 ms | ~1.47 ms | **~2.3×** |

The optimized implementation scales better as image size increases, indicating improved cache utilization and reduced per-pixel overhead.

---

## Notes

- Have tested against old implementation and it is identical
- affine, remap and perspective use these util so they also should have similar speedups
- ran tests for all of these to ensure nothing broke in the newer implementation

### DIFF
--- a/crates/kornia-imgproc/src/resize.rs
+++ b/crates/kornia-imgproc/src/resize.rs
@@ -65,27 +65,48 @@ pub fn resize_native<const C: usize, A1: ImageAllocator, A2: ImageAllocator>(
     }
 
     let (dst_rows, dst_cols) = (dst.rows(), dst.cols());
-    let step_x = if dst.cols() > 1 {
-        (src.cols() - 1) as f32 / (dst.cols() - 1) as f32
-    } else {
-        0.0
-    };
-    let step_y = if dst.rows() > 1 {
-        (src.rows() - 1) as f32 / (dst.rows() - 1) as f32
-    } else {
-        0.0
-    };
-    let (map_x, map_y) = meshgrid_from_fn(dst_cols, dst_rows, |x, y| {
-        Ok((x as f32 * step_x, y as f32 * step_y))
-    })?;
 
-    // iterate over the output image and interpolate the pixel values
-    parallel::par_iter_rows_resample(dst, &map_x, &map_y, |&x, &y, dst_pixel| {
-        // interpolate the pixel values for each channel
-        dst_pixel.iter_mut().enumerate().for_each(|(k, pixel)| {
-            *pixel = interpolate_pixel(src, x, y, k, interpolation);
+    let step_x = if dst_cols > 1 {
+        (src.cols() - 1) as f32 / (dst_cols - 1) as f32
+    } else {
+        0.0
+    };
+
+    let step_y = if dst_rows > 1 {
+        (src.rows() - 1) as f32 / (dst_rows - 1) as f32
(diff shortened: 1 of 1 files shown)


crates/kornia-imgproc/src/interpolation/bilinear.rs
```diff
@@ -7,47 +7,55 @@ use kornia_image::{allocator::ImageAllocator, Image};
...
+
+    let base00 = (iv0 * cols + iu0) * C;
+    let base01 = (iv0 * cols + iu1) * C;
+    let base10 = (iv1 * cols + iu0) * C;
+
+    let base11 = (iv1 * cols + iu1) * C;
+
+    let data = image.as_slice();
+
+    let p00 = unsafe { data.get_unchecked(base00..base00 + C) };
```
edgarriba, line 50: what's the point of slice from/to the same index ?

VyasGuru: hmm? Are you referring to the p00? it is slicing from base00 to base00+C which is the entire pixel value RGB in contiguous memory

edgarriba: oh yes, my bad i was exopecting the semantic to be explicit `(base00..(base00 + C))` :)
all good

crates/kornia-imgproc/src/resize.rs
```diff
@@ -60,40 +58,55 @@ pub fn resize_native<const C: usize, A1: ImageAllocator, A2: ImageAllocator>(
...
+    dst.as_slice_mut()
+        .par_chunks_mut(row_stride)
+        .enumerate()
+        .for_each(|(y, row)| {
+            let v = y as f32 * step_y;
+
+            for (x, pixel) in row.chunks_exact_mut(C).enumerate() {
+                let u = x as f32 * step_x;
+
+                match interpolation {
```
edgarriba, line 92: i'm also curios to see the benchmark whether we match at every pixel, or just outside
to figure out this https://github.com/kornia/kornia-rs/pull/600 /cc @sidd-27

Issue #638, opened 2026-01-11
[Feature]: otsu thresholding


VyasGuru:
### 🚀 Feature Description

This is currently listed in the todos, not really difficult to implement, very well could follow openCV implementation. however i would suggest if we want to allow for multilevel otsu thresholding, then we should also offer non-trad method like the one ill attach down below, because as the no of levels grows, the histogram search becomes more and more expensive for the calculation of threshold. this makes the algo run fuch faster at the cost of a bit of deviation from trad otsu result, definitely a consideration imo.
[https://www.researchgate.net/publication/352067523_Multilevel_thresholding_of_images_with_improved_Otsu_thresholding_by_black_widow_optimization_algorithm](url)

i'd be happy to work on this, do lmk

### 📂 Feature Category

Rust Core Library

### 💡 Motivation

currently otsu is not yet implemented, leaving users only with constant thresholding, need adaptive thresholding.

### 💭 Proposed Solution

implement opencv version for the trad otsu, provide a diff option for faster thresholding

### 📚 Library Reference

OpenCV

### 🤝 Contribution Intent

- [x] I plan to submit a PR to implement this feature
- [ ] I'm requesting this feature but not planning to implement it

aviralgarg05:
I would like to solve the issue, please assign me

flipflagler:
Hi 👋  
I’ve been working on this issue for the past few days and have most of the implementation ready.  

I’d like to continue working on it and submit a PR soon.  
Could you please assign this issue to me?  

Thanks!

PR #661, opened 2026-01-22
feat(imgproc): add Otsu's thresholding method

flipflagler:
📝 Description

**Fixes:** #638

Implements Otsu's automatic thresholding algorithm for 1-channel u8 images. This is a classic image binarization technique that automatically determines the optimal threshold value by minimizing within-class variance.

---

## 🛠️ Changes Made
- [x] Implement `otsu_threshold()` function for automatic threshold selection using Otsu's method
- [x] Add `compute_histogram` import from histogram module  
- [x] Add comprehensive documentation with usage examples
- [x] Add unit test `test_otsu_threshold` for bimodal histogram case
- [x] Update TODO comment in `threshold.rs` (removed `otsu`, leaving `triangle`)

---

## 🧪 How Was This Tested?
- [x] **Unit Tests:** Added `test_otsu_threshold` - creates a bimodal histogram with 50 pixels at value 100 and 50 pixels at value 200, verifies correct binary separation
- [x] **Manual Verification:** Ran `cargo test -p kornia-imgproc --lib threshold::tests` - all 7 threshold tests pass
- [x] **Performance/Edge Cases:** Algorithm handles edge cases where one class has very few pixels (skips computation when class weight < 1e-6 to avoid division by zero)

### DIFF
--- a/crates/kornia-imgproc/src/threshold.rs
+++ b/crates/kornia-imgproc/src/threshold.rs
@@ -4,6 +4,7 @@ use std::cmp::PartialOrd;
 use kornia_image::{allocator::ImageAllocator, Image, ImageError};
 
 use crate::parallel;
+use crate::histogram::compute_histogram;
 
 /// Apply a binary threshold to an image.
 ///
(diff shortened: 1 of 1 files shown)


VyasGuru:
Since the image type is u8, only 255 values are possible, so maybe consider a LUT instead of the current approach. It would help a lot with larger images, removing branches. would recommend looking into it and bench it.
`dst.as_slice_mut()
        .iter_mut()
        .zip(src.as_slice().iter())
        .for_each(|(d, &s)| {
            *d = if s >= threshold { max_value } else { 0 };
        });`

sidd-27:
also provide benchmarking against openCV's implementation

PR #683, opened 2026-01-29
Create agents.md for kornia-rs

sidd-27:
used GeminiCLI's /init command + some manual changes, lemme know what else to add/remove


dhruvkjain:
can add something like `pre-commit run --show-diff-on-failure --color=always --all-file` for running all pre-commit checks locally

also for toml formatting:
- checking: `pixi run toml-fmt-check` (here list the environments from [pixi.toml](https://github.com/kornia/kornia-rs/blob/d772070acc80784aaf33fa838a753dd6de012038/pixi.toml#L12))
- auto formatting: `pixi run toml-fmt`

In performance, can add to use rayon parallelization wherever possible
maybe remove [best_practices.md](https://github.com/kornia/kornia-rs/blob/main/best_practices.md) and add its content here

Issue #671, opened 2026-01-27
[Refactor]: remove clone in Apriltag families


sidd-27:
**Title:** Optimize `TagFamilyKind::Custom` cloning to avoid deep copies

### Description

The `TagFamilyKind` enum currently stores custom tag families using a `Box`:

```rust
Custom(Box<TagFamily>),

```

This causes a full deep copy of the `TagFamily` data every time the enum is cloned. Since this data is typically read-only configuration, duplicating it is inefficient in terms of both memory and performance.

The goal is to address the `TODO` in the codebase by refactoring this variant to support cheap cloning (avoiding the allocation and copy) while maintaining necessary thread-safety guarantees.

littleKitchen:
Hi, I would like to work on this if assigned to me. Thanks!

AdityaShome:
Hello, I have locally refactored the variant to `Custom(Arc<TagFamily>)` that simply increments a reference count and copies the pointer, rather than duplicating the immutable underlying data. This enables cloning by sharing ownership of the read only family data instead of duplicating it.
I have added unit tests which asserts Arc::ptr_eq() on cloned instances, confirming that both reference pointers are identical and point to the same shared memory address. 
Happy to help in the issue.

JagerBomb69:
Hi,
I’d like to work on #671 (Refactor: remove clone in AprilTag families) and would appreciate your assignment before opening a PR.
What I’ve done
Refactored TagFamilyKind::Custom from Box<TagFamily> to Arc<TagFamily> so cloning only increments a refcount and shares the same data instead of doing a full deep copy.
Updated both the Rust crate (crates/kornia-apriltag) and the Python bindings (kornia-py) in one branch (all relevant call sites: add, into_tag_family_kind, with_max_hamming, etc.).
Added a unit test that uses Arc::ptr_eq() to confirm cloned Custom instances share the same allocation.
Removed the old TODO and added a short doc comment that Custom is stored in an Arc for cheap cloning.
All existing tests pass (kornia-apriltag and kornia-py).
Benefits
Performance: Cloning TagFamilyKind (including Custom) is O(1); no more copying large TagFamily data on every clone.
Memory: Read-only family data is shared via Arc instead of duplicated.
Single PR: Rust refactor and Python bindings are in one PR, so it’s easier to review and merge.
Verification: The Arc::ptr_eq() test documents and enforces the intended “cheap clone, shared allocation” behavior.
[...]

PR #686, opened 2026-01-30
Refactor: apriltag family arc cloning and update bindings

AdityaShome:
#### **Fixes/Relates to:** #671
---

## 🛠️ Changes Made
- [x] Changed TagFamilyKind::Custom storage from Box to Arc for cloning.
- [x] Optimized `From` function implementation to move data into Arc without copying.
- [x] Used Arc::try_unwrap in TryFrom to allow zero copy extraction for unique owners.
- [x] Updated Python bindings in apriltag.rs to handle Arc ownership and shared references.
---

## 🧪 How Was This Tested?
- [x] **Unit Tests:** 
- verifies that cloning TagFamilyKind::Custom Arc pointers shares the same memory location
- added unit test for Arc unwrapping and shared ownership logic
- [x] **Manual Verification:** Successfully ran pixi python binding tests to ensure FFI stability.

---

## 🕵️ AI Usage Disclosure
*Check one of the following:*
- [ ] 🟢 **No AI used.**

---

## 🚦 Checklist

---


AdityaShome:
In the this branch, made initial commits TagFamilyKind holding an Arc pointer are made, which enables zero copy cloning.
However, it is incomplete because the main detector DecodeTagsConfig still stores the old TagFamily struct. This forces it to "unwrap" and deep copy the data when adding it to the detector, negating the performance benefit until we switch the storage type to TagFamilyKind (which is done in the migration branch).

So I would like to update DecodeTagsConfig to store `Vec<TagFamilyKind>` instead of `Vec<TagFamily>`. 
CLI was failing because of premature usage of `DecodeTagsConfig::add(TagFamilyKind)` in `self.0.add(family_kind.0);`, which I was not commiting for scope of issue. But commiting as a whole might be a better idea.
I had done the changes and verified by maturin and python script in the env.

AdityaShome:
image
image
@sidd-27 I ran benchmarks to measure the impact of this refactor: cloning performance improved 106x (392ns → 3.7ns for TagFamilyKind::Custom), and decoder detection performance improved 4.1% (6.44ms → 6.18ms per frame) by hoisting the 
TagFamily conversion out of the quad loop. The optimization is particularly impactful for larger tag families (e.g., TagStandard52H13) and scales better with noisy images that generate hundreds of candidate quads, as we now allocate TagFamily buffers once per frame instead of once per quad candidate.

AdityaShome:
The benchmark is pushed, could please you review it once?

sidd-27 (approved):

sidd-27:
thanks for the benchmark, but i dont think theres any need to push it to main, we'd like to benchmark operations like decoding etc, apart from that LGTM

sidd-27:
> Building tag_families clones each TagFamilyKind multiple times per entry and stores both (TagFamilyKind, TagFamily), which may add avoidable overhead in hot paths. Consider restructuring to clone once per element (e.g., clone into a local kind_cloned) and/or avoid storing redundant data if feasible.
> 
> ```
> let mut tag_families: Vec<(TagFamilyKind, TagFamily)> = config
>     .tag_families
>     .iter()
>     .filter_map(|kind| {
>         TagFamily::try_from(kind.clone())
>             .ok()
>             .map(|family| (kind.clone(), family))
>     })
>     .collect();
> 
> quads.iter_mut().for_each(|quad| {
>     if config.refine_edges_enabled {
>         refine_edges(src, quad);
>     }
> 
>     if !quad.update_homographies() {
>         return;
>     }
> 
>     tag_families.iter_mut().for_each(|(kind, family)| {
>         if family.reversed_border != quad.reversed_border {
> ```

can this be addressed?

AdityaShome:
The original code called `TagFamily::try_from(kind.clone())` cloned the `&TagFamilyKind` into an owned value to use the `TryFrom<TagFamilyKind> impl`, since `TryFrom<&TagFamilyKind>` exists so we can use the reference, eliminating the first unnecessary clone and only cloning for the tuple storage.
Another one, there is todo comment to avoid allocation every decode call, it is addressed partially by moving TagFamily conversion outside quad loop, thats good 50-200 reduction in heap  allocation. But still tagfamily is rebuilt every decode call, so caching in AprilTagDecoder could be done for it for making it one time cost, should I do it in follow up as it extends beyond the scope of issue?

sidd-27:
if the change is small, just do it here only, otherwise leave a todo and maybe someone else can pick it up

AdityaShome:
Sure, will do it here.

AdityaShome:
- Replaced the per call tag family rebuild logic with cache.
- Added a cache field to the AprilTag decoder struct to store pre-built tag family data
- The cache is built once during decoder construction and reused, new tag family also updates the cache
- Changed the decode function to accept a slice of pre-built tag family pairs and individual config values

Decoder Benchmark Results:

All 9 Tag Families:
| Variant | Time | Speedup |
|---------|------|---------|
| **Cached** | 8.89 ms | **31% faster** |
| Uncached | 12.92 ms | baseline |

With all 9 families, caching saves ~4 ms per frame by avoiding repeated lookup table construction.
I have resolved the merge conflicts. Could you review once?

sidd-27:
theres still conflicts, LGTM otherwise

sidd-27:
the code doesn't compile

AdityaShome:
In local, `cargo test -p kornia-apriltag`, `cargo test -p kornia-py` and `cargo test --all-targets --no-run` are compiling, so was trying to find the bug, could you please specify in which crate there is a compilation issue, once its done, I will fix linting errors.

edgarriba:
Hey @AdityaShome 👋 — _(posting on behalf of @edgarriba)_

CI is currently failing on this PR and needs some fixes before we can merge:

- ❌ **Clippy** — there are Rust lint errors to address
- ❌ **cargo check** — compilation check failing
- ❌ **Test Debug/Release x86_64** — tests failing on x86_64
- ❌ **Test Debug aarch64** — tests failing on aarch64
- ❌ **Validate PR Requirements** — also needs attention

The C++, Python bindings, pre-commit, and format checks are all passing 👍, so you're most of the way there.

Could you check the Clippy output and test failures and push a fix? Happy to help look at the errors if you paste them here. Thanks!)

sidd-27:
> In local, cargo test -p kornia-apriltag, cargo test -p kornia-py and cargo test --all-targets --no-run are compiling, so was trying to find the bug, could you please specify in which crate there is a compilation issue, once its done, I will fix linting errors.

there was a duplicate import in the code, no idea how it compiled locally, fixed it dw

sidd-27 (approved):

crates/kornia-apriltag/src/family/mod.rs
```diff
@@ -162,35 +193,23 @@ impl From<&mut TagFamily> for TagFamilyKind {
...
-            TagFamilyKind::Tag25H9 => TagFamily::tag25_h9(),
-            TagFamilyKind::Tag36H10 => TagFamily::tag36_h10(),
-            TagFamilyKind::Tag36H11 => TagFamily::tag36_h11(),
-            TagFamilyKind::TagCircle21H7 => TagFamily::tagcircle21_h7(),
-            TagFamilyKind::TagCircle49H12 => TagFamily::tagcircle49_h12(),
-            TagFamilyKind::TagCustom48H12 => TagFamily::tagcustom48_h12(),
-            TagFamilyKind::TagStandard41H12 => TagFamily::tagstandard41_h12(),
-            TagFamilyKind::TagStandard52H13 => TagFamily::tagstandard52_h13(),
-            TagFamilyKind::Custom(tag_family) => Ok(*tag_family),
+            // Keeping the Custom optimization
```
sidd-27: Comments like these could be confusing

AdityaShome: It is resolved

sidd-27:
LGTM, feels ready to merge, ill let someone else take a look too

sidd-27 (commented):
## Review: PR #686 — Refactor apriltag family Arc cloning and update bindings

Overall a solid refactoring. The `Box` → `Arc` change for `TagFamilyKind::Custom` is well-motivated and the cached families optimization in `AprilTagDecoder` is a nice architectural improvement. A few issues to address:

---

### 🔴 Issues

**1. `unwrap_or_default()` in Python bindings (apriltag.rs:L26)**
```rust
let kinds = tag_family_kinds.unwrap_or_default();
```
Per project guidelines, `.unwrap()` variants should be avoided in library code. While `unwrap_or_default()` is safe, this silently accepts `None` and creates an empty config — is that intentional? If callers should always provide families, keep it required. If optional is intended, document the behavior.

**2. `From<TagFamily>` now always wraps in `Custom` (family/mod.rs)**
```rust
impl From<TagFamily> for TagFamilyKind {
    fn from(value: TagFamily) -> Self {
        TagFamilyKind::Custom(Arc::new(value))
    }
}
```
Previously this mapped known families (e.g. tag36_h11) back to their enum variants. Now **every** `TagFamily` becomes `Custom(Arc<...>)`, even built-in ones. The `test_safety_no_silent_data_loss` test shows this is intentional for preserving user modifications, but this is a **behavioral change** that should be called out more explicitly in the PR description. Downstream code pattern-matching on specific `TagFamilyKind` variants (not `Custom`) could break silently.

**3. Fallback magic number in `add()` (lib.rs)**
```rust
(9, false)  // fallback for standard tags if try_from fails
```
If `TagFamily::try_from(kind.clone())` fails for a standard variant, that's a bug — not something to silently default. This should propagate the error or at least log a warning. A silent `(9, false)` could mask real issues.

**4. Redundant `min_tag_width` guard (lib.rs)**
```rust
if min_tag_width == usize::MAX {
    min_tag_width = 9;
}
```
This handles the empty tag families case, but `DecodeTagsConfig::new` with empty families should arguably be an error, not silently defaulting. Combined with the `Default` derive, someone could create a config with no families and get unexpected behavior.

---

### 🟡 Suggestions

**5. `From<&mut TagFamily>` still uses `to_tag_family_kind_impl` which maps builtins back to enum variants.** But `From<TagFamily>` (owned) always returns `Custom`. This asymmetry could confuse callers — `From<&TagFamily>` and `From<&mut TagFamily>` behave differently from `From<TagFamily>`.

**6. `FromStr` implementation** — Good addition, but missing doc comment per project guidelines. All `pub` items need documentation with `# Errors` section.

**7. `builtin_tag` function** — Should be `pub(crate)` or documented if intended for external use. Currently it's private, which is fine, but a brief doc comment would help.

**8. `decode_tags` signature change** — Breaking out `config` into individual params (`tag_families`, `refine_edges_enabled`, `decode_sharpening`) improves flexibility, but the doc comment for `sharpening_buffer` was removed without adding one for the new params in the `# Arguments` section. Keep docs in sync.

---

### ✅ What's Good

- Arc sharing is verified with `Arc::ptr_eq` test — nice
- `try_unwrap` with clone fallback is the right pattern for Arc extraction
- Cached families in `AprilTagDecoder` avoids repeated `TryFrom` conversions
- Python bindings simplified nicely by accepting `TagFamilyKind` directly
- Tests cover both unique and shared ownership paths

---

**Verdict:** Needs a few fixes before merge — mainly the behavioral change in `From<TagFamily>` should be documented/intentional, the silent fallbacks need addressing, and doc comments should be updated per project conventions.

sidd-27:
testing out automated reviews a couple of raised concerns are valid here

sidd-27 (changes requested):
look at the review comments

sidd-27:
@AdityaShome  are you still working on this?

AdityaShome:
For issues 1-4:

  - Removed the optional None path in the Python DecodeTagsConfig constructor, so callers must now pass tag families explicitly instead of silently getting an empty config.
  - Kept From<TagFamily> for TagFamilyKind returning Custom(Arc<_>), but documented that this is intentional so user-provided family data is preserved exactly rather than normalized by builtin name.
  - Removed the silent (9, false) fallback in DecodeTagsConfig::add(). Builtin family conversion now uses the existing conversion path and fails explicitly if that invariant is ever broken.
  - Replaced the empty-family fallback in DecodeTagsConfig::new() with an explicit EmptyTagFamilies error, and removed the derived Default impl so an invalid empty config cannot be constructed implicitly.
  - Added Rust and Python tests covering the empty-config rejection.

  For suggestions 5-8:

  - Aligned From<&TagFamily> and From<&mut TagFamily> with the owned conversion so they also preserve the exact family as Custom(Arc<_>), removing the previous asymmetry.
  - Added documentation for FromStr, including an # Errors section.
  - Added a brief doc comment for builtin_tag.
  - Checked the decode_tags docs: the # Arguments section was already updated for the refactored signature in the earlier change, so there was no missing-doc fix left to make. I only made a small wording cleanup there so the follow-up is reflected in the diff.

AdityaShome:
I also checked other PRs and the same py314t job is failing there as well. I removed the earlier speculative path tweak from this PR to keep the scope focused on the apriltag review fixes. The remaining py314t failure looks like a separate CI/tooling issue in the maturin/free-threaded Python setup.

edgarriba (approved):

edgarriba:
Hi @AdityaShome — could you please rebase this branch onto latest `main` to resolve the merge conflicts, and look into the failing `py3.14t-linux` job ([log](https://github.com/kornia/kornia-rs/actions/runs/24290136133/job/70926033019))? Once those are green and @sidd-27's review comments are addressed, we can get this merged. Thanks!

edgarriba:
Test need to be fixed

edgarriba (approved):
Approved — landing per maintainer.

PR #704, opened 2026-02-09
docs: update algebra docs

cjpurackal:


edgarriba (commented):
nice start on the docs @cjpurackal — the lie group hierarchy diagram and the SU(2) → SO(3) double cover explanation is really well written. the quotient pattern discussion is a great addition.

keep going, this kind of mathematical context is exactly what we need in the codebase. would be great to also document the optimization module patterns (Factor trait, Variable, Problem) when you get to it.

PR #706, opened 2026-02-10
CI Pipeline Fix

zeel991:
## 📝 Description

Fixes the sccache server initialization issue causing all Python test workflows (py3.8-py3.14) to fail with "unexpected end of file" errors.

**Fixes/Relates to:** N/A (Infrastructure bug fix - no issue exists)

**Note:** This is a critical CI infrastructure fix that unblocks all PRs. The sccache server wasn't properly initialized before builds, causing connection failures.

---

## 🛠️ Changes Made

- [x] Added explicit sccache server initialization step in `.github/actions/setup-rust-cache/action.yml`
- [x] Server is stopped (if running) and restarted fresh before each build
- [x] Added verification step to confirm server is running
- [x] Fixed file formatting (added newline at EOF)

---

## 🧪 How Was This Tested?

- [x] **Local Testing:** Ran `pixi run -e py38 py-test` successfully with sccache enabled
  - sccache stats showed proper operation (229 compile requests, 152 cache misses, 178 MiB cache)
  - No connection errors
- [x] **Fork Testing:** Running CI in my fork to verify all Python test jobs pass
- [x] **Manual Verification:** Confirmed sccache server starts and responds to `--show-stats` command

**Test output from local run:**
```
Compile requests                    229
Cache hits rate                    0.00 %
Cache size                          178 MiB
Max cache size                        6 GiB
```

---

## 🕵️ AI Usage Disclosure
- [x] 🟢 **No AI used.**
- [ ] 🟡 **AI-assisted:** I used AI for boilerplate/refactoring but have manually reviewed and tested every line.
- [ ] 🔴 **AI-generated:** (Note: These PRs may be subject to stricter scrutiny or immediate closure if the logic is not explained).
---

## 🚦 Checklist

- [x] I have performed a **self-review** of my code
- [x] My code follows the existing style guidelines of this project
- [x] I have tested the fix locally and it works
- [x] This fix unblocks critical CI infrastructure

---

### Problem
All Python test workflows were failing with:
```
sccache: error: failed to get stats from server
sccache: caused by: Failed to send data to or receive data from server
sccache: caused by: io error: unexpected end of file
```

### Root Cause
The `mozilla-actions/sccache-action@v0.0.9` doesn't guarantee the sccache server is fully started before builds begin, leading to connection failures.

### Solution
Added explicit server management:
1. Stop any stale server instances
2. Start a fresh server
3. Verify it's running before proceeding

This ensures the server is ready before any Rust compilation occurs.

PR #710, opened 2026-02-14
ci: prevent sccache stats from failing CI jobs

cjpurackal:
`sccache --show-stats` fails with exit code 2 when the sccache server has already stopped or has a version mismatch, causing the entire job to be marked as failed even though all tests passed. Add continue-on-error and if: always() so stats are best-effort.

### DIFF
--- a/.github/workflows/python_test.yml
+++ b/.github/workflows/python_test.yml
@@ -74,5 +74,5 @@ jobs:
 
       - name: Show sccache stats
         if: always()
-        continue-on-error: true
-        run: sccache --show-stats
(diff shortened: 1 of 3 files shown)


cjpurackal:
https://github.com/kornia/kornia-rs/pull/706 this did not seem to fix the sccache issue. I've tried to address it here but some of the tests are timing out meaning that caching might still not be working. (It looks like a rate-limit issue ?). Need to investigate further.  For the time being, let's get this merged so that this isn't a blocker for other PRs.

Issue #750, opened 2026-02-24
[Bug]: bench_product1_float_kernel  fails for i8 due to integer overflow


namanguptagit:
### 🐛 Describe the bug

Running the benchmark suite for kornia-tensor-ops fails with a panic during the i8 execution of the bench_product1_float_kernel benchmark due to an integer overflow. The benchmark generates random i8 vectors and accumulates their product, which rapidly overflows the i8 capacity limit of 127.

### 🔄 Steps to Reproduce

```bash
1 .Checkout the repository.
2 .Run cargo bench -p kornia-tensor-ops (or pixi run cargo bench -p kornia-tensor-ops).
3 .Observe the benchmark runner panic during i8_size_8 or larger.
```

### 💻 Minimal Code Example

```rust
use kornia_tensor_ops::kernels::dot_product1_kernel;
use rand::Rng;

// Inside bench_product1_float_kernel:
let mut rng = rand::rng();
let size = 8;
let a: Vec<i8> = (0..size).map(|_| rng.random::<i8>()).collect();
let b: Vec<i8> = (0..size).map(|_| rng.random::<i8>()).collect();

// Multiplying and accumulating 8 pairs of random i8 values overflows `i8` capacity
let result = dot_product1_kernel(&a, &b).unwrap();
```

### ✅ Expected behavior

The benchmarks should successfully execute for all data types without panicking.

### ❌ Actual behavior

The benchmark panics with an arithmetic overflow during the i8 runs:

```
thread 'main' panicked at crates/kornia-tensor-ops/src/kernels.rs:41:64:
attempt to multiply with overflow
```

### 🔧 Environment

```shell
- kornia-rs version: 0.1.11
- Rust version: 1.82 (or your local version)
- Cargo version: 1.82 (or your local version)
- OS: macOS
- Target architecture: aarch64
- Python version: (N/A)
```

### 📝 Additional context

The root cause is in crates/kornia-tensor-ops/src/kernels.rs where dot_product1_kernel
 accumulates into the same type T. When generating random i8 values in the benchmark, rand::rng().random::<i8>() returns values between -128 and 127. A single multiplication of two randomly generated numbers (e.g., 50 * 50 = 2500) exceeds the i8 maximum bounds, leading to the panic.

### 🤝 Contribution Intent

- [x] I plan to submit a PR to fix this bug
- [ ] I'm reporting this bug but not planning to fix it

sidd-27:
[...]

PR #758, opened 2026-02-25
fix: remove i8 benchmark case to prevent integer overflow

Avi-47:
## Description

This PR fixes issue #750 by removing the i8 test case from the `bench_product1_float_kernel` benchmark.

The benchmark was panicking when running with i8 data types because multiplying two random i8 values (which range from -128 to 127) can easily exceed the i8 maximum value of 127, causing an integer overflow. For example, 50 * 50 = 2500 which is way beyond what i8 can hold.

After discussing in the issue, we agreed that removing the i8 case is the best approach since dot product on i8 vectors isn't a realistic use case for image processing anyway. The kernel itself already handles overflow gracefully with saturating arithmetic, so there's no need to change the core implementation.

## Changes

Removed the i8 benchmark loop (lines 20-27) from `crates/kornia-tensor-ops/benches/bench_kernels.rs`.

## Testing

Ran `cargo bench -p kornia-tensor-ops --bench bench_kernels` and verified all f32 benchmarks run successfully without any panics.

## Additional Notes

- I'm assigned to issue #750
- The maintainer (sidd-27) approved removing the i8 case in the issue discussion
- This is a minimal, targeted fix that only removes the problematic benchmark case

### DIFF
--- a/crates/kornia-tensor-ops/benches/bench_kernels.rs
+++ b/crates/kornia-tensor-ops/benches/bench_kernels.rs
@@ -1,7 +1,35 @@
+//! Benchmark suite for tensor operation kernels.
+//!
+//! # Limitations & Design Decisions
+//!
+//! - **Integer types (i8, i16, etc.)**: Only floating-point (f32) benchmarks are included.
+//!   Integer dot product operations can easily overflow during multiplication (e.g., 50 * 50 = 2500
+//!   exceeds i8's max value of 127). While the kernel uses saturating arithmetic to handle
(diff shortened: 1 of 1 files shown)


sidd-27:
Could you also add comments there which help future devs make decisions easier? document the current limitations etc

Avi-47:
@sidd-27 I have added comments. Are those ok?

Avi-47:
@sidd-27 my PR validation failed because it requires a linked issue. I have added "Fixes #750" to the commit message. can you please check

PR #772, opened 2026-02-27
Fix/Eliminate redundant meshgrid allocation in warp_affine and warp_perspective

namanguptagit:
Eliminate redundant meshgrid allocation in warp_affine and warp_perspective
## 📝 Description
This PR completely eliminates the massive, redundant memory allocations of intermediate `map_x` and `map_y` meshgrids in both warp_affine(kornia-rs/crates/kornia-imgproc/src/warp/affine.rs) and warp_perspective (kornia-rs/crates/kornia-imgproc/src/warp/perspective.rs) operations.
Previously, the code would allocate two full-image sized `Tensor2<f32>` arrays just to hold the reverse-warped coordinates, leading to enormous memory bloat and cache locality destruction, especially on high-resolution images like 1080p and 4K.
Now, an allocation-free parallel iterator (par_iter_rows_spatial_mapping(kornia-rs/crates/kornia-imgproc/src/parallel.rs)) is utilized to compute the spatial mapping coordinates dynamically "on-the-fly" for each pixel.

**⚠️ Issue Link Required**: This PR must be linked to an approved and assigned issue. See [Contributing Guide](CONTRIBUTING.md#pull-request) for details.
https://github.com/kornia/kornia-rs/issues/771
**Fixes/Relates to:** # 771

**Important**:
- Ensure you are assigned to the linked issue before submitting this PR
- This PR should strictly implement what the linked issue describes
- Do not include changes beyond the scope of the linked issue

---

## 🛠️ Changes Made
- [x] Added [par_iter_rows_spatial_mapping] to [kornia-imgproc/src/parallel.rs], avoiding the need for pre-allocated coordinate tensors by accepting a mapping closure.
- [x] Refactored [warp_affine]to use the new iterator, directly passing the inverted affine transform logic inline and dropping `meshgrid_from_fn`.
- [x] Refactored [warp_perspective]to use the new iterator, passing the inverse perspective matrix logic inline and dropping `meshgrid_from_fn`.
- [x] Removed unused `meshgrid_from_fn` imports due to the above.

---

## 🧪 How Was This Tested?
- [x] **Unit Tests:** Verified that `cargo test -p kornia-imgproc warp` fully passes. The exact mathematical output and pixel correctness of all affine/perspective transformations remain identical.
[...]

PR #777, opened 2026-02-28
fix: replace interpolation panics with errors, remove wasted clone in morphology

### LINKED ISSUE #770: [Bug]:  Public API panics for Lanczos/Bicubic interpolation + wasted clone in morphological open/close
### 🐛 Describe the bug

Two issues I found while going through kornia-imgproc:

1) InterpolationMode has Lanczos and Bicubic variants, but interpolate_pixel() in interpolation/interpolate.rs:42-47 just does unimplemented!() for both. So if you pass InterpolationMode::Bicubic to resize_native(), warp_perspective(), warp_affine(), or remap() — it compiles fine but panics at runtime. No Result::Err, just a crash. The annoying part is resize_fast_rgb actually handles Lanczos/Bicubic fine through fast_image_resize (resize.rs:225-228), so the same enum works in one function and panics in another.

2) In morphology/ops.rs, both open() (line 221) and close() (line 254) do let mut temp_img = src.clone() — full allocation + pixel copy of the source — and then immediately pass temp_img to erode()/dilate() which overwrites every single pixel. The cloned data is never read. On a 1920×1080×3 u8 image that's ~6MB wasted per call.

### 🔄 Steps to Reproduce

```bash
Panic:
1. Call resize_native() with InterpolationMode::Bicubic
2. Crashes with "not yet implemented: Bicubic interpolation is not yet implemented"
3. Same thing through warp_perspective(), warp_affine(), remap()

# Allocation:
1. Call morphology::open() on a large image
2. Profile with heaptrack or similar — you'll see a full-size clone that gets immediately overwritten
```

### 💻 Minimal Code Example

```rust
use kornia_image::{Image, ImageSize};
use kornia_image::allocator::CpuAllocator;
use kornia_imgproc::resize::resize_native;
use kornia_imgproc::interpolation::InterpolationMode;

fn main() {
    let src = Image::<f32, 3, _>::from_size_val(
        ImageSize { width: 100, height: 100 }, 0.5, CpuAllocator
    ).unwrap();
    let mut dst = Image::<f32, 3, _>::from_size_val(
        ImageSize { width: 50, height: 50 }, 0.0, CpuAllocator
    ).unwrap();

    // compiles, panics at runtime
    resize_native(&src, &mut dst, InterpolationMode::Bicubic).unwrap();
}
```

### ✅ Expected behavior

For the panic — resize_native and the other 3 functions should return an error when you pass an unsupported interpolation mode, not crash.

For the clone — just allocate a zero-initialized temp instead of cloning source. Image::from_size_val(src.size(), T::default(), CpuAllocator)? — the data gets overwritten anyway.

### ❌ Actual behavior

interpolate_pixel hits unimplemented!() at interpolate.rs:42-47 and panics. This propagates through resize_native, warp_perspective, warp_affine, and remap.

open() at line 221 and close() at line 254 both src.clone() allocating width×height×channels×sizeof(T) bytes that are immediately overwritten by the next call.

### 🔧 Environment

```shell
- kornia-rs version: 0.1.11
- Rust version (`rustc -V`): rustc 1.82.0
- Cargo version (`cargo -V`): cargo 1.82.0
- OS (e.g., Linux, macOS, Windows): macOS
- Target architecture (if cross-compiling): arm64
- Python version (if using Python bindings): N/A
```

### 📝 Additional context

**Wasted allocation per call in open()/close():**

| Image Size | Type | Wasted Clone | At 30fps |
|---|---|---|---|
| 640×480×3 | u8 | 0.92 MB | 27.6 MB/s |
| 1280×720×3 | u8 | 2.76 MB | 82.9 MB/s |
| 1920×1080×3 | u8 | 6.22 MB | 186.6 MB/s |
| 1920×1080×3 | f32 | 24.88 MB | 746.5 MB/s |
| 3840×2160×3 | u8 | 24.88 MB | 746.5 MB/s |

In a standard morphological noise removal pipeline (open followed by close), you pay this cost twice — so double those numbers. At 1080p f32 that's nearly 1.5 GB/s of garbage allocation just from the unnecessary clone.

Also worth noting: there's no benchmark for morphology ops at all right now. kornia-imgproc/benches/ has benchmarks for color, crop, distance_transform, features, filters, flip, gradient, metrics, pyramid, resize, warp — but nothing for morphology.

**4 public APIs hit by the panic:**

| Function | File |
|---|---|
| resize_native | resize.rs:59 |
| warp_perspective | warp/perspective.rs:101 |
| warp_affine | warp/affine.rs:135 |
| remap | interpolation/remap.rs:27 |

For comparison — OpenCV returns cv::Error with CV_HAL_ERROR_NOT_IMPLEMENTED for unsupported modes. It never panics. Meanwhile resize_fast_rgb in the same module handles Lanczos/Bicubic fine:
```rust
// resize.rs:225-228 — these work
InterpolationMode::Lanczos => fr::ResizeAlg::Convolution(fr::FilterType::Lanczos3),
InterpolationMode::Bicubic => fr::ResizeAlg::Convolution(fr::FilterType::CatmullRom),
```
So user sees the same enum, tries it in a different function, gets a crash.

**Implementation approach:**

Can't just change interpolate_pixel to return Result<f32, ImageError> because it's called inside Rayon parallel closures (par_iter_rows_resample) which can't propagate Result. The right approach:

1. Add validate_interpolation() that returns Err(ImageError::UnsupportedInterpolation) for unsupported modes
2. Each of the 4 public APIs calls validate_interpolation(interpolation)? before entering the parallel loop — error gets caught early with a proper Result before any allocation or computation
3. interpolate_pixel keeps its f32 return, unimplemented!() becomes unreachable!()
4. Replace src.clone() with Image::from_size_val(src.size(), T::default(), CpuAllocator)? in both open() and close()
5. Add morphology benchmarks (dilate, erode, open, close)
6. Add tests for each affected API verifying unsupported modes return Err

When Bicubic/Lanczos eventually gets implemented in interpolate_pixel, the validation just needs to accept the new modes — the error propagation path is already wired up.

The morphology fix is a one-liner per function:
```diff
// open() — line 221
- let mut temp_img = src.clone();
+ let mut temp_img = Image::from_size_val(src.size(), T::default(), CpuAllocator)?;

// close() — line 254
- let mut temp_img = src.clone();
+ let mut temp_img = Image::from_size_val(src.size(), T::default(), CpuAllocator)?;
```

### 🤝 Contribution Intent

shreed27:
## Summary

Fixes #770

- `interpolate_pixel()` used `unimplemented!()` for Lanczos/Bicubic which caused runtime panics through 4 public APIs (`resize_native`, `warp_perspective`, `warp_affine`, `remap`). Added `validate_interpolation()` that returns `Err(ImageError::UnsupportedInterpolation)` before entering parallel dispatch, so users get a proper error instead of a crash.

- `morphology::open()` and `close()` called `src.clone()` for an intermediate buffer that immediately gets overwritten by `erode()`/`dilate()`. Replaced with `Image::from_size_val(src.size(), T::default(), CpuAllocator)?` to skip the pointless copy.

- Added morphology benchmarks (`dilate`, `erode`, `open`, `close`) — there were none before.

## Changes

| File | What |
|---|---|
| `kornia-image/src/error.rs` | Add `UnsupportedInterpolation(String)` variant |
| `interpolation/interpolate.rs` | Add `validate_interpolation()`, replace `unimplemented!()` with `unreachable!()` |
| `interpolation/mod.rs` | Export `validate_interpolation` |
| `resize.rs` | Early validation before parallel loop |
| `warp/perspective.rs` | Early validation before parallel loop |
| `warp/affine.rs` | Early validation before parallel loop |
| `interpolation/remap.rs` | Early validation before parallel loop |
| `morphology/ops.rs` | Replace `src.clone()` with zero-init in `open()` and `close()` |
| `benches/bench_morphology.rs` | New: dilate/erode/open/close benchmarks |
| `Cargo.toml` | Register morphology bench |

## Test plan

- [x] All 139 lib tests pass (`cargo test -p kornia-imgproc --lib`)
- [x] All 41 doc tests pass (`cargo test -p kornia-imgproc --doc`)
- [x] Clippy clean (`cargo clippy -p kornia-imgproc -p kornia-image --all-targets`)
- [x] 4 new tests verify unsupported interpolation returns `Err` (one per affected API)
- [x] Existing morphology tests (`test_open_remove_noise`, `test_close_fill_hole`) pass with the `from_size_val` change

### DIFF
--- a/crates/kornia-imgproc/benches/bench_resize.rs
+++ b/crates/kornia-imgproc/benches/bench_resize.rs
@@ -75,7 +75,7 @@ fn resize_ndarray_zip(src: &Image<f32, 3, CpuAllocator>, dst: &mut Image<f32, 3,
 
             // write the pixel values to the output image
             for (k, pixel) in pixels.enumerate() {
-                out[k] = pixel;
+                out[k] = pixel.unwrap_or(0.0);
             }
         });
 }


crates/kornia-imgproc/src/interpolation/interpolate.rs
```diff
@@ -39,11 +55,9 @@ pub fn interpolate_pixel<const C: usize, A: ImageAllocator>(
...
-            unimplemented!("Lanczos interpolation is not yet implemented")
-        }
-        InterpolationMode::Bicubic => {
-            unimplemented!("Bicubic interpolation is not yet implemented")
-        }
+        _ => unreachable!(
+            "unsupported interpolation mode {:?} — call validate_interpolation() first",
+            interpolation
+        ),
     }
```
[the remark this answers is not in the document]
shreed27: fixed

crates/kornia-imgproc/src/interpolation/remap.rs
```diff
@@ -60,6 +62,40 @@ mod tests {
...
+            Tensor2::from_shape_vec([2, 2], vec![0.0, 0.0, 1.0, 1.0], CpuAllocator).unwrap();
+        let mut dst = Image::<_, 1, _>::from_size_val(
+            ImageSize {
+                width: 2,
+                height: 2,
+            },
+            0.0,
+            CpuAllocator,
+        )
+        .unwrap();
```
edgarriba: no unwraps

crates/kornia-imgproc/src/interpolation/interpolate.rs
```diff
@@ -39,11 +55,9 @@ pub fn interpolate_pixel<const C: usize, A: ImageAllocator>(
     match interpolation {
         InterpolationMode::Bilinear => bilinear_interpolation(image, u, v, c),
         InterpolationMode::Nearest => nearest_neighbor_interpolation(image, u, v, c),
-        InterpolationMode::Lanczos => {
-            unimplemented!("Lanczos interpolation is not yet implemented")
-        }
-        InterpolationMode::Bicubic => {
-            unimplemented!("Bicubic interpolation is not yet implemented")
-        }
+        _ => unreachable!(
```
edgarriba: use the error you created here

shreed27: updated.

crates/kornia-imgproc/src/interpolation/remap.rs
shreed27: updated.

shreed27:
@edgarriba addressed both review comments:

- `interpolate_pixel` now returns `Result<f32, ImageError>` with `UnsupportedInterpolation` error instead of `unreachable!()`
- all test `.unwrap()` calls replaced with `?` operator

also resolved the merge conflicts from #772 — dropped the meshgrid allocation path and updated the `par_iter_rows_spatial_mapping` calls to use `.unwrap_or(0.0)` on the result (safe since `validate_interpolation` runs before entering the loop).

crates/kornia-imgproc/benches/bench_morphology.rs
dhruvkjain, line 1: provide benchmark results
also benchmark kornia's implementation against opencv's

edgarriba:
LGTM — @sidd-27 can you merge? Thanks!

edgarriba:
LGTM — @sidd-27 can you merge? Thanks!

PR #831, opened 2026-03-19
fix(io): return Image<u8, 1> from decode_gray8

### LINKED ISSUE #830: [Bug]: `decode_gray8` returns `Image<u8, 3>` instead of `Image<u8, 1>`
### 🐛 Describe the bug

`decode_gray8()` (line 166) in `crates/kornia-io/src/jpegturbo.rs` returns `Image<u8, 3>` but should return `Image<u8, 1>`, as it decodes with `PixelFormat::GRAY` (1 byte per pixel) and the docstring already says it should return `Image<u8, 1>`. 
Meanwhile the pitch (line 188) is hardcoded to `3 * image_size.width` instead of `C * image_size.width`, which works for `decode_rgb8` but breaks `decode_gray8`.

### 🔄 Steps to Reproduce

```bash
1. Create a JpegTurboDecoder and call `decode_gray8` with any valid JPEG
2. Check num_channels() on the result, which returns 3 instead of 1
```

### 💻 Minimal Code Example

```rust
use kornia_io::jpegturbo::JpegTurboDecoder;

fn main() -> Result<(), Box<dyn std::error::Error>> {
    let decoder = JpegTurboDecoder::new()?;
    let jpeg_data = std::fs::read("path/to/dog.jpeg")?;
    let image = decoder.decode_gray8(&jpeg_data)?;
    // Returns 3, should be 1
    println!("channels: {}", image.num_channels());
    Ok(())
}
```

### ✅ Expected behavior

`decode_gray8` should return `Image<u8, 1>` with 1 channel and correct pixel layout, consistent with its docstring and `PixelFormat::GRAY`.

### ❌ Actual behavior

`decode_gray8` returns `Image<u8, 3>` with 3 channels. The buffer is 3x too large and the pitch is wrong, resulting in incorrect grayscale data.

### 🔧 Environment

```shell
- kornia-rs version: 0.1.11 (workspace)
- Rust version (`rustc -V`): rustc 1.94.0 (4a4ef493e 2026-03-02)
- Cargo version (`cargo -V`): cargo 1.94.0 (85eff7c80 2026-01-15)
- OS (e.g., Linux, macOS, Windows): Linux
```

### 📝 Additional context

The fix should be really simple by changing the return type of `decode_gray8` from `Image<u8, 3>` to `Image<u8, 1>` and change `pitch: 3 * image_size.width` to `pitch: C * image_size.width` in `decode<C>`.  
Happy to submit a PR if approved!

### 🤝 Contribution Intent

mvanhorn:
## Summary

`decode_gray8()` returned `Image<u8, 3>` instead of `Image<u8, 1>` because the pitch in the generic `decode()` helper was hardcoded to `3 * width` rather than `C * width`. This caused incorrect behavior for grayscale decoding.

## Changes

- Fixed return type of `decode_gray8` from `Image<u8, 3>` to `Image<u8, 1>`
- Fixed pitch calculation in `decode()` to use `C * image_size.width` instead of `3 * image_size.width`

Fixes #830

This contribution was developed with AI assistance (Claude Code).

### DIFF
--- a/crates/kornia-io/src/jpegturbo.rs
+++ b/crates/kornia-io/src/jpegturbo.rs
@@ -85,6 +85,36 @@ impl JpegTurboEncoder {
             .compress_to_vec(buf)?)
     }
 
+    /// Encodes the given Gray/Mono8 image into a JPEG image.
+    ///
+    /// # Arguments
+    ///
+    /// * `image` - The image to encode.
+    ///
+    /// # Returns
+    ///
+    /// The encoded data as `Vec<u8>`.
+    pub fn encode_gray8<A: ImageAllocator>(
+        &self,
+        image: &Image<u8, 1, A>,
+    ) -> Result<Vec<u8>, JpegTurboError> {
+        let image_data = image.as_slice();
+
+        let buf = turbojpeg::Image {
+            pixels: image_data,
+            width: image.width(),
+            pitch: image.width(),
+            height: image.height(),
+            format: turbojpeg::PixelFormat::GRAY,
+        };
+
+        Ok(self
+            .0
+            .lock()
+            .map_err(|_| JpegTurboError::MutexPoisoned)?
+            .compress_to_vec(buf)?)
+    }
+
     /// Sets the quality of the encoder.
     ///
     /// # Arguments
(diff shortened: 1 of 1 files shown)


sidd-27:
could you implement the same fix for the encoder too? i think that has the same issue

mvanhorn:
Added `encode_gray8` to `JpegTurboEncoder` in b725c66 - mirrors the decoder-side fix with `PixelFormat::GRAY` and pitch of `width` (1 channel).

PR #852, opened 2026-03-26
fix(python): include Cargo.lock in PyPI source distribution

### LINKED ISSUE #827: [Bug]: The soorce tarball on PyPI doesn't contain the Cargo.lock file
### 🐛 Describe the bug

[This tarball](https://files.pythonhosted.org/packages/ab/17/8b3518ece01512a575b18f86b346879793d3dea264b314796bbd44d42e11/kornia_rs-0.1.10.tar.gz) has no Cargo.lock

### 🔄 Steps to Reproduce

```bash
n/a
```

### 💻 Minimal Code Example

```rust

```

### ✅ Expected behavior

n/a

### ❌ Actual behavior

n/a

### 🔧 Environment

```shell
- kornia-rs version: 0.1.10
```

### 📝 Additional context

_No response_

### 🤝 Contribution Intent

- [ ] I plan to submit a PR to fix this bug

Moka-4444:
Fixes #827 
## Problem
The PyPI source tarball for `kornia_rs` was missing the `Cargo.lock` file, making builds from the sdist non-reproducible.

## Solution
Added `[tool.maturin]` configuration to `kornia-py/pyproject.toml`:

- `locked = true` -- Ensures `cargo build` uses `--locked` during sdist builds, preventing dependency resolution drift.
- - `sdist-include = ["../Cargo.lock"]` -- Explicitly includes the workspace-level `Cargo.lock` in the source distribution tarball.
## Verification
After this change, building an sdist with `maturin sdist` will include the `Cargo.lock` file.

### DIFF
--- a/kornia-py/pyproject.toml
+++ b/kornia-py/pyproject.toml
@@ -62,5 +62,13 @@ dev = [
 requires = ["maturin>=1,<2"]
 build-backend = "maturin"
 
+[tool.maturin]
+# Include the workspace Cargo.lock in the source distribution (sdist)
+# to ensure reproducible builds from the PyPI tarball.
+# See: https://github.com/kornia/kornia-rs/issues/827
+locked = true
+sdist-include = ["../Cargo.lock"]
+
+
 [tool.uv]
 extra-index-url = ["https://download.pytorch.org/whl/cpu"]


sidd-27:
https://github.com/kornia/kornia-rs/issues/827#issuecomment-4080131184 

seems like this issue was already solved a while back

PR #877, opened 2026-04-20
feat(imgproc): SIMD backend PoC — AVX2 grayscale + threshold (up to 21× faster)

sidd-27:
## Summary

Introduces a new `simd` module in `kornia-imgproc` (gated by a default-on `simd` cargo feature) as the scaffolding for a portable-SIMD backend. The PoC ports two representative ops with hand-written AVX2 intrinsics tuned against LLVM's auto-vectorization baseline:

- **`simd::gray_from_rgb_u8`** — RGB8→Gray8 via `pshufb`+`perm2i128` 3-way deinterleave, **32 pixels/iter** AVX2, with SSSE3 (16 px/iter) and `wide` fallbacks.
- **`simd::threshold_binary`** — u8 elementwise threshold via `vpminub`/`vpcmpeqb`/`vpandn`, **4× unrolled** AVX2 (128 bytes/iter).

Runtime dispatch via `is_x86_feature_detected!` picks the best available path per call; public API shape matches the existing scalar functions. The scalar implementations in `color::` and `threshold::` are untouched — this PoC exposes new entry points so A/B benchmarking is trivial and the scalar path stays as the reference.

## How I designed the kernels

Initial attempts using just the `wide` crate (portable SIMD) were actually **slower** than scalar, because `-C target-cpu=native` lets LLVM auto-vectorize the scalar `chunks_exact(3).zip(iter_mut)` pattern to full AVX2 already. I used `cargo asm` to dump what LLVM was emitting:

- Scalar grayscale → AVX2 `vpmullw ymm / vpaddw ymm / vpsrlw / vpackuswb / vpermq 0xD8`, 32 px/iter.
- Scalar threshold → **4× unrolled AVX2** with the `vpminub + vpcmpeqb + vpandn` trick (avoids the signed/unsigned compare issue entirely).

Hand-written intrinsics only win when they match LLVM's vector width and elide its scalar tail/bounds-check paths, so the final kernels do exactly that.

### Single-threaded (`RAYON_NUM_THREADS=1`) — pure kernel speedup

| Op | Size | Scalar | SIMD | **Speedup** |
|---|---|---|---|---|
| GrayU8 | 256×224 | 31.0 µs | 21.5 µs | **1.44×** |
| GrayU8 | 512×448 | 81.1 µs | 59.3 µs | **1.37×** |
| GrayU8 | 1024×896 | 362 µs | 298 µs | **1.21×** |
| ThresholdU8 | 256×224 | 53.8 µs | 3.71 µs | **14.5×** |
| ThresholdU8 | 512×448 | 229 µs | 15.0 µs | **15.2×** |
| ThresholdU8 | 1024×896 | 763 µs | 36.3 µs | **21.0×** |

### Multi-threaded (rayon, 8 threads) — end-to-end

| Op | Size | Scalar | SIMD | **Speedup** |
|---|---|---|---|---|
| GrayU8 | 256×224 | 28.4 µs | 34.5 µs | 0.82× |
| GrayU8 | 512×448 | 43.8 µs | 54.5 µs | 0.80× |
| GrayU8 | 1024×896 | 122 µs | 98.4 µs | **1.25×** |
| ThresholdU8 | 256×224 | 34.4 µs | 39.6 µs | 0.87× |
| ThresholdU8 | 512×448 | 74.2 µs | 57.2 µs | **1.30×** |
| ThresholdU8 | 1024×896 | 169 µs | 76.2 µs | **2.22×** |

### What the numbers mean

- **Single-thread**: SIMD always wins; threshold is a massive **15–21× speedup** because the scalar path (through generic + rayon closures) doesn't autovec as aggressively as the pure scalar reference.
- **Multi-thread**: two different ceilings kick in:
  - Small images (256×224): rayon's per-row task scheduling overhead (~5–10 µs fixed cost) dominates both scalar and SIMD; the faster SIMD kernel finishes sooner and just waits on the scheduler.
  - Large images (1024×896): DRAM bandwidth becomes the wall; 8 cores saturate the memory bus around the same throughput regardless of per-thread compute speed. We still see 1.25–2.2× wins because threshold's scalar path wasn't memory-bound single-thread.
- A coarser rayon chunking strategy (multi-row batches) or bypassing rayon entirely at small image sizes would almost certainly recover the single-thread wins in multi-thread mode — flagged as follow-up work.

## Correctness

8 unit tests in the new `simd` module cover:
- SIMD vs scalar byte-exact match on aligned widths (multiples of 16 and 32).
- Scalar-tail correctness at widths that aren't multiples of the SIMD block (17, 19, 33).
- Boundary semantics for threshold (strict `>`).
- The existing u8 regression fixture (`[0, 128, 255, 128, 0, 128]` → `[103, 53]`).
- Size-mismatch error path.

All 8 pass. No existing tests are modified.

## Files

**New**:
- `crates/kornia-imgproc/src/simd/mod.rs` — module wiring, feature gate, design notes
- `crates/kornia-imgproc/src/simd/color.rs` — public entry + dispatch + `wide` fallback + tests
- `crates/kornia-imgproc/src/simd/color_x86.rs` — SSSE3 & AVX2 u8 intrinsic kernels
- `crates/kornia-imgproc/src/simd/threshold.rs` — public entry + dispatch + `wide` fallback + tests
- `crates/kornia-imgproc/src/simd/threshold_x86.rs` — AVX2 threshold kernel (4× unroll)
- `crates/kornia-imgproc/benches/bench_threshold.rs` — scalar vs SIMD threshold

**Modified**:
- `crates/kornia-imgproc/Cargo.toml` — add `[features] default = [\"simd\"], simd = []`; register `bench_threshold`
- `crates/kornia-imgproc/src/lib.rs` — `#[cfg(feature = \"simd\")] pub mod simd;`
- `crates/kornia-imgproc/benches/bench_color.rs` — trim to just scalar_u8 vs simd_u8 (dropped external-crate baselines unrelated to the PoC)

No changes to `Image`, scalar ops, or any public API beyond adding the `simd` module.

## Scope / non-goals

- **f32 grayscale** was prototyped and **dropped** from the PoC: AVX2 `vgatherdps`-based gather was ~2× slower than autovec on the test hardware (gather is micro-coded on Zen/Alder Lake). A shuffle-based f32 deinterleave is future work and not needed to demonstrate the backend pattern.
- **No runtime multi-versioned dispatch** yet (one `is_x86_feature_detected!` check per row is cheap and sufficient for a PoC).
- **Not yet ported**: normalize, flip, separable filter, histogram, warp, morphology — this PR is the scaffolding they plug into.
- **No ARM NEON intrinsic path** — the `wide` fallback covers NEON via `wide`'s SIMD128/NEON backends; a native `std::arch::aarch64` path can come later for ops where `wide` isn't ideal.

## Test plan

- [ ] `cargo check -p kornia-imgproc` (default features)
- [ ] `cargo check -p kornia-imgproc --no-default-features`
- [ ] `cargo clippy -p kornia-imgproc --lib --tests` — clean
- [ ] `cargo test -p kornia-imgproc simd` — 8/8 pass
- [ ] `cargo test -p kornia-imgproc color::tests` — existing tests still pass
- [ ] `RUSTFLAGS=\"-C target-cpu=native\" cargo bench -p kornia-imgproc --bench bench_color --bench bench_threshold` — reproduces the speedups above
- [ ] Verify on non-AVX2 hardware that the SSSE3 and `wide` fallbacks are exercised (manual, or via `RUSTFLAGS=\"-C target-feature=-avx2,-ssse3\"`)

### DIFF
--- a/crates/kornia-imgproc/Cargo.toml
+++ b/crates/kornia-imgproc/Cargo.toml
@@ -60,6 +60,10 @@ name = "bench_morphology"
 
 [[bench]]
 harness = false
+name = "bench_threshold"
+
+[[bench]]
+harness = false
 name = "bench_distance_transform"
 required-features = ["opencv_bench"]
 
@@ -94,4 +98,6 @@ ndarray = { version = "0.15", features = ["rayon"] }
 rand = { workspace = true }
 
 [features]
+default = ["simd"]
 opencv_bench = ["dep:opencv"]
+simd = []
--- a/crates/kornia-imgproc/benches/bench_threshold.rs
+++ b/crates/kornia-imgproc/benches/bench_threshold.rs
@@ -1,17 +1,11 @@
-//! Focused PoC benchmark: scalar vs SIMD grayscale on u8 RGB.
-//!
-//! Run with `RAYON_NUM_THREADS=1` to see the pure per-row kernel speedup; run
-//! without for realistic end-to-end (which at large sizes is memory-bandwidth
-//! bound, so absolute speedup shrinks).
-
 use criterion::{criterion_group, criterion_main, BenchmarkId, Criterion};
 use kornia_image::Image;
-use kornia_imgproc::color::gray_from_rgb_u8;
 use kornia_imgproc::simd;
+use kornia_imgproc::threshold::threshold_binary as scalar_threshold_binary;
 use kornia_tensor::CpuAllocator;
 
-fn bench_grayscale_u8(c: &mut Criterion) {
-    let mut group = c.benchmark_group("GrayU8");
+fn bench_threshold(c: &mut Criterion) {
+    let mut group = c.benchmark_group("ThresholdU8");
     group.warm_up_time(std::time::Duration::from_millis(300));
     group.measurement_time(std::time::Duration::from_secs(1));
     group.sample_size(25);
@@ -20,14 +14,14 @@ fn bench_grayscale_u8(c: &mut Criterion) {
         group.throughput(criterion::Throughput::Elements((*w * *h) as u64));
         let ps = format!("{w}x{h}");
         let size = [*w, *h].into();
-        let data: Vec<u8> = (0..w * h * 3).map(|i| (i * 31 % 251) as u8).collect();
-        let src = Image::<u8, 3, _>::new(size, data, CpuAllocator).unwrap();
+        let data: Vec<u8> = (0..w * h).map(|i| (i * 37 % 256) as u8).collect();
+        let src = Image::<u8, 1, _>::new(size, data, CpuAllocator).unwrap();
         let dst = Image::<u8, 1, _>::from_size_val(size, 0, CpuAllocator).unwrap();
 
         group.bench_with_input(BenchmarkId::new("scalar", &ps), &(&src, &dst), |b, i| {
             let (s, mut d) = (i.0, i.1.clone());
             b.iter(|| {
-                gray_from_rgb_u8(s, &mut d).unwrap();
+                scalar_threshold_binary(s, &mut d, 100u8, 255u8).unwrap();
                 std::hint::black_box(&mut d);
             })
         });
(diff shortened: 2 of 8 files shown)


edgarriba:
Can you provide benchmark against a the fastest library? Opencv/pil/vpn

PR #887, opened 2026-04-22
perf(imgproc,3d,py): cross-arch ORB + two-view pose + Python feature surface (NEON / AVX2)

edgarriba:
End-to-end CV stack on aarch64 (NEON) + x86_64 (AVX2): pure-CPU ORB feature pipeline, two-view pose estimation that beats OpenCV's USAC variants, and a numpy-zero-copy Python surface — all behind a runtime CPU-feature dispatcher so a single binary serves both arches.

## What lands

**Features (ORB, FAST, matcher, homography)**
- NEON FAST-9 detector — single-pass `uint8x16_t` chain-counter arc test (`features/fast.rs`).
- AVX2 FAST-9 mirror — same algorithm on `__m128i`, `_mm_subs_epu8` saturating arc check.
- Full ORB pipeline: FAST → Harris → intensity-centroid orientation → BRIEF, with per-octave rayon parallelism, fused detect+extract, and OpenCV-aligned `DistributeOctTree` keypoint distribution.
- LO-RANSAC homography (DLT + inlier-refit) and a rayon-parallel brute-force 32-byte binary-descriptor matcher (NEON `vcntq_u8` / AVX2 `popcnt` Hamming).

**Two-view pose**
- `find_fundamental` — 8-point with Householder null-vector + LO-RANSAC refit.
- 5-point Nistér essential solver (null-space + cubic constraints + Gauss-Jordan).
- LM (R, t) post-refinement on Sampson cost.
- Cheirality ambiguity guard for near-degenerate motion.
- AᵀA streaming RANSAC scoring + NEON / AVX2 score_inliers paths + adaptive iteration counts.

**Cross-arch SIMD infra**
- All hot kernels in `imgproc` (gray, crop, fast, match, orb/extractor, filter/ops, flip, resize/{fused,kernels}, warp/{common,kernels}, normalize) and `kornia-3d` two-view paths now have NEON + AVX2 + scalar variants behind a `cpu_features()` dispatcher.
- Atomic feature-detection load is hoisted out of rayon row closures so the probe runs once per call, not per row.

**Python surface (zero-copy numpy)**
- `orb_detect_and_compute`, `fast_detect`, `match_descriptors`, `find_homography`, `find_fundamental`, `recover_pose_two_view`.
- All benches consolidated under `kornia-py/benchmarks/`.

## Performance on Jetson Orin (ms; `taskset -c 0-5`, 200 iters × 5 rounds)

| Op | Size | kornia | OpenCV | VPI CUDA |
|---|---|---:|---:|---:|
| FAST-9 | 640 | **0.76** | 2.19 | 7.06 |
| FAST-9 | 1080p | **1.12** | 1.38 | 6.86 |
| ORB | 640 | **2.66** | 10.00 | 7.52 |
| ORB | 1080p | **10.65** | 44.23 | 11.77 |

ORB beats OpenCV by 3.5–4.2× and is in the same ballpark as VPI CUDA at 1080p (10.65 ms vs 11.77 ms) with zero GPU/device-transfer overhead. The sub-ms FAST kernel beats VPI CUDA 6–9× because CUDA launch cost dominates at that scale.

Two-view pose on EuRoC MH_01 beats every OpenCV USAC variant on accuracy (see `kornia-py/benchmarks/bench_two_view_pose.py`).

Quality (keypoint overlap vs OpenCV within 3 px on `dog.jpeg`): kornia 75%, VPI 66%.

## Test plan
- [ ] `cargo test --workspace`
- [ ] `cargo clippy --workspace --no-deps --all-targets --features ci -- -D warnings`
- [ ] `pytest kornia-py/tests/`
- [ ] `python kornia-py/benchmarks/bench_orb.py` (and `bench_fast.py`, `bench_two_view_pose.py`)


edgarriba:
@cjpurackal @ducha-aiki — this is now ready for review. Could you take a look at the reported numbers (criterion benches in `crates/kornia-imgproc/benches/`, Python benches under `kornia-py/benchmarks/`) and sanity-check that the speedups land where I claim?

Planning to merge tomorrow, so a check before then would be great.

@cjpurackal — could you also validate the new ORB / FAST surface end-to-end inside kornia-slam? The Python entry points (`kornia_rs.features.fast_detect`, `kornia_rs.features.orb_detect_and_compute`, `kornia_rs.features.match_descriptors`, `kornia_rs.k3d.find_homography`, two-view pose) all match the OpenCV-shape contract; want to make sure the slam pipeline picks them up cleanly before this lands. Thanks!

crates/kornia-3d/src/pose/twoview.rs
```diff
@@ -128,6 +209,11 @@ impl Default for TwoViewConfig {
             ransac_h: RansacParams::default(),
             homography_inlier_ratio: 0.8,
             triangulation: TriangulationConfig::default(),
+            lm_enabled: true,
+            lm: LmPoseConfig::default(),
+            use_5pt_essential: true,
```
cjpurackal:
```suggestion
            use_5pt_essential: false,
```
Setting this to true is breaking kornia-slam (requires calculating fundamental matrix)

cjpurackal (approved):

edgarriba:
@ducha-aiki @cjpurackal — this is ready for a look. Planning to merge tomorrow, so a pass on the reported numbers (in `kornia-py/benchmarks.md` and `crates/kornia-imgproc/SIMD.md`) would be much appreciated.

@cjpurackal — could you validate this end-to-end in kornia-slam? The Python ORB + two-view bindings (`OrbDetector`, `match_orb_descriptors`, `two_view_estimate`) mirror the Rust API used by `try_initialize_two_view` in `crates/kornia-slam/src/estimation/two_view.rs`, so the bootstrap path is the main thing I want exercised on a real sequence.

Thanks both!

crates/kornia-3d/src/pose/twoview.rs
edgarriba: Fixed in 9dfcc7d — flipped the default to `false` and updated the doc comment to call out kornia-slam's downstream F dependency. The 5pt path stays available behind the flag for callers who want translation-direction priority. Thanks for catching this!

edgarriba: Quick follow-up after re-reading the bench numbers and your call site.

I corrected the rationale in d72fac7 — the previous commit defended the default flip with a "kornia-slam needs F downstream" story that's not actually true: `examples/orb_slam/src/pipeline.rs:374` gates on `Fundamental(_)` but the body only reads `inlier_indices` and re-triangulates from poses computed elsewhere. The F matrix itself is never consumed.

The default flip stands on its own merits — on EuRoC MH_01: 8pt pose is **2.27× faster** (1.42 ms vs 3.21 ms) and **4× more rotation-accurate** (0.040° vs 0.164°) than 5pt; 5pt only wins translation direction by ~0.8°. So 8pt is the right default regardless.

A question about the gate itself: was the `matches!(tv.model, TwoViewModel::Fundamental(_))` check defensive, or is there a planned future use of F I'm missing? If defensive, would you be open to relaxing it to `Fundamental(_) | Essential(_)` in a follow-up kornia-slam PR? Otherwise anyone who later flips `use_5pt_essential = true` on a particular sequence (for the t-direction win on planar-ish scenes) silently gets `return 0` from the bootstrap, which is a footgun.

No blocker for this PR either way — just want to flag it before it bites.

PR #926, opened 2026-05-28
feat(kornia-py): expose VideoReader bindings

### LINKED ISSUE #916: [Feature]: Port VideoReader to PyO3 with time + keyframe seek
### 🚀 Feature Description

Bind `kornia_io::stream::video::VideoReader` (and `ImageFormat`) into `kornia-py` so Python users can open a video, iterate frames, and seek by **wall-clock time** or **nearest keyframe** without dropping to ffmpeg/OpenCV.

The Rust API in `kornia-io/src/gstreamer/video.rs` already exposes everything needed under the hood — `new(path, ImageFormat)`, `start/pause/close/reset`, `get_fps/get_pos/get_duration`, `grab_rgb8`, and a `seek(SeekFlags, Duration)` that already accepts GStreamer's `ACCURATE` / `KEY_UNIT` flags. This ticket is purely about exposing that to Python in an idiomatic way.

### 📂 Feature Category

Python Bindings (kornia-py)

### 💡 Motivation

`kornia-py` already has decoders for static images (jpeg/png/tiff/jpegturbo/rvl) but no video entry point at all (`kornia-py/src/io/` confirmed — no `video.rs`). Anyone doing video-in/SLAM/VLM work from Python currently has to bring in a separate library (OpenCV, decord, ffmpeg-python) just to read frames, then convert back to numpy, even though `kornia-rs` ships a perfectly capable GStreamer-backed `VideoReader`.

This shows up immediately in:
- Driving [`kornia-slam-py`](https://github.com/kornia/kornia-slam/pull/34) from a recorded `.mp4` instead of a sequence of PNGs.
- Feeding `kornia-vlm` clips to `SmolVLM` (the existing Rust `kornia-vlm/src/video.rs` uses `VideoReader` — Python users of the VLM crate can't easily replicate that workflow).
- Generic notebook/eval workflows where "open this video, jump to t=12.5s, give me the next 100 frames as numpy" is the natural primitive.

**Seeking matters specifically:**
- **Accurate (time-based) seek** is what users intuitively want ("show me frame at t=12.345s").
- **Keyframe seek** is dramatically cheaper for long videos — `SeekFlags::KEY_UNIT` snaps to the previous keyframe without decoding intervening frames, which is the difference between hundred-of-ms latency and seconds on a Jetson when scrubbing.

The underlying Rust `seek(SeekFlags, Duration)` already supports both — the feature is purely making that ergonomic and discoverable from Python.

### 💭 Proposed Solution

Add `kornia-py/src/io/video.rs` and register a `VideoReader` `#[pyclass]` plus an `ImageFormat` enum on the io submodule.

**Suggested Python surface:**

```python
from kornia_rs.io import VideoReader, ImageFormat

reader = VideoReader("clip.mp4", format=ImageFormat.Rgb8)
reader.start()

print(reader.fps, reader.duration_sec)        # float, float|None
frame = reader.grab()                          # np.ndarray (H, W, 3) u8, or None at EOS

# Time-based seek (accurate, slow — decodes to exact pos)
reader.seek_time(12.345)                       # default accurate=True

# Keyframe seek (fast, snaps to nearest keyframe before t)
reader.seek_keyframe(12.345)
# OR a flag on the same call:
reader.seek_time(12.345, accurate=False)

# Iterator protocol — most Python-friendly
for frame in reader:
    ...                                        # frame: numpy ndarray

reader.close()
```

**Mapping to existing Rust surface:**

| Python | Rust |
|---|---|
| `VideoReader(path, format)` | `VideoReader::new(path, ImageFormat::Rgb8 \| Mono8)` |
| `reader.start()` / `.pause()` / `.close()` / `.reset()` | direct |
| `reader.fps` (property) | `get_fps()` |
| `reader.position_sec` / `.duration_sec` | `get_pos()` / `get_duration()` as `f64` seconds |
| `reader.state` | `get_state()` (expose `StreamerState` as a `#[pyclass]` enum) |
| `reader.grab()` → numpy or `None` | `grab_rgb8()` (+ a `grab_mono8` equivalent for `ImageFormat::Mono8`) |
| `reader.seek_time(sec, *, accurate=True)` | `seek(SeekFlags::FLUSH \| (ACCURATE \| KEY_UNIT), Duration::from_secs_f64(sec))` |
| `reader.seek_keyframe(sec)` | `seek(SeekFlags::FLUSH \| KEY_UNIT, Duration::from_secs_f64(sec))` |
| `reader.set_playback_speed(s)` | `set_playback_speed(s)` |
| `__iter__` / `__next__` | loop wrapper around `grab_rgb8()` returning `None` at EOS |
| Context manager (`__enter__` / `__exit__`) | calls `start()` / `close()` |

**Numpy interop:** `grab()` returns a fresh `np.ndarray` of shape `(H, W, 3)` (Rgb8) or `(H, W)` (Mono8), dtype `uint8`. Always copy out of the `GstAllocator` buffer — that frees us from lifetime-coupling the numpy array to GStreamer's buffer pool.

**GIL release:** `grab()` and the seeks should release the GIL (`Python::allow_threads`) since they wait on the GStreamer pipeline and can block for tens of ms; this keeps the Python event loop / other threads responsive.

**Out of scope for v1 (good follow-ups):**
- Stereo / multi-stream `.mp4` (we have `grab_rgb8` only at the Rust layer for now).
- Hardware-accelerated decode toggles.
- Audio.
- Frame-index seek (would need a frame-by-frame index built on open; the underlying GStreamer API is time-based).

### 📚 Library Reference

- `kornia-io/src/gstreamer/video.rs` — the `VideoReader` impl (already has `seek(SeekFlags, Duration)`).
- `kornia-py/src/io/{jpeg,png,tiff}.rs` — existing house style for IO-module PyO3 wrappers.
- `gstreamer::SeekFlags` — `FLUSH | KEY_UNIT` for cheap keyframe seek, `FLUSH | ACCURATE` for exact-time seek.

### 🔄 Alternatives Considered

- **Tell Python users to use OpenCV / decord / ffmpeg-python.** Works today but means losing the kornia memory model and adding a heavy dep that doesn't share the rest of the kornia type ecosystem (Image, dlpack, etc.). It also means each downstream Python binding that wants video has to make the same choice.
- **dlpack-export the GStreamer buffer directly.** Tempting for zero-copy, but the buffer's lifetime is owned by GStreamer's pool and reusing it across a `grab()` call is fragile. Default to copy for v1; an optional zero-copy path can come later.
- **Only expose accurate seek, no keyframe seek.** Half-measure — keyframe seek is the differentiator for working with long-form video on edge devices.

### 🎯 Use Cases

- Drive `kornia-slam-py` from a `.mp4` instead of a PNG sequence.
- Notebook scrubbing: "give me the frame at t = 12.345s" without leaving Python.
- VLM eval pipelines that need to sample 1 frame per N seconds across long clips — keyframe seek makes this practical.
- Building dataset preprocessors (extract every K-th frame, dump to disk).
- Replacing OpenCV `VideoCapture` for users who are already in the kornia ecosystem.

### 📝 Additional Context

This is motivated directly by the work in https://github.com/kornia/kornia-slam/pull/34 — right now `kornia-slam-py` consumes images via numpy and the example crate has its own `EurocSource` / `OakdSource` for frames. A Python-facing `VideoReader` would round out the same workflow for arbitrary recorded videos. Companion ticket to https://github.com/kornia/kornia-rs/issues/915 (port `kornia-algebra` to PyO3).

### 🤝 Contribution Intent

- [ ] I plan to submit a PR to implement this feature
- [x] I'm requesting this feature but not planning to implement it

Varun-sai-500:
## 📝 Description

Adds a minimal Python binding for `kornia_io::stream::video::VideoReader` under `kornia_rs.io`.

This PR intentionally keeps the first version small and focused: opening a video, controlling the reader lifecycle, querying basic timing metadata, and grabbing RGB frames as NumPy arrays.

Fixes: #916 

---

## 🛠️ Changes Made
- [x] Added `kornia-py/src/io/video.rs` with `VideoReader` and `ImageFormat` PyO3 bindings.
- [x] Registered `VideoReader` and `ImageFormat` under the `kornia_rs.io` submodule.
- [x] Exposed `start`, `pause`, `close`, `reset`, `fps`, `position_sec`, `duration_sec`, and `grab()`.
- [x] Implemented `grab()` for `Rgb8`, returning `np.ndarray` with shape `(H, W, 3)` and dtype `uint8`.
- [x] Kept `Mono8` construction available but return a clear runtime error from `grab()` until the Rust backend exposes a mono grab path.

---

## 🧪 How Was This Tested?
- [ ] **Unit Tests:** Not added in this PR because video reader tests require GStreamer/video fixtures.
- [x] **Manual Verification:** Built `kornia-py` locally and verified that `VideoReader` / `ImageFormat` are importable from `kornia_rs.io`.
- [x] **Performance/Edge Cases:** `grab()` returns `None` when the backend reaches EOS/no frame, and copies the GStreamer-backed frame into an owned NumPy array to avoid lifetime coupling with the GStreamer buffer pool.

---

## 🕵️ AI Usage Disclosure
*Check one of the following:*
- [ ] 🟢 **No AI used.**

---

## 🚦 Checklist

---

## 💭 Additional Context
Add any other context or screenshots about the pull request here.

This is the minimal first step for exposing video IO in `kornia-py`. Follow-up PRs can add seeking APIs, iterator/context-manager ergonomics, GIL release around blocking calls, and additional examples/tests once the base binding surface is accepted.


Varun-sai-500:
@edgarriba  CI IS GREEN!

kornia-py/src/io/video.rs
```diff
@@ -0,0 +1,131 @@
...
+                let width = image.width();
+
+                let array = unsafe { PyArray::<u8, _>::new(py, [height, width, 3], false) };
+
+                unsafe {
+                    array
+                        .as_slice_mut()
+                        .expect("freshly allocated numpy array should be contiguous")
+                        .copy_from_slice(image.as_slice());
+                }
```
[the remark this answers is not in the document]
Varun-sai-500: pyO3 current version doesn't support py.allow_threads currently

kornia-py/src/io/video.rs
```diff
@@ -0,0 +1,131 @@
...
+
+                Ok(Some(array.unbind()))
+            }
+            PyImageFormat::Mono8 => Err(PyRuntimeError::new_err(
+                "Mono8 video grab is not supported yet",
+            )),
```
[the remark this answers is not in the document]
Varun-sai-500: First of, this PR is a V1, more shall come

kornia-py/src/io/video.rs
```diff
@@ -0,0 +1,135 @@
...
+
+    #[getter]
+    pub fn duration_sec(&self) -> PyResult<Option<f64>> {
+        Ok(self
+            .lock_reader()?
+            .get_duration()
+            .map(|duration| duration.as_secs_f64()))
+    }
+
+    pub fn grab(&self, py: Python<'_>) -> PyResult<Option<Py<PyArray3<u8>>>> {
```
edgarriba: Can we return the new Image api instead so that we can keep the image buffer in memory without copies?

edgarriba (changes requested):

kornia-py/src/io/video.rs
Varun-sai-500: @edgarriba I checked kornia-py/src/image.py and replaced the np.ndarray with kornia's function.

kornia-py/src/io/video.rs
```diff
@@ -0,0 +1,130 @@
...
+
+        let height = image.height();
+        let width = image.width();
+
+        let array = unsafe { PyArray::<u8, _>::new(py, [height, width, 3], false) };
+
+        unsafe {
+            array
+                .as_slice_mut()
+                .expect("freshly allocated numpy array should be contiguous")
```
edgarriba, line 120: This should throw a proper python error

kornia-py/src/io/mod.rs
```diff
@@ -5,3 +5,5 @@ pub mod jpegturbo;
 pub mod png;
 pub mod rvl;
 pub mod tiff;
+#[cfg(feature = "gstreamer")]
```
edgarriba, line 8: We should make sure thy when we make a release that gstreamer is included or throw proper errors. I'd suggest that you try locally to generate the python wheels and try this new  functionality

kornia-py/src/io/video.rs
```diff
@@ -0,0 +1,130 @@
...
+        match value {
+            PyImageFormat::Rgb8 => ImageFormat::Rgb8,
+            PyImageFormat::Mono8 => ImageFormat::Mono8,
+        }
+    }
+}
+
+#[pyclass(name = "VideoReader")]
+pub struct PyVideoReader {
+    reader: Mutex<VideoReader>,
```
edgarriba, line 27: Is there any workaround to remove the mutex here ? If the grab method is called at high frequency  the locker might be an issue

kornia-py/src/io/video.rs
```diff
@@ -0,0 +1,130 @@
...
+            .get_duration()
+            .map(|duration| duration.as_secs_f64()))
+    }
+    pub fn grab(&self, py: Python<'_>) -> PyResult<Option<PyImageApi>> {
+        let image = self
+            .lock_reader()?
+            .grab_rgb8()
+            .map_err(|e| PyRuntimeError::new_err(e.to_string()))?;
+
+        let Some(image) = image else {
```
edgarriba, line 108: There are better semantics to be fused with the above to do that

kornia-py/src/io/video.rs
```diff
@@ -0,0 +1,130 @@
...
+            return Ok(None);
+        };
+
+        let height = image.height();
+        let width = image.width();
+
+        let array = unsafe { PyArray::<u8, _>::new(py, [height, width, 3], false) };
```
edgarriba, line 115: It's fine for now but we should be able in the future that the Image api can handle the returned image directly from the reader to keep the data buffer of gstreamer instead of copying data to numpy

kornia-py/src/io/video.rs
```diff
@@ -0,0 +1,130 @@
...
+            array
+                .as_slice_mut()
+                .expect("freshly allocated numpy array should be contiguous")
+                .copy_from_slice(image.as_slice());
+        }
+
+        Ok(Some(PyImageApi::wrap(
+            py,
+            array.unbind(),
+            Some("RGB".to_string()),
```
edgarriba, line 127: This should come from the set image format

edgarriba (changes requested):

kornia-py/src/io/video.rs
```diff
@@ -0,0 +1,130 @@
+use crate::image::PyImageApi;
+use kornia_io::stream::video::{ImageFormat, VideoReader};
+use numpy::{PyArray, PyArrayMethods};
+use numpy::{PyArray, PyArrayMethods};
```
sidd-27, line 4: this wont compile

sidd-27 (changes requested):

PR #964, opened 2026-07-01
feat(imgproc): Preprocessor builder — RGBA source, mean/std normalize, configurable pad

edgarriba:
## Motivation

The GPU `Preprocessor` (added in #960) does one thing: tight-RGB8 `Image<u8,3>` → resize+pad → CHW `f32` `[0,1]`, **CUDA-only**. Three gaps:

1. **CPU support** — it's unusable (and untestable in CI) without a GPU.
2. **RGBA sources** — camera/NVMM surfaces are often RGBA and need a separate repack pass first.
3. **Mean/std normalize** — most backbones want `(v/255-mean)/std`, currently a second pass.

## What changed

- **CPU + CUDA backends.** `PreprocessorBuilder::build()` → a **CPU** preprocessor (host image → host tensor, always available, no CUDA toolchain). `build_cuda(stream)` → the fused GPU kernel (feature `cudarc`). `run` dispatches on the backend; the CPU path is a host loop that mirrors the kernel exactly. The `preprocess` module is no longer feature-gated (only the GPU parts are).
- **`run` is channel-generic**: `Image<u8, 3>` (RGB) or `Image<u8, 4>` (RGBA — alpha skipped). One fused pass RGBA→CHW, no separate RGBA→RGB step.
- **`Normalize`** — `UnitScale` (default) or `MeanStd { mean, std }` with a `Normalize::imagenet()` helper, folded into the same pass.
- **Configurable letterbox pad value** (was hardcoded 114).

### Design notes
- One **parameterized** kernel covers every mode × format × normalization (`src_bpp`, pitch, per-channel `mean`/`inv_std`, pad are launch args — single JIT compile, no divergence). `UnitScale` = identity affine.
- The **CPU path is the test oracle**: `cpu_matches_cuda` asserts CPU output == CUDA output.
- Backward compatible: `letterbox()`/`stretch()`/`with_mode()` (now `cudarc`-gated, they take a stream) and `run(&Image<u8,3>, ..)` are unchanged.

## Proof of verification (per AI policy)

AI-assisted (Claude Code); I can explain every line. Jetson Orin (SM87, CUDA 12.6):

```
# default features (CPU, no CUDA)
$ cargo test -p kornia-imgproc preprocess
test result: ok. 4 passed; 0 failed; ...

# with the GPU backend (includes the CPU==CUDA equivalence oracle)
$ cargo test -p kornia-imgproc --features cudarc preprocess -- --include-ignored
test result: ok. 5 passed; 0 failed; ...
```
`clippy --all-targets -- -D warnings` clean for **both** default and `--features cudarc`; `cargo fmt --check` clean.

## Open questions
- Naming: `build()`/`build_cuda()` split OK, or prefer explicit backend enum?
- Should a standalone pitched→tight `Image` repack (for non-model consumers) live here too?


edgarriba:
Self-review pass (independent reviewer over the diff) — no correctness issues found: CPU and CUDA paths verified numerically identical (there's a `cpu_matches_cuda` test asserting this for letterbox+stretch under ImageNet-normalize), bilinear indexing sound, the `C ∈ {3,4}` + RGBA alpha-skip guards correct, and feature-gating clean (CPU path builds with no `cudarc`).

Two robustness nits raised (both integer-cast truncation in the CUDA launch: `usize→u32` thread count and `usize→i32` dims, for tensors beyond the kernel's `int` pixel index — >2.1B px, unreachable on real GPUs). Addressed in `a37a76af`: `run_cuda` now returns `DimensionsTooLarge` instead of silently truncating. The CPU path is `usize` throughout, unaffected.

Also did a small simplify pass (`ad3d2855`): `run` is a thin CPU/CUDA dispatch, geometry threaded via an `Affine` struct.

Re-verified on Jetson Orin: CPU 4 tests + `--features cudarc` 5 tests (incl the equivalence oracle) + `clippy --all-targets -D warnings` for both feature sets + fmt — all green.

edgarriba:
Addressed the Copilot review in f7a817e4: `Normalize::MeanStd` now rejects non-finite/non-positive std (and non-finite mean) at build time (`InvalidNormalize`); `run` validates the destination is `[1,3,H,W]` before writing (`BadOutputShape` — anything else would be an OOB device write on the CUDA path); the kernel's "arbitrary row pitch" comment now matches reality (pitch param kept for generality, launcher always passes tight `src_w * C`). Two new rejection tests. Re-verified on Jetson: CPU 6 + cudarc 7 tests (incl the CPU==CUDA oracle), clippy -D warnings both feature sets, fmt.

edgarriba:
`af2c4058` — two changes from self-review of the CPU backend:

**1. Sampling modes.** New `sampling(InterpolationMode)` builder option: `Nearest`, `Bilinear` (default), or `Lanczos` — the filters in common CV/ML use — supported on both backends. `Bicubic` (declared in the enum but unimplemented) is rejected at build.

**2. CPU path now reuses the crate's optimized kernels instead of a hand-rolled scalar loop.** It rides `resize_fast_u8_aa` (the NEON-accelerated separable resampler, PIL-grade antialiased Lanczos) to the letterbox content box, then a fused scale+bias normalize / HWC→CHW placement — and the common RGB stretch+bilinear case collapses to the crate's fully-fused `resize_normalize_to_tensor_u8_to_f32_bilinear` (one kernel call). The CUDA kernel gains the `sampling` branch (nearest / bilinear / lanczos-3, direct sampling).

**Backend semantics note:** the CPU resampler is antialiased + centre-aligned (higher quality); the GPU kernel samples directly on the fractional affine. Outputs agree on content but aren't bit-identical, and with a fractional letterbox pad the content/pad boundary can sit one pixel apart (documented in the code). The exact-equality oracle is therefore replaced by per-backend constant-preservation tests (all modes, both backends) plus a CPU-vs-CUDA band test on a smooth gradient with integer-pad geometry.

Verified on Jetson Orin: full kornia-imgproc CPU suite **292 passed / 0 failed** (default features), cudarc preprocess suite **10/10** incl the GPU sampling + band tests, `clippy --all-targets -D warnings` clean on both feature sets, fmt clean.

edgarriba:
`945a9d7e` — structural consistency pass on the resize module + fused-first dispatch:

**Kernels seam everywhere.** `bilinear.rs` / `nearest.rs` were the only resize algorithm files carrying raw inner loops; their per-row loops moved into `kernels.rs` (`bilinear_row_u8` + `BilinearXTaps`, `nearest_row_u8`), matching the module's own layering and the color-op convention — algorithm files do LUT precompute + rayon, `kernels.rs` owns per-arch code (scalar + aarch64 NEON + x86_64 AVX2 dispatchers; the new rows are scalar with the backend slots ready per "Adding a new backend"). Pure code motion: resize output is bit-identical (full imgproc suite unchanged, 292/292).

**Fused ops first.** `resize/mod.rs` now documents the fused entry points as the preferred path for model inputs, and the `Preprocessor` CPU path follows it: **every** RGB+bilinear case (stretch *and* letterbox) runs the fused resize+normalize+CHW kernel in one f32 pass — letterbox fuses into the content box then row-memcpys it into the padded planes, eliminating the intermediate u8 requantization it previously had. Nearest / AA-Lanczos / RGBA keep the NEON fast resizer + fused scale+bias placement.

Verified on Jetson Orin: full CPU suite 292/292 + 51 integration, cudarc preprocess 10/10 (GPU sampling + band tests), clippy `-D warnings` both feature sets, fmt.

edgarriba:
Benchmark regression pass vs `main` (Jetson Orin, MAXN, release; Rust micro-bench + the `bench_fused_preprocess` example + the Python bindings):

| bench (p50) | main | PR | Δ |
|---|---|---|---|
| `resize_fast` C3 1080p→640×480 nearest | 0.154 ms | 0.143 ms | ✅ |
| `resize_fast` C3 bilinear | 0.566 ms | 0.483 ms | ✅ |
| `resize_fast` C3 lanczos (AA) | 5.44 ms | 5.33 ms | ✅ |
| `resize_fast` C4 bilinear | 0.657 ms | 0.60–0.82 ms | noise band |
| pyrdown exact-2× | 0.193 ms | 0.191 ms | ✅ |
| fused `resize_normalize_to_tensor` (example) | 0.474 ms | 0.447 ms | ✅ |
| py `Image.resize` bilinear | 0.705 ms | 0.555 ms | ✅ |
| py fused `resize_normalize_to_tensor` | 0.888 ms | 0.936 ms | noise band |
| **GPU `Preprocessor` letterbox 720p→640²** | **0.121 ms** | **0.344 ms → 0.114 ms** | **fixed in `137f8337`** |

The pass caught one real regression: the single branching CUDA kernel let the lanczos 6×6 accumulators inflate register pressure for every launch — 2.8× slower on the common bilinear path. Since sampling is build-time config, `build_cuda` now compiles only the selected variant (`resize_normalize_to_chw_{nearest,bilinear,lanczos}`, shared `__device__` helpers); after the split the GPU path is at parity with main's original kernel *with* the normalize fold included. Everything else is at parity or better (CPU numbers wobble ±10% run-to-run on this board; the "noise band" rows moved in both directions across repeats of identical code).

CI was 27/27 green on `945a9d7e`; re-running on `137f8337`. No reviewer comments outstanding (the three Copilot items were addressed earlier).

edgarriba:
`9faab9a6` — comparative benchmarks vs OpenCV showed two losses; both fixed. kornia is now **at-or-ahead of cv2 like-for-like on every mode** (Jetson Orin, 1080p→640×480 RGB u8, p50, python-to-python):

| op | OpenCV 4.13 | kornia-rs | |
|---|---|---|---|
| resize nearest | 0.18–0.32 ms | **0.178 ms** | ✅ (was 0.299) |
| resize bilinear | 0.697 ms | **0.613 ms** | ✅ |
| resize lanczos, cv2 semantics (`antialias=False`) | 2.84 ms | **1.79 ms** | ✅ 1.6× (was only AA at 5.3) |
| resize lanczos, PIL/AA semantics | n/a in cv2 | 5.35 ms | quality option |
| full preprocess (resize+norm+CHW f32) | 11.31 ms | **1.21 ms** (fused) | ✅ 9.3× |

Rust-side (criterion `bench_resize`, nearest 2×-down 256×224): kornia `resize_fast` 18 µs vs the `image` crate 479 µs.

Changes:
1. **`nearest_row_u8`** moved pixels with per-byte 3-byte copies; now RGB moves as overlapping u32 load/stores (xmap is monotonic → one precomputed cutoff separates the u32-safe body from the exact tail) and RGBA as exact u32s. 0.143 → **0.115 ms** at the Rust level.
2. **py `Image.resize` gains an `antialias` kwarg** (default `True`, behavior unchanged). It was hardcoded to AA, so python users couldn't reach the fixed-kernel lanczos that matches (and beats) cv2's `INTER_LANCZOS4`.

Full suites green: imgproc CPU 292/292 + 51 integration, clippy `-D warnings` (imgproc both feature sets + kornia-py), fmt.

edgarriba:
Per-operation breakdown of the preprocess chain — where the fused win comes from (Jetson Orin CPU, 1080p→640×480 RGB u8 → CHW f32 ImageNet-normalized, p50, python↔python):

| step | cv2 + numpy | kornia-rs |
|---|---|---|
| 1. resize bilinear (u8→u8) | 0.684 ms | **0.650 ms** |
| 2. `astype(f32) / 255` | 4.042 ms | — |
| 3. `(x − mean) / std` | 7.880 ms | — |
| 4. HWC→CHW + contiguous | 0.740 ms | — |
| sum of steps | 13.35 ms | |
| chained as one expression | 7.36 ms | |
| **all 4 ops fused** (`resize_normalize_to_tensor`) | — | **1.188 ms** |
| torch CPU chain, reference (`permute+interpolate+norm`) | 5.80 ms | |

The resize was never the bottleneck — the elementwise normalize (steps 2+3) is ~12 of the 13.3 ms, because numpy materializes three full-size f32 temporaries and re-walks memory for each. The fused kernel touches every byte once, so the whole chain costs ≈ resize + ~0.5 ms: **6.2×** vs numpy's best one-liner, **4.9×** vs the torch CPU chain, **9.3×** vs the step-by-step pipeline. Standalone, each op also wins individually (nearest 0.178 vs 0.18–0.32, bilinear 0.650 vs 0.684, lanczos same-semantics 1.79 vs 2.84 ms).

edgarriba:
Addendum — the **non-fused kornia chain** for completeness (the repo's own `bench_fused_preprocess` example, Rust-level, 1080p→540p exact-2×):

| kornia path | p50 |
|---|---|
| resize alone (u8→u8) | 0.190 ms |
| separate: resize → normalize → CHW (3 passes) | 3.776 ms |
| **fused (one pass)** | **0.447 ms — 8.4× vs separate** |

Even with vectorized Rust kernels, the unfused chain pays ~8× — these ops are memory-bound, so the cost is the number of passes over the buffer, not the per-kernel speed. That's the case for keeping the fused entry points the preferred path (as now documented in `resize/mod.rs`).

edgarriba:
`8b9e8bbb` — multi-resolution + byte-fidelity pass, one more real perf find, and a permanent benchmark.

## Speed across resolutions (Orin CPU, p50 ms, python↔python)

| case | cv2 bilinear | kornia | cv2 LANCZOS4 | kornia noAA | PIL lanczos AA | kornia AA | chain cv2+numpy | kornia FUSED |
|---|---|---|---|---|---|---|---|---|
| 4K→1080p | 1.86 | **1.35** | 16.9 | **13.5** | 153.0 | **~18.5** | 75.1 | **2.46 (30×)** |
| 1080p→640×480 | 0.73 | 0.81 | 2.88 | **1.99** | 33.4 | **5.4** | 8.23 | **0.90 (9×)** |
| 720p→512² | 0.57 | **0.52** | 2.42 | **1.68** | 16.1 | **2.9** | 6.35 | **0.88 (7×)** |
| 640×480→224² | 0.53 | **0.16** | 2.16 | **0.46** | 5.19 | **1.13** | 1.63 | **0.21 (8×)** |

The 4K AA number includes another fix this commit: the strip-fused separable orchestration re-runs the horizontal pass over the ky-row inter-strip overlap, and the L2-only heuristic ignored that — at 4K AA (ky≈13, strip_h≈4) most of the horizontal work was redundant: **48.6 → 18.2 ms (2.7×), byte-identical** (checksums verified across both plans). The choice now requires the strip's source band to cover ≥6 kernel heights; short-kernel strip wins are preserved.

## Byte-to-byte fidelity

| comparison | result |
|---|---|
| nearest vs PIL | **100% byte-exact** (shared center convention; cv2 uses floor — PIL sides with kornia) |
| bilinear vs cv2 | ≤ **1 LSB** (88–94% identical) |
| bilinear vs **float64 ground truth** (torch, same op) | kornia **100.0%** correctly-rounded bytes at 1080p→640×480 and **99.1%** at 640×480→224² vs cv2's **88.8%** — the bytes that differ from cv2 are cv2's Q11 rounding error, not ours (Q14) |
| lanczos AA vs PIL | 96–98% identical, mean ≤0.06 |
| lanczos noAA vs cv2 | differs by design (3-lobe vs LANCZOS4's 4-lobe) |

## Permanent benchmark

`kornia-py/benchmarks/bench_resize_vs_libs.py` joins the existing python benchmarks: the full wall-time matrix + the byte-fidelity checks (incl. the f64-truth rounding score), so these comparisons are reproducible.

Deliberately NOT done (would change bytes): pyramid-prefiltered 2-stage AA lanczos and separable-triangle bilinear — both faster but drift from PIL/cv2 fidelity. A possible future fidelity+~1% lever is keeping fraction bits in the i16 intermediate, but the shifts are NEON/AVX2 immediates across six kernels — left as follow-up.

edgarriba:
`34bebcb0` — roadmap-driven feature pass. Rationale: measured against real pipelines (TensorRT detectors at 40–150 ms/frame), preprocess is 0.2–2% of the frame budget — so the remaining wins live on the **pipeline's critical path** (zero-copy input, fp16 output, pass elimination, batch amortization), not in single-op CPU races that are already at the DRAM wall.

**New in this commit:**

1. **`PitchedSurface` + `run_surface[_f16]`** (CUDA) — consume camera/NVMM **pitched** interleaved surfaces directly: resize + normalize + (RGBA→)RGB in the one fused kernel. Jetson pipelines currently burn a whole GPU repack pass (pitched RGBA → tight RGB) before preprocessing; this deletes it. GPU test: pitched output **byte-identical** to the tight-image path.
2. **fp16 output** (CUDA) — `run_f16` / `run_surface_f16` emit half-precision CHW for fp16 TensorRT engines (halves output traffic, removes the pre-inference cast pass). In-kernel RNE conversion with no `cuda_fp16.h` dependency; GPU test proves **bit-equality** with `half::f16::from_f32` of the f32 result. Kernel source refactored into shared `__device__` samplers + thin per-(sampling × dtype) entries, keeping per-variant compilation.
3. **CPU fused chains for every RGB mode** — nearest (single-pass gather+FMA) and **AA lanczos** (SIMD separable horizontal + f32-emitting vertical) now go straight to the normalized CHW tensor like bilinear: one less pass, and the torchvision-parity AA chain loses its intermediate u8 quantization.
4. **`pipeline.resize_normalize_to_tensor_batch`** (python) — one GIL release for a whole list; images run back-to-back (each fused call already saturates rayon — per-image fan-out oversubscribes, measured slower). 1080p→640×480 ×8: **0.75 ms/img** vs 1.20 single-call (10.8× the cv2+numpy chain). Parallelism belongs *across* dataloader workers, not inside one call.
5. Benchmark script now reports **p99** alongside p50 (real-time systems are judged at tails) and covers the batch path.

Deliberate non-goals, from the analysis: chasing cv2's nearest/bilinear cells (5× there breaks the LPDDR5 bandwidth floor — confirmed against fast_image_resize's ARM64 numbers) and byte-changing speed tricks (2-stage AA, triangle bilinear). A PIL-bit-parity precision mode (fraction-bits in the i16 intermediate) is noted as a separate-PR follow-up since it changes output bytes and needs x86 hardware to validate the AVX2 paths.

Verified on Orin: imgproc CPU 292/292, cudarc **12/12** GPU tests (incl the two new byte-equality proofs), clippy `-D warnings` on default + cudarc + kornia-py, fmt.

edgarriba:
Final benchmark state of the PR (`34bebcb0`, all features in) — Jetson Orin (aarch64, MAXN), RGB u8, p50 ms, python↔python, fully reproducible via `kornia-py/benchmarks/bench_resize_vs_libs.py`:

## Resize ops

| op | | 4K→1080p | 1080p→640×480 | 720p→512² | 640×480→224² |
|---|---|---|---|---|---|
| nearest | cv2 | 1.184 | **0.177** | 0.176 | 0.139 |
| | **kornia** | **0.988** | 0.220 | **0.125** | **0.036 (3.9×)** |
| bilinear | cv2 | 1.577 | 1.049 | **0.575** | 0.531 |
| | **kornia** | **1.201** | **0.570 (1.8×)** | 0.591 | **0.190 (2.8×)** |
| lanczos noAA | cv2 LANCZOS4 | 17.28 | 2.88 | 2.36 | 2.16 |
| | **kornia** | **13.68** | **2.31** | **1.38 (1.7×)** | **0.500 (4.3×)** |
| lanczos AA | PIL | 153.0 | 33.7 | 16.2 | 5.29 |
| | **kornia** | **20.3 (7.6×)** | **4.39 (7.7×)** | **2.69 (6.0×)** | **1.05 (5.1×)** |

## Preprocess chain (u8 → normalized CHW f32)

| | 4K | 1080p | 720p | 224² |
|---|---|---|---|---|
| cv2 + numpy | 70.1 | 8.25 | 6.32 | 1.63 |
| **kornia FUSED** | **2.44 (29×)** | **1.33 (6.2×)** | **0.89 (7.1×)** | **0.185 (8.8×)** |
| fused p50 / p99 | 2.52 / 2.83 | 1.07 / 4.47 | 1.22 / 2.69 | 0.21 / 0.89 |
| **batch×8 per-img p50 / p99** | 6.05 / 6.79 ¹ | **0.89 / 1.36** | **1.03 / 1.27** | **0.142 / 0.164** |

¹ 4K batch saturates LPDDR bandwidth (8×24 MB sustained) — batching is a dataloader tool for small targets, where it shines: 224² at 0.142 ms/img with the tightest p99 in the table.

## Byte fidelity

| | 4K | 1080p | 720p | 224² |
|---|---|---|---|---|
| nearest vs PIL | 100% | 100% | 100% | 91.1% ² |
| bilinear vs cv2 | 100% | ≤1 LSB | ≤1 LSB | ≤1 LSB |
| **bilinear vs f64 ground truth (kornia / cv2)** | 100 / 100 | 100 / 100 | **100 / 88.0** | **99.1 / 88.7** |
| lanczos AA vs PIL | 97.9% | 97.2% | 97.8% | 96.1% |

² half-pixel tie-rounding convention at that exact ratio.

**Summary**
- AA lanczos is now **≥5× vs PIL at every resolution** (5.1–7.7×), at 96–98% byte parity with PIL's own output.
- kornia wins every op at every size except two coin-flips inside cv2's memcpy-bound noise band (they flip across runs; 4K nearest now wins).
- Fused chain is **6.2–29×** the cv2+numpy pipeline; the batch API reaches **0.142 ms/img** at 224² with p99 0.164.
- Where kornia and cv2 bilinear bytes differ (≤1 LSB), the f64 ground truth sides with kornia: **99–100% correctly-rounded vs cv2's 88%** at the sizes where they diverge (Q14 vs Q11 coefficients).

edgarriba:
Self-review of the feature commit (`34bebcb0`): one real finding, independently confirmed by a second reviewer — the kernel-side `f2h` converter collapsed **NaN to ±Inf** (exp==255 path ignored the mantissa). Fixed in 1903d8e2 with a quiet-NaN mantissa bit; unreachable through the preprocessor (u8 inputs + validated-finite normalize params) but wrong as a converter. Everything else came back clean (py batch borrow soundness, band splitting, surface validation, kernel macro args). Same commit also simplifies: shared validators for the repeated channel/shape/surface checks, and the two duplicate plane→band splitters in fused.rs merged into one `plane_bands()` helper. Suites re-verified green (CPU 292/292, cudarc 12/12).

edgarriba:
Reconciled with main's merged CUDA color stack (#966/#969) and extended:

- **Merge repair**: the origin/main merge had silently dropped all of main's CUDA work (gpu/color_cuda kernels, residency dispatch, pinned memory, kornia_rs.cuda, benchmarks, CI job). Restored verbatim; the 4 real conflicts (imgproc Cargo.toml, py lib.rs, py cuda.rs, Cargo.lock) hand-merged.
- **Fusion ported onto this builder**: `SourceFormat::{Rgb,Bgr,Rgba,Bgra,Gray,Nv12,Yuyv}` decodes **inside the resize taps** via a warp-uniform `fmt` launch arg — the single-JIT-compile design of this PR is preserved (no per-format kernel entries, zero extra NVRTC cost). Q20 BT.601 decode byte-identical to `gpu/color_cuda/video`; fused == chained bit-for-bit (tested per format × sampling).
- **New entries**: `run_raw` / `run_raw_f16` (raw device frame → CHW in one launch), `run_raw_batch` (N frames → `[N,3,H,W]`, one sync per batch).
- **Python**: `CudaPreprocessor` rewritten onto the builder (adds lanczos + `pad_value`), `run_batch`, and `kornia_rs.cuda.from_dlpack` — device-resident torch/cupy tensor → `CudaImage` by owned d2d copy (`upgrade_device_ptr` + `memcpy_dtod` + `leak`, stream=1 per the DLPack protocol).

Jetson: 315 imgproc lib tests (16 preprocess incl. GPU-ignored), 88 tensor, 26 image, 10 pytest — all green. clippy/fmt clean.

PR #971, opened 2026-07-03
perf(imgproc): optimize CPU warp-affine — incremental coords, valid-range skip, 16-row Rayon chunks

Incharajayaram:
## 📝 Description

**Fixes/Relates to:** Follow-up to #967 (merged). Optimizes the f32 CPU `warp_affine` path which was the bottleneck identified during GPU benchmarking.

---

## 🛠️ Changes Made

- **`crates/kornia-imgproc/src/warp/affine.rs`**
  - [x] Replace per-pixel `transform_point` (6 mults + 6 adds each) with incremental coordinate update: precompute row-start `(sx0, sy0)` and step by `(dsx, dsy)` per column
  - [x] Add analytical valid-range computation per row: solve `0 ≤ d·x + s₀ < upper` for x in O(1), zero-fill borders with `fill()`, skip inner loop entirely for fully out-of-bounds rows
  - [x] Lift `InterpolationMode` match outside the parallel loop; switch from per-row Rayon tasks to 16-row chunks (same granularity as `parallel.rs`) — reduces task count from 1080 → 68 at 1920×1080
  - [x] Fix zero-threshold in `compute_range` from `1e-12` → `1e-6` to handle f32 trig imprecision (`cos(π/2) ≈ −4.4e-8` which caused spurious empty ranges for axis-aligned rotations)
  - [x] Remove dead `transform_point` helper and unused `crate::parallel` import

- **`crates/kornia-imgproc/benchmarks.md`**
  - [x] Update nearest-neighbor CPU column with criterion medians (was pre-optimization stale baseline)
  - [x] Add `kornia-rs CPU ms` and `kornia CPU vs cv2 CPU` columns to bilinear table with live-measured cv2 numbers
  - [x] Update key findings

---

## 🧪 How Was This Tested?

- [x] **Unit Tests:** All 5 existing warp-affine tests pass. `warp_affine_correctness_rot90` specifically exercises the `1e-6` threshold fix — it fails without it.
- [x] **Manual Verification:** Criterion bench + live `bench_opencv_warp_affine.py` run on the same machine (i5-10300H, 8 Rayon threads, `--release`).
- [x] **Performance/Edge Cases:** Identity transform, 45° and 90° rotations, fully out-of-bounds rows all produce correct output.

**Measured results — 45° rotation, 3-ch f32, criterion median:**

| Size | Before (ms) | After (ms) | vs cv2 CPU |
|---|---:|---:|---|
| nearest 256×224 | 0.540 | 0.263 | — |
| nearest 512×448 | 2.132 | 0.860 | — |
| nearest 1024×896 | 8.544 | 3.983 | — |
| nearest 1920×1080 | 19.15 | 9.869 | — |
| bilinear 256×224 | — | 0.161 | **3.6× faster** |
| bilinear 512×448 | — | 1.535 | **1.5× faster** |
| bilinear 1024×896 | — | 5.775 | **1.4× faster** |
| bilinear 1920×1080 | — | 13.82 | **2.3× faster** |

The optimized CPU bilinear beats cv2 CPU (`cv2.warpAffine` f32, OpenCV 4.12.0) at every tested size without any SIMD. GPU nearest is still 23–28× faster than the optimized CPU.

---

## 🕵️ AI Usage Disclosure

---

## 🚦 Checklist

### DIFF
--- a/crates/kornia-imgproc/src/warp/affine.rs
+++ b/crates/kornia-imgproc/src/warp/affine.rs
@@ -4,7 +4,7 @@ use kornia_image::{Image, ImageError};
 use rayon::prelude::*;
 
 use super::kernels::process_affine_span;
-use crate::interpolation::{interpolate_pixel_fast, validate_interpolation, InterpolationMode};
+use crate::interpolation::{validate_interpolation, InterpolationMode};
 
 /// Inverts a 2x3 affine transformation matrix.
 ///
--- a/crates/kornia-imgproc/src/warp/affine.rs
+++ b/crates/kornia-imgproc/src/warp/affine.rs
@@ -147,6 +147,8 @@ pub fn warp_affine<const C: usize>(
     let compute_range = |sx0: f32, sy0: f32| -> (usize, usize) {
         let (mut lo, mut hi) = (0i64, dst_w as i64);
         for (d, s0, upper) in [(dsx, sx0, src_w_f), (dsy, sy0, src_h_f)] {
+            // 1e-6 covers f32 trig imprecision (e.g. cos(π/2) ≈ −4.4e-8);
+            // assumes source step ≥ ~1e-6 px/col (extreme downscale > 1e6:1 is not a use case here).
             if d.abs() < 1e-6 {
                 if s0 < 0.0 || s0 >= upper {
                     return (0, 0);
@@ -259,34 +261,9 @@ pub fn warp_affine<const C: usize>(
                     );
                 });
         }
-        _ => {
-            dst.as_slice_mut()
-                .par_chunks_mut(row_len * ROWS_PER_TASK)
-                .enumerate()
-                .for_each(|(ci, chunk)| {
-                    let y_base = ci * ROWS_PER_TASK;
-                    for (dy, dst_row) in chunk.chunks_exact_mut(row_len).enumerate() {
-                        let y_f = (y_base + dy) as f32;
-                        let sx0 = m_inv[1] * y_f + m_inv[2];
-                        let sy0 = m_inv[4] * y_f + m_inv[5];
-                        let (x_lo, x_hi) = compute_range(sx0, sy0);
-                        dst_row[..x_lo * C].fill(0.0);
-                        dst_row[x_hi * C..].fill(0.0);
-                        let mut sx_fb = sx0;
-                        let mut sy_fb = sy0;
-                        for (x, dst_pixel) in dst_row.chunks_exact_mut(C).enumerate() {
-                            if x >= x_lo && x < x_hi {
-                                dst_pixel.iter_mut().enumerate().for_each(|(k, p)| {
-                                    *p =
-                                        interpolate_pixel_fast(src, sx_fb, sy_fb, k, interpolation);
-                                });
-                            }
-                            sx_fb += dsx;
-                            sy_fb += dsy;
-                        }
-                    }
-                });
-        }
+        // validate_interpolation at the top of this function rejects every mode
+        // other than Nearest and Bilinear, so this arm is unreachable.
+        _ => unreachable!(),
     }
 
     Ok(())
--- a/crates/kornia-imgproc/src/warp/affine.rs
+++ b/crates/kornia-imgproc/src/warp/affine.rs
@@ -1,10 +1,10 @@
 use std::f32::consts::PI;
 
 use kornia_image::{Image, ImageError};
+use rayon::prelude::*;
 
 use super::kernels::process_affine_span;
 use crate::interpolation::{interpolate_pixel_fast, validate_interpolation, InterpolationMode};
-use crate::parallel;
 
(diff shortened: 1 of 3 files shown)


edgarriba (commented):
Reviewed the CPU `warp_affine` rewrite — verified locally on branch `perf/cpu-warp-affine`.

**Verified:**
- `cargo test -p kornia-imgproc warp` → 13 passed, 0 failed. `cargo clippy` clean.
- Memory-safe: nearest clamps `xi/yi`, bilinear clamps `sx_c/sy_c` to `[0,dim-1]`, so float-accumulation drift past the analytical range cannot cause OOB reads.
- Math checks out: `dsx=m_inv[0]`, `dsy=m_inv[3]`, row-start `m_inv[1]*y+m_inv[2]`; the `0 ≤ d·x + s0 < upper` solve with ceil-based half-open `[lo,hi)` is correct.
- Inline bilinear matches the old `bilinear_interpolation` replicate-edge behavior at borders (clamp + `fx=0` is equivalent to the old `val01=val00` fold).
- `u8` path (`process_affine_span`) untouched.

Nice speedup. Three non-blocking notes below.

crates/kornia-imgproc/src/warp/affine.rs
```diff
@@ -134,23 +128,166 @@ pub fn warp_affine<const C: usize>(
...
+    // Lift the interpolation branch outside the parallel loop so each task is
+    // branch-free in its inner loop.
+    macro_rules! run_rows {
+        ($chunk:expr, $y_base:expr, $inner:expr) => {
+            for (dy, dst_row) in $chunk.chunks_exact_mut(row_len).enumerate() {
+                let y_f = ($y_base + dy) as f32;
+                let sx0 = m_inv[1] * y_f + m_inv[2];
+                let sy0 = m_inv[4] * y_f + m_inv[5];
+                let (x_lo, x_hi) = compute_range(sx0, sy0);
+                dst_row[..x_lo * C].fill(0.0);
```
edgarriba, line 185: **Behavior change — OOB pixels now zero-filled.** Old path left out-of-bounds dst pixels *unwritten* (retained prior dst content); this now explicitly zeros them, i.e. a constant-0 border. This is more consistent with `warpAffine` border semantics and all tests pass — just flagging it as an intentional semantic change so it is not a surprise to callers that relied on the old fill-nothing behavior. Confirm intended.

crates/kornia-imgproc/src/warp/affine.rs
```diff
@@ -134,23 +128,166 @@ pub fn warp_affine<const C: usize>(
...
+                                        + w11 * src_slice[b11 + k];
+                                }
+                                *sx += dsx;
+                                *sy += dsy;
+                            }
+                        }
+                    );
+                });
+        }
+        _ => {
```
edgarriba: **Dead match arm.** `validate_interpolation` rejects Lanczos/Bicubic at the top of `warp_affine`, so only `Nearest`/`Bilinear` ever reach this `match` — this `_ =>` arm is unreachable. It also does not clamp `sx_fb/sy_fb` and calls `interpolate_pixel_fast`, which `debug_assert!(false)`s on those modes. Harmless while dead, but confusing. Suggest deleting the arm (add `unreachable!()`/`unsafe { unreachable_unchecked() }` or a comment) rather than keeping a partial non-clamped fallback.

crates/kornia-imgproc/src/warp/affine.rs
```diff
@@ -134,23 +128,166 @@ pub fn warp_affine<const C: usize>(
...
+    let src_w_f = src_w as f32;
+    let src_h_f = src_h as f32;
+
+    // Analytical valid-x range helpers — same logic as warp_affine_u8.
+    // Solve 0 <= dsx*x + sx0 < src_w and 0 <= dsy*x + sy0 < src_h for x,
+    // then intersect to get [x_lo, x_hi) where both coords are in-bounds.
+    let compute_range = |sx0: f32, sy0: f32| -> (usize, usize) {
+        let (mut lo, mut hi) = (0i64, dst_w as i64);
+        for (d, s0, upper) in [(dsx, sx0, src_w_f), (dsy, sy0, src_h_f)] {
+            if d.abs() < 1e-6 {
```
edgarriba, line 152: **`1e-6` threshold is scale-independent.** Correct for normal rotate/translate/mild-scale (and the fix for `cos(π/2)≈-4.4e-8` spurious empties is a good catch). Edge case: an extreme downscale where a legitimate `dsx < 1e-6` gets misclassified as a constant column. Not blocking — worth a one-line comment noting the assumption (source step ≥ ~1e-6 px/col).

edgarriba (approved):
CI green across all targets (Rust, py3.8–3.14t, wheels x86_64+aarch64, C++, Windows, macOS). Verified locally: 13 warp tests pass, clippy clean, memory-safe, bilinear parity with old kernel. Non-blocking notes left inline. Approving for 0.1.15-rc.1.

PR #972, opened 2026-07-03
docs(py): fused-preprocess → torch/TensorRT inference example (DLPack zero-copy)

edgarriba:
Closes the deployment loop on the CUDA preprocessing stack (#964/#970):

`kornia-py/examples/preprocess_to_inference.py` — raw 1080p NV12 frames → `CudaPreprocessor.run_batch` (one fused decode+letterbox+normalize+CHW **fp16** kernel per frame) → `torch.from_dlpack` (**zero-copy**) → ResNet-18.

Both engines consume the fused kernel's fp16 output directly — the TensorRT engine is built with **fp16 I/O**, so there is no copy *and no dtype cast* at the framework boundary.

Measured on Jetson Orin Nano, batch 4 → 224², **including preprocessing**:

| engine | ms/frame | FPS |
|---|---|---|
| torch fp16 (cuDNN autotune) | 2.44 | 409 |
| TensorRT 10 fp16 I/O (`--engine trt`) | 0.74 | 1354 |

Notes:
- Engine cache is keyed by (batch, size) — engines bake in their shapes, and a stale engine bound to differently-shaped raw pointers would misbench or scribble device memory.
- ONNX exported with the legacy exporter (TRT 10.3 parser rejects the dynamo exporter's initializer layout); torch device pointers bound directly (`execute_async_v3`). First engine build is slow on Orin Nano; cached afterward.

### DIFF
--- a/crates/kornia-imgproc/src/preprocess.rs
+++ b/crates/kornia-imgproc/src/preprocess.rs
@@ -88,16 +88,18 @@ pub enum Normalize {
     },
 }
 
-/// The standard ImageNet (mean, std) in RGB order, matching torchvision —
+/// The standard ImageNet per-channel mean (RGB order), matching torchvision —
 /// the single source for [`Normalize::imagenet`] and language bindings.
-pub const IMAGENET_NORMALIZE: ([f32; 3], [f32; 3]) = ([0.485, 0.456, 0.406], [0.229, 0.224, 0.225]);
+pub const IMAGENET_MEAN: [f32; 3] = [0.485, 0.456, 0.406];
+/// The standard ImageNet per-channel std (RGB order), matching torchvision.
+pub const IMAGENET_STD: [f32; 3] = [0.229, 0.224, 0.225];
 
 impl Normalize {
     /// The standard ImageNet mean/std (RGB), matching torchvision.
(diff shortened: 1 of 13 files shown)


edgarriba:
Follow-up commit applies the previously-skipped /simplify items:

- **Stream-ordered DLPack export** (the real one): `__dlpack__` now honors the consumer's `stream` argument per the array-API protocol — event fence for foreign streams, no-op for the producer's own default stream, host sync only for legacy consumers that pass nothing. No more device-wide `synchronize()` per handoff. Numbers: torch 2.44 → **2.32 ms/frame**, TRT 0.74 → **0.68 ms/frame (1480 FPS)**.
- `kornia_rs.cuda.IMAGENET_MEAN/STD` constants sourced from the new Rust `IMAGENET_NORMALIZE` const — single truth with `Normalize::imagenet()`; 5 scattered copies reduced.
- `--frames` → `--iters` (no silent rounding); shared `nv12_frame` test fixture.

Still deliberately not done: importing `benchmarks/_bench.py` into the example — its per-call timing forces a per-iteration host sync, which would reintroduce exactly the serialization the stream-ordered handoff just removed.

PR #974, opened 2026-07-05
feat(imgproc): bicubic GPU kernels for resize and warp-affine (NVRTC/cudarc)

Incharajayaram:
## Summary

- Adds `launch_warp_affine_bicubic_cuda` and `launch_resize_bicubic_cuda` using Keys cubic interpolation (`a = -0.5`, matching OpenCV `INTER_CUBIC`).
- Each kernel samples a 4×4 tap neighborhood (16 reads) around the inverse-mapped source coordinate.
- Out-of-range taps are clamped to the nearest border pixel (BORDER_REPLICATE); pixels whose center maps outside the source are zero-filled (BORDER_CONSTANT).
- All 16 reads go through `__ldg` (read-only L1 cache). Same 32×8 thread block and `CU_FUNC_CACHE_PREFER_L1` as the existing bilinear/nearest kernels.
- Resize uses half-pixel center alignment (matches OpenCV/PIL) and handles both upscale and downscale.

## Kernel optimizations

Weights are computed via Horner-form polynomials (branch-free, no `fabsf`) rather than a generic `cubic_w` helper. Since `frac ∈ [0,1)` each of the 4 tap distances falls into a known polynomial region:

| tap | distance | polynomial |
|-----|----------|------------|
| -1  | 1+frac   | second (−0.5t³+2.5t²−4t+2) |
|  0  | frac     | first (1.5t³−2.5t²+1) |
| +1  | 1−frac   | first |
| +2  | 2−frac   | second |

Additional micro-opts: `wx[4]/wy[4]` precomputed in registers before the loop, row base addresses hoisted outside the inner loop, `#pragma unroll` on both loop levels, `fmaf()` for all accumulations.

**Speedup vs initial implementation (GTX 1650):**
- Resize bicubic downscale: **+7–10%**
- Resize bicubic upscale: **+33–34%** (was compute-bound, not DRAM-bound)
- Warp-affine bicubic: ~flat (scattered reads from rotation dominate)

### Bicubic resize

| Source → Dest | kornia-rs ms | GB/s | cv2 CPU ms | cv2 CUDA ms | PyTorch GPU ms | vs cv2 CPU | vs cv2 CUDA | vs PyTorch |
|---------------|-------------|------|-----------:|------------:|---------------:|------------|-------------|------------|
| 1024²→512²    | 0.120       | 52.6 | 1.071      | 0.320       | 0.803          | **8.9×**   | **2.7×**    | **6.7×**   |
[...]

PR #979, opened 2026-07-06
perf(imgproc): texture object kernels for warp-affine GPU ops

Incharajayaram:
## Summary

- Adds a `CudaTexObject` RAII wrapper (`texture.rs`) for `CU_TR_ADDRESS_MODE_BORDER` pitch-2D textures, used internally by warp-affine kernels.
- Replaces the explicit OOB bounds-check branch in `warp_affine_bilinear` and `warp_affine_nearest` with texture border mode — OOB fetches silently return 0.0, eliminating the warp divergence that penalises rotated images where ~30-40% of output pixels map outside the source frame.
- Adds an optional `block_dim: Option<(u32, u32)>` parameter to bilinear and nearest warp-affine launchers for tuning small images.
- Texture objects were evaluated for resize kernels but reverted after benchmarking revealed a 10% regression (`tex2D` has ~100-cycle latency vs ~30 for `__ldg`; sequential downscale access coalesces through unified L1 just as well as the texture cache, so the higher instruction latency dominates with no OOB benefit).

> **Depends on #974** — built on top of the bicubic kernel branch. Intended to be reviewed/merged after #974.

### Warp-affine bilinear — 45° rotation (texture vs previous `__ldg` + explicit OOB check)

| size | before (ms) | after (ms) | speedup |
|---|---|---|---|
| 256 × 224 | 0.025 | 0.017 | **32%** |
| 512 × 448 | 0.092 | 0.060 | **35%** |
| 1024 × 896 | 0.274 | 0.220 | **20%** |
| 1920 × 1080 | 0.572 | 0.513 | **10%** |

### Resize bilinear downscale (unchanged — `__ldg` retained)

| size | ms/iter |
|---|---|
| 1920×1080 → 960×540 | 0.178 |

## Why texture helps warp-affine but not resize

For a 45° rotation, roughly 30-40% of output pixels map outside the source frame. The old `if (sx < 0 || sx >= src_w || sy < 0 || sy >= src_h)` guard causes massive warp divergence — threads that take the OOB branch stall the entire warp. `CU_TR_ADDRESS_MODE_BORDER` eliminates this branch entirely; border texels return 0.0 in hardware. The scattered access pattern of rotation also benefits more from the 2D-spatial texture cache than sequential downscale reads.

For resize, coordinates are always clamped in bounds, there is no OOB divergence to remove, and the access pattern is sequential — `__ldg` through unified L1 already coalesces as well as the texture cache while avoiding the higher instruction latency of `tex2D`.

## Test plan

- [ ] `cargo clippy -p kornia-imgproc --features "gpu-cuda" -- -D warnings` passes
- [ ] `cargo run --example bench_gpu_warp_affine --features gpu-cuda --release` shows improvement vs baseline on a CUDA-capable machine
- [ ] `cargo run --example bench_gpu_resize --features gpu-cuda --release` shows no regression vs baseline
- [ ] Bicubic warp-affine is unaffected (retains `__ldg` path, no texture change)

### DIFF
--- a/crates/kornia-imgproc/examples/bench_gpu_resize.rs
+++ b/crates/kornia-imgproc/examples/bench_gpu_resize.rs
@@ -281,11 +281,11 @@ fn run_gpu_cuda() {
                           dst: &mut cudarc::driver::CudaSlice<f32>| {
                 match method {
                     "nearest" => launch_resize_nearest_downscale_cuda(
-                        &ctx, &stream, src, dst, sw, sh, dw, dh,
+                        &ctx, &stream, src, dst, sw, sh, dw, dh, None,
                     )
                     .expect("nearest launch"),
                     _ => launch_resize_bilinear_downscale_cuda(
-                        &ctx, &stream, src, dst, sw, sh, dw, dh,
+                        &ctx, &stream, src, dst, sw, sh, dw, dh, None,
                     )
                     .expect("bilinear launch"),
                 }
@@ -422,6 +422,7 @@ fn run_gpu_cuda_fused_normalize() {
                 sh,
                 dw,
                 dh,
+                None,
             )
             .expect("bilinear launch");
         }
@@ -438,6 +439,7 @@ fn run_gpu_cuda_fused_normalize() {
                 sh,
                 dw,
                 dh,
+                None,
             )
             .expect("bilinear launch");
         }
@@ -457,6 +459,7 @@ fn run_gpu_cuda_fused_normalize() {
                 dh,
                 mean,
                 std,
+                None,
             )
             .expect("fused launch");
         }
@@ -475,6 +478,7 @@ fn run_gpu_cuda_fused_normalize() {
                 dh,
                 mean,
                 std,
+                None,
             )
             .expect("fused launch");
         }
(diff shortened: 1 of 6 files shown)


crates/kornia-3d/src/pnp/epnp.rs
```diff
@@ -314,7 +314,7 @@ fn select_control_points(points_world: &[Vec3AF32]) -> [Vec3AF32; 4] {
         (s_diag[1].sqrt(), Vec3AF32::new(v_y.x, v_y.y, v_y.z)),
         (s_diag[2].sqrt(), Vec3AF32::new(v_z.x, v_z.y, v_z.z)),
     ];
-    axes_sig.sort_by(|a, b| b.0.total_cmp(&a.0));
+    axes_sig.sort_by(|a, b| a.0.total_cmp(&b.0).reverse());
```
edgarriba: Why is this needed

crates/kornia-apriltag/src/decoder.rs
edgarriba, line 1030: Why is this needed

edgarriba (commented):
Fix ci

crates/kornia-apriltag/src/decoder.rs
Incharajayaram: > Why is this needed

clippy's unnecessary_sort_by was failing on this

edgarriba (approved):
LGTM -- but rebase from main to have clear history

PR #995, opened 2026-07-07
feat(kornia-py): unified device-aware Image API (CUDA) + memory-leak tests

edgarriba:
## What

Gives `kornia-py` a **unified, device-aware Python API** that mirrors the Rust crates 1:1 instead of inventing GPU-only parallel types. Device-resident pixels live in the same `kornia_rs.image.Image`; the fused preprocessor and its output tensor are single types that dispatch CPU/GPU internally; DLPack / CUDA Array Interface interop is zero-copy throughout.

A device `Image` is a plain `Image` whose `.device` is `"cuda:{id}"` — move with `.to_cuda()` / `.cpu()`, download with `.numpy()`, share zero-copy via `__dlpack__` / `__cuda_array_interface__`.

## API surface

- **`Image`** — one type for host and device. `.device` (`"cpu"` / `"cuda:{id}"`), `.to_cuda(stream=None)`, `.cpu()`, `Image.zeros(w, h, c, dtype, stream=None)` (stream => direct device alloc), `Image.from_dlpack(obj)` (device inferred: CPU tensor -> host image, CUDA tensor -> device image on its own source device).
- **`Tensor`** (top-level `kornia_rs.Tensor`) — mirrors `kornia_tensor::Tensor`, residency-aware `[N, C, H, W]` model input. Zero-copy `data_ptr` for TensorRT `set_tensor_address`, plus `__dlpack__` / `__cuda_array_interface__`. Host `f32` `.numpy()` is a true zero-copy view.
- **`Preprocessor`** (top-level `kornia_rs.Preprocessor`) — fused resize + normalize + HWC->CHW. **The constructor's `stream=` selects the device** (mirroring Rust `build_cuda(&stream)` / `Image::to_cuda(&stream)`): `stream=None` -> CPU, `stream=<Stream>` -> the fused CUDA kernel on that stream's device. One `run(frame, w, h, out_h, out_w, out=None, consumer_stream=None)` entry point that dispatches on argument shape (single frame / batch list / `out=` in-place) **and accepts an already-decoded interleaved `Image`** (device Image -> zero-copy kernel input, host Image -> uploaded like raw bytes). `consumer_stream=` fences a caller's engine stream after the preprocess with no host sync.
- **`imgproc.*` color ops** — accept numpy **or** `Image`, dispatch on residency (numpy/host -> CPU SIMD, device -> GPU kernel), return the same kind. One entry per conversion instead of separate `cuda.*` variants.
- **`kornia_rs.cuda`** — keeps only genuinely CUDA-exclusive bits: `Stream` (adopts foreign NVIDIA streams; selects the device), `is_available`, `mem_get_info`, IMAGENET consts.

`Tensor` / `Preprocessor` are **not gated on the `cuda` cargo feature** — `Preprocessor(stream=None)` and host tensors (including host `Image` input to `run()`) work on a CPU-only build; constructing with `stream=<Stream>` requires a CUDA-featured build.

## Memory safety (device DLPack / staging)

Several latent bugs found by a multi-agent audit and fixed:

- **DLPack device-import use-after-free** — the import borrowed the producer's `__dlpack__` capsule but never *consumed* it, so on return the capsule GC'd and ran the producer's deleter at import time, dangling the zero-copy alias for any producer that hands buffer ownership to the managed tensor (torch/cupy were only accidentally safe). Now the capsule is **consumed** (renamed to `used_dltensor`, suppressing the producer's destructor) and the source object kept alive for the alias's lifetime. *(A follow-up own-managed-tensor variant that also runs the producer's deleter to release its manager context was reverted — it segfaulted on py3.12/3.13 in CI — so releasing that context is a known deferred item, tracked separately.)*
- **`run_cpu` integer-overflow** — `width*height*c` was unchecked; an oversized shape could wrap past the size guard and read/write out of bounds. Now checked via `backing::byte_len`, plus an up-front overflow reject on DLPack import.
- **`run(out=)` residency check** — a host tensor passed as `out=` to a device preprocessor was written through as a device pointer (UB; aliases device memory on Jetson unified memory). Now rejects non-device / wrong-device `out=`.

## Multi-GPU correctness

- `Tensor.__dlpack_device__` reports the tensor's real ordinal (was hardcoded `0`).
- `to_cuda` fences the foreign consumer against the image's own producing stream (not an idle one) on the already-device path.
- `Stream.from_handle` probes a foreign stream's actual device (was always device 0).
- Device color ops allocate their destination on the source image's own stream/device (avoids spurious `DeviceMismatch`).
- `imgproc.rgb_from_rgba` / `rgb_from_bgra` honor `background=` on device images (was silently dropping alpha).

## API changes

- `Image.from_dlpack` no longer takes `copy` — **always zero-copy**, mirroring torch. Callers wanting an owned buffer copy the result themselves. (`from_numpy`'s `copy` and the `__dlpack__` protocol kwarg are unchanged.)
- `CudaImage` / `CudaTensor` / `CudaPreprocessor` / `Image.cuda.*` / `download`/`upload` are gone — folded into the unified `Image` / `Tensor` / `Preprocessor` and `.to_cuda` / `.cpu` / `.numpy`.

## Structure

`cuda.rs` (~1800 lines) was split into a focused `cuda_ext/` module: `mod.rs` (stream helpers, `Tensor`, `Preprocessor`), `capsule.rs` (DLPack pack/export + device import), `cuda_color.rs` (GPU color ops). Extensive dedup (macros for the `TensorInnerEnum` match, DLPack capsule twins, color-op dispatch, f16 widening).

## Review fixes (post-unification)

A second multi-agent review pass over the Preprocessor unification + `Image::to_cuda` merge caught four issues, all fixed:

- **Workspace compile break (CI)** — the `to_cuda_image` -> `to_cuda` merge made `Image::to_cuda` return an `Image`, but `examples/cuda_imgproc` still bound it to a `Tensor` (E0308). CI's full-workspace `cargo check` compiles `examples/*`, so this failed **Rust Lint**; use the backing Tensor's unchanged `to_cuda`.
- **Silent garbage on non-uint8 host `Image`** — `run()`'s host-`Image` branch read the buffer via a debug-only-asserted `u8_elems()`, so a float32/uint16 host `Image` was reinterpreted as a wrong-sized u8 slice in release. Now rejected with a clear error (mirrors the device branch).
- **Cross-stream read-before-write** — a device `Image` produced on a different stream than the preprocessor could be read before its writes completed. The launch stream now event-fences against the Image's producing stream (skipped when the same stream — the common case), matching imgproc's `cuda_dispatch` discipline.
- **CPU-build host-`Image` input** — the `run()` Image arm was `cuda`-gated, so `run(host_image)` on a CPU-only build fell to a misleading numpy error despite the docstring promising support. Un-gated; host Images route to the CPU kernel there.

## Testing

- `tests/test_image_device.py` — device API + DLPack / CAI interop.
- `tests/test_cuda_memory.py` — no-leak integration tests bracketing hot loops (H2D/D2H, color kernels, DLPack export/import keepalive, preprocessor staging) with `mem_get_info`, plus a positive control proving the probe is sensitive.
- `tests/test_interop.py` — torch/TensorRT zero-copy hand-off; `benchmarks/bench_cuda_interop.py`.

Full suite: **583 passing** on a Jetson Orin (torch 2.11 CUDA + TensorRT 10.3); the only failure is the known flaky `test_mem_probe_is_sensitive` (device-memory high-water-mark timing on this box, not a code issue). Both `--features cuda` and CPU-only (`--no-default-features`) builds compile and lint clean.


edgarriba:
Follow-up (036625f): **foreign NVIDIA stream adoption**, so kornia device ops integrate with an existing `cuda-python` / `cuda.core` / CuPy stream.

- `Stream.from_handle(int)` / `Stream.from_cuda_stream(obj)` — adopts a raw `CUstream`. Accepts the `__cuda_stream__()` protocol, a bare int, or a `.ptr` / `.handle` integer attribute.
- A `Stream` is now `Owned` (kornia's default) or `Foreign` (a **borrowed** handle — never `cuStreamDestroy`'d by us). cudarc 0.19 exposes no non-owning `CudaStream`, so ops run on kornia's default stream and are **fenced into** the foreign stream with a CUDA event (`cuStreamWaitEvent`) — equivalent for cross-stream ordering, no host block. Wired into `Image.cuda.from_numpy` / `.to_cuda` / `Image.cuda.zeros`.
- Tests cover protocol extraction (int / `.ptr` / `.handle` / `__cuda_stream__` / rejection), end-to-end use through all three constructors, and a no-leak test for the per-op event fence.

Note: full integration is one-directional-plus-fence, not literal submission onto the caller's stream — that would need a non-owning stream wrapper cudarc doesn't currently provide. Semantically equivalent for ordering; flagged here in case we later want to upstream a borrowed-stream API into cudarc to drop the extra event.

edgarriba:
Follow-up (3f162f8): **TensorRT-friendly** device model-input path — camera → preprocess → engine with no host round-trip.

**`CudaTensor` interop surface**
- `data_ptr` (raw device pointer int) → bind straight to `ctx.set_tensor_address(name, t.data_ptr)`; plus a `device` getter.
- `__cuda_array_interface__` (v3) for cupy / numba / cuda-python, advertising the producing stream.
- `__dlpack__` now honors `max_version`: versioned `dltensor_versioned` capsule for DLPack 1.0+ consumers (torch / NumPy 2), unversioned fallback otherwise. (The device `Image` already did this; this closes the gap for `CudaTensor`.)

**`CudaPreprocessor`**
- `run` / `run_batch` take an optional consumer `stream=` and fence the output into it (event + `cuStreamWaitEvent`), so `execute_async_v3(stream)` is ordered after the preprocess with no host sync.
- `alloc_output()` + `run_into(out, ...)` for an allocation-free serving loop: preallocate once, overwrite a fixed TensorRT input binding each frame. The destination is wrapped as a non-owning tensor so we never free the caller's buffer.

**Correctness fix:** device `Image` / `CudaTensor` CAI now report `stream = 1` (legacy default) instead of the ambiguous `0` that real CAI consumers (CuPy, Numba) reject.

Tests: interop surface, consumer-stream run, `run_into` (f32/f16 + dtype-mismatch guard), versioned-capsule negotiation, and a `run_into` no-leak test — full CUDA suite **42 passed, 3 skipped** (torch absent). New example `examples/cuda_preprocess_tensorrt.py` runs end-to-end on a GTX 1650.

This lands items 1–4 from the discussion. The only remaining "nice-to-have" is upstreaming a borrowed-stream wrapper into cudarc so foreign streams could be submitted onto directly (dropping the fence event) — out of scope here.

crates/kornia-tensor/src/cuda.rs
```diff
@@ -1043,6 +1094,45 @@ mod tests {
...
+        let stream = ctx.default_stream();
+
+        // SAFETY: the tensor is fully overwritten via memcpy_htod before any read.
+        let mut dev = unsafe { uninit_cuda::<u8, 1>([4], &stream) }.unwrap();
+        assert!(dev.as_cudaslice().is_some());
+
+        let slice = dev.as_cudaslice_mut().unwrap();
+        stream.memcpy_htod(&[7u8, 8, 9, 11], slice).unwrap();
+
+        let back = dev.download().unwrap();
```
edgarriba: No download idiom

kornia-py/benchmarks/bench_cuda_interop.py
```diff
@@ -0,0 +1,67 @@
...
+        mb = a.nbytes / 1e6
+        t_zc = bench(lambda: torch.from_dlpack(dev))
+        t_cpy = bench(lambda: torch.from_numpy(dev.numpy()).cuda())
+        t_gpu = bench(lambda: ip.gray_from_rgb(dev))
+        t_cpu = bench(lambda: ip.gray_from_rgb(a), n=100)
+        print(f"{label:>7} {mb:5.1f} | {t_zc:9.1f}us {t_cpy:11.1f}us {t_cpy / max(t_zc, 1e-6):6.0f}x"
+              f" | {t_gpu:7.1f}us {t_cpu:7.1f}us")
+
+    # Fused serving path: preallocate once, run_into each frame (zero per-frame alloc).
+    pre = kornia_rs.cuda.CudaPreprocessor(mode="letterbox", format="rgb", f16=True)
```
edgarriba: Preprocessor should be also unified

kornia-py/examples/cuda_preprocess_tensorrt.py
```diff
@@ -0,0 +1,85 @@
...
+- ``t.__cuda_array_interface__`` → cupy / numba
+- ``torch.from_dlpack(t)``  → PyTorch (versioned DLPack when negotiated)
+
+For a steady serving loop, preallocate the output once with ``alloc_output`` and
+overwrite it every frame with ``run_into`` (no per-frame allocation), and pass
+your engine's execution stream so the preprocess is ordered before
+``execute_async_v3`` without a host sync.
+
+Run:
+    pip install "kornia-rs[cuda]"       # needs an NVIDIA driver + NVRTC
```
edgarriba: Is this real the cud's feature in pip ?

kornia-py/examples/image_cuda_basics.py
```diff
@@ -0,0 +1,86 @@
...
+            assert img3.device == "cuda:0"
+    except ImportError:
+        pass
+
+    # 8) Zero-copy interop with cupy / numba / cuda-python via the CUDA Array
+    #    Interface (device images expose __cuda_array_interface__).
+    try:
+        import cupy as cp
+
+        arr = cp.asarray(img)  # zero-copy view of the device buffer
```
edgarriba, line 76: And from cupy tensor to img?

kornia-py/python/kornia_rs/cuda.pyi
```diff
@@ -22,46 +62,43 @@ IMAGENET_STD: Tuple[float, float, float]
...
+    Wraps ``cuMemGetInfo`` (synchronizes the default stream first). Bracket a
+    loop with it to assert the free byte count returns to baseline — i.e. no
+    device memory leaked across the iterations."""
 
-    copy=True: pixels are copied device-to-device into an owned buffer; the
-    producer may free its tensor immediately after.
-    copy=False: zero-copy — the image aliases the producer's memory and keeps
-    the producer object alive; mutating the producer mutates the image.
-    """
+class CudaTensor:
```
edgarriba: Shouldn't exist

kornia-py/python/kornia_rs/image.pyi
```diff
@@ -93,10 +93,15 @@ class Image:
...
+    ``Image.cuda.zeros(...)``, ``Image.cuda.from_dlpack(...)``."""
+
     @staticmethod
     def from_dlpack(obj: Any) -> Image:
-        """Import a DLPack tensor (numpy >= 1.22, PyTorch, CuPy CPU) as a
-        zero-copy Image.  The producer object is kept alive as a keep-alive."""
+        """Import a DLPack tensor (numpy >= 1.22, PyTorch, CuPy, cuda-python…),
+        **inferring the device**: a CPU tensor becomes a host ``Image`` (zero-copy,
+        producer kept alive), a CUDA tensor becomes a device-resident ``Image``."""
         ...
```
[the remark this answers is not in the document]
edgarriba: Default should be zero copies

kornia-py/src/color.rs
```diff
@@ -40,30 +123,68 @@ pub fn gray_from_rgb_f32(py: Python<'_>, image: PyImageF32) -> PyResult<PyImageF
...
     background: Option<[u8; 3]>,
-) -> PyResult<PyImage> {
-    let src = unsafe { numpy_as_image::<4>(py, &image)? };
-    let (mut dst, out) = unsafe { alloc_output_pyarray::<3>(py, src.size())? };
-    py.detach(|| color::rgb_from_rgba(&src, &mut dst, background))
-        .map_err(to_pyerr)?;
-    Ok(out)
+) -> PyResult<Py<PyAny>> {
+    // The GPU path (opaque alpha drop) ignores `background`.
+    #[cfg(feature = "cuda")]
```
edgarriba: We should measure that this new dispatcher doesn't affect performance. Profiling is needed

kornia-py/tests/test_cuda.py
```diff
@@ -74,7 +113,7 @@ def test_preprocessor_nv12_fused():
     t = pre.run(frame, w, h, 64, 64)
     assert t.shape == (1, 3, 64, 64)
     assert t.dtype == "float32"
-    out = t.download()
+    out = t.numpy()
```
edgarriba, line 117: Let's force that t.to_host().numpy() where to_host can raise errors

kornia-py/tests/test_interop.py
```diff
@@ -0,0 +1,164 @@
...
+
+
+def test_dlpack_import_zero_copy_alias():
+    """torch CUDA tensor -> Image.from_dlpack(copy=False) aliases the producer."""
+    torch = pytest.importorskip("torch")
+    if not torch.cuda.is_available():
+        pytest.skip("no torch CUDA")
+    a = _rgb()
+    t = torch.as_tensor(a, device="cuda")
+    img = Image.from_dlpack(t, copy=False)
```
edgarriba: Review dlpack protocol wether this is the expected behaviour

PR #998, opened 2026-07-11
feat(imgproc): add CUDA remap kernel with architecture decision benchmark

Incharajayaram:
## 📝 Description

**Fixes/Relates to:** 

This PR adds a generic CUDA remap primitive: the caller precomputes `(map_x, map_y)` float coordinate maps and the kernel samples the source image at those locations using bilinear or nearest-neighbor interpolation.

---

## 🛠️ Changes Made

- [x] `crates/kornia-imgproc/src/cuda/remap.rs` — new module with:
  - `launch_remap_bilinear_cuda` — bilinear remap, 3-ch f32, texture-backed with `CU_TR_ADDRESS_MODE_BORDER` (OOB → 0)
  - `launch_remap_nearest_cuda` — nearest-neighbor remap, same signature
  - `remap_maps_from_affine` — CPU helper: invert 2×3 affine, emit `(map_x, map_y)` for each output pixel
  - `remap_maps_from_homography` — CPU helper: invert 3×3 homography via adjugate/det, emit maps
  - `CudaRemapError` with `Cuda(String)` and `SliceTooSmall { what, got, need }` variants
- [x] `crates/kornia-imgproc/examples/bench_cuda_remap.rs` — architecture decision benchmark comparing remap vs fused warp-affine (bilinear, nearest, homography)
- [x] `crates/kornia-imgproc/src/cuda/mod.rs` — register `pub mod remap`

---

## 🧪 How Was This Tested?

- [x] **Build:** `cargo check -p kornia-imgproc --features cuda` — clean
- [x] **Lint:** `cargo clippy -p kornia-imgproc --features cuda` — no errors
- [x] **Format:** `cargo fmt -p kornia-imgproc` — applied
- [x] **Manual Verification:** `cargo run --example bench_cuda_remap --features cuda` on GTX 1650
  - Bilinear remap overhead vs fused affine: −4% to +21% (size-dependent, borderline at smaller sizes)
  - Nearest-neighbor remap overhead vs fused affine: +30% to +51% (consistently slower)
  - Architecture decision recorded in benchmark output: warp-perspective will use separate fused kernels
- [x] **Performance/Edge Cases:**
  - Zero-dimension guard in both launchers (returns `CudaRemapError::Cuda` before any GPU work)
  - Src slice size validated before texture creation (`check_image_slice`)
  - Degenerate homography (det ≈ 0) maps to `(-1.0, -1.0)` sentinel → texture border returns 0
  - OOB coordinates pass through to texture sampler (BORDER_CONSTANT = 0), not clamped
- [x] **Self-review:** ran `/review` (4 bugs found and fixed) then `/simplify` (6 cleanup items applied)

---

## 🕵️ AI Usage Disclosure

---

## 🚦 Checklist

---

## 💭 Additional Context

**Architecture decision rationale** (from benchmark on GTX 1650, 200 iters, 45° rotation):

| case | remap-BL ms | affine-BL ms | overhead |
|------|-------------|--------------|----------|
| 512×512 | 0.094 | 0.098 | −4% |
| 1280×720 | 0.310 | 0.256 | +21% |
| 1920×1080 | 0.699 | 0.661 | +6% |
| 3840×2160 | 2.731 | 2.613 | +5% |

| case | remap-NN ms | affine-NN ms | overhead |
|------|-------------|--------------|----------|
| 512×512 | 0.081 | 0.062 | +30% |
| 1920×1080 | 0.623 | 0.420 | +48% |

Nearest-neighbor remap overhead is consistently 30–51% → warp-perspective will be implemented as separate fused kernels (PR F), not via remap. Remap remains as a standalone primitive for arbitrary non-linear distortions (lens undistortion, fisheye, etc.).


crates/kornia-imgproc/src/cuda/remap.rs
```diff
@@ -0,0 +1,472 @@
...
+    for dy in 0..dst_height {
+        let y = dy as f32;
+        let row_sx = i[1] * y + i[2];
+        let row_sy = i[4] * y + i[5];
+        let row_w = i[7] * y + i[8];
+        for dx in 0..dst_width {
+            let x = dx as f32;
+            let w = i[6] * x + row_w;
+            if w.abs() < 1e-10 {
+                mx.push(-1.0);
```
edgarriba: **Correctness — negative w not culled (doc/code mismatch).** The doc above (L271) says pixels with w "zero **or negative**" map to the `(-1,-1)` sentinel, but `w.abs() < 1e-10` only catches near-zero. A **negative** w passes this check and gets divided normally, yielding a ghost/mirrored source coord for points behind the plane instead of the sentinel. OpenCV's `warpPerspective` treats `w <= 0` as outside.

```rust
if w < 1e-10 {
```

Sign-aware check matches the doc and correctly culls back-projected pixels. Currently silent wrong output on strong perspective tilts.

crates/kornia-imgproc/src/cuda/remap.rs
```diff
@@ -0,0 +1,472 @@
...
+
+/// Build a `(map_x, map_y)` coordinate map on the CPU from a 3×3 homography matrix.
+///
+/// `h` is a row-major 3×3 homography (forward: source → destination).
+/// The map encodes the **inverse** perspective transform (destination → source),
+/// so that remap produces the same result as a fused warp-perspective kernel.
+///
+/// The returned vecs have `dst_w * dst_h` elements each, in row-major order.
+/// Pixels whose homogeneous w-coordinate is zero or negative are mapped to
+/// `(-1.0, -1.0)`, which falls outside the source and returns 0 via border mode.
```
edgarriba: This doc says "zero or negative", but the implementation (L311) only tests `w.abs() < 1e-10`. Either fix the code to `w < 1e-10` (preferred — see comment there) or soften this sentence to "near-zero". Keep the two in sync.

crates/kornia-imgproc/src/cuda/remap.rs
```diff
@@ -0,0 +1,472 @@
...
+//!   neither kernel uses shared memory.
+//! * **32×8 thread block (default)** — full warp per output row for coalesced
+//!   destination writes; same rationale as resize and warp-affine.
+//!
+//! # Public API
+//!
+//! * [`launch_remap_bilinear_cuda`] — bilinear remap, 3-ch f32.
+//! * [`launch_remap_nearest_cuda`]  — nearest-neighbor remap, 3-ch f32.
+//! * [`remap_maps_from_affine`]     — build map_x/map_y from a 2×3 affine matrix.
+
```
edgarriba, line 41: Public API list omits `remap_maps_from_homography` — it's public and documented below. Add a bullet for it.

crates/kornia-imgproc/src/cuda/remap.rs
```diff
@@ -0,0 +1,472 @@
...
+}
+
+fn make_tex(
+    src: &CudaSlice<f32>,
+    stream: &Arc<CudaStream>,
+    src_width: u32,
+    src_height: u32,
+) -> Result<(CudaTexObject, u64), CudaRemapError> {
+    let (dev_ptr, _guard) = src.device_ptr(stream);
+    let tex =
```
edgarriba: nit: `warp_affine.rs::make_tex` carries a `// Safety:` comment explaining that `_guard` records the stream event on drop *after* the texture is destroyed (correct ordering). Behavior here is identical but the comment was dropped — worth copying for the next reader.

edgarriba (commented):
Automated review pass. Three inline items — the negative-w cull is a real correctness bug (contradicts the function's own doc); the rest are a missing test suite and doc nits. Nice work overall; matches `warp_affine.rs` conventions closely.

**Not inlined:** no unit tests added. `remap_maps_from_affine` / `remap_maps_from_homography` are pure deterministic CPU math — the cheapest thing to test and where the subtle logic lives (affine inversion, adjugate/det homography inverse, sentinel handling). Suggest a `#[cfg(test)]` mod: identity affine → `(x,y)`; known rotation vs hand-computed coords; degenerate homography (det≈0) → all `(-1,-1)`; and the negative-w case below (would catch the bug).

edgarriba (changes requested):
Thanks Inchara — the remap primitive is genuinely wanted (it's the right base for undistortion and calibrated-rig rectification, which the fused warps can't do), and the two review rounds show. Unfortunately the ground moved underneath this PR: since it was opened, #1001–#1011 landed and they change both what this code can build on and what "done" looks like for a CUDA kernel in this repo. Requesting changes with, I hope, enough specifics to make the rework mechanical.

## 1. Does not compile against main (verified by merge-build)

`use super::texture::CudaTexObject` — `cuda/texture.rs` was **deleted in #1001**. The pitch-2D texture path required `src_width * 3 * 4` bytes to be a multiple of the 32-byte texture pitch alignment, so it failed with `CUDA_ERROR_INVALID_VALUE` for every source width not divisible by 8 (issue #1000). As written, both remap kernels inherit that bug.

**Fix direction:** mirror what the warp kernels became in #1001/#1004 — read the source via `__ldg` on the raw pointer with an explicit border guard (see `warp_perspective_bilinear_3c` in `cuda/warp_perspective.rs`, which is the closest template: per-pixel arbitrary coordinates, exactly like remap). The maps side of your kernels (`__ldg` on `map_x`/`map_y`) is already the right shape and survives unchanged.

## 2. The repo now has a byte-exact CPU↔GPU contract — and remap has a CPU twin to match

Since #1003/#1004 every geometry kernel is bit-identical to its CPU path (all NVRTC kernels compile with `--fmad=false`; expression shapes are mirrored; parity is asserted with `f32::to_bits`). CPU `remap` (`interpolation/remap.rs`) supports all four interpolation modes since #1009 and dispatches to the same samplers the perspective kernels mirror. Concretely: your bilinear kernel should reproduce `bilinear_interpolation`'s tap/weight expressions (the perspective kernel's inner loop is copy-adaptable), and nearest should be `roundf` + clamp like `nearest_neighbor_interpolation`.

One caveat we owe you here, not the reverse: **CPU `remap` currently has a soundness hole** — it feeds raw user map coordinates into samplers that use `get_unchecked`, so an out-of-bounds map value is UB on the CPU. That means the border contract your kernel should match isn't fully defined yet. I'd suggest: kernel writes 0 for maps outside `[0,src)` (your current semantics — correct), and we fix the CPU side to guard the same way in a separate PR. Happy to take that CPU fix myself.

## 3. Integration expectation changed: kernels shouldn't be launcher-only anymore

#1006/#1007 added residency dispatch: `resize`/`warp_affine`/`warp_perspective` route device images to CUDA automatically. Remap should join: a device prologue in `interpolation::remap` via `cuda::dispatch::pair_residency`, with the maps accepted as device tensors (uploading per call would dominate — the maps are as big as the output). That also gets Python support nearly for free via the #1008 pattern.

## 4. Duplicated (and weaker) homography inversion

`remap_maps_from_homography` hand-rolls adjugate/determinant with a raw `det.abs() < 1e-10` test — this duplicates `crate::warp::invert_homography` (which your own #999 added!), and drops its scale-invariant singularity check (`det(kH) = k³·det(H)`, so a raw ε rejects valid small-scaled homographies — the exact issue from the #999 review). Please reuse the shared helper. `remap_maps_from_affine` already reuses `invert_affine_transform` correctly.

## 5. Consider moving the map-from-matrix helpers into the example

Your own benchmark (and #999's) established that the fused warp kernels beat remap-with-matrix-maps. Post-#1007 those fused paths are one call away for users. The library value of remap is **arbitrary** maps (undistortion); `remap_maps_from_affine/homography` mainly serve the architecture-decision benchmark, so I'd keep the library surface to the pure remap launchers + dispatch and move the helpers into `bench_cuda_remap.rs`.

## Summary

The kernel core is sound and the maps design is right — the work is: rebase onto main, swap texture for `__ldg` (perspective kernel as template), match the CPU sampler expressions for byte-exactness (add a `to_bits` parity test like `cuda/warp_perspective.rs`'s), wire the residency dispatch, and reuse `invert_homography`. The 64-bit index math from #1004 (`(y*w + x)*3ull`) should come along too. If it helps, I can push the texture→`__ldg` mechanical rewrite to your branch and leave the dispatch integration to you — say the word.

edgarriba:
One correction to my review: disregard the offers at the end — this is your feature to land, including the texture→`__ldg` rewrite and the CPU `remap` bounds fix if you want it (it pairs naturally with defining your kernel's border contract). The pointers stand: `warp_perspective_bilinear_3c` in `cuda/warp_perspective.rs` is the template for the kernel shape, `cuda/warp_perspective.rs`'s test module for the parity tests, and #1007's adapters for the dispatch wiring. Ping me on the PR if any of the post-#1001 conventions are unclear — happy to answer questions.

Incharajayaram:
@edgarriba rebased on main

crates/kornia-imgproc/src/calibration/distortion.rs
```diff
@@ -138,15 +138,15 @@ pub fn generate_correction_map_polynomial(
...
     let (dst_rows, dst_cols) = (size.height, size.width);
-    let (map_x, map_y) = meshgrid_from_fn(dst_cols, dst_rows, |x, y| {
+    let (t_x, t_y) = meshgrid_from_fn(dst_cols, dst_rows, |x, y| {
         let (xdst, ydst) = distort_point_polynomial(x as f64, y as f64, intrinsic, distortion);
         Ok((xdst as f32, ydst as f32))
-    })?;
-
+    })
+    .map_err(|e| kornia_image::ImageError::Cuda(e.to_string()))?;
+    let map_x = Image::<f32, 1>::new(*size, t_x.as_slice().to_vec())?;
```
edgarriba: to_vec is copying -- we should create the maps with zero-copy

crates/kornia-imgproc/src/interpolation/cuda.rs
```diff
@@ -0,0 +1,170 @@
...
+
+    fn identity_maps(w: usize, h: usize) -> (Image<f32, 1>, Image<f32, 1>) {
+        let size = ImageSize {
+            width: w,
+            height: h,
+        };
+        let mx: Vec<f32> = (0..h).flat_map(|_| (0..w).map(|x| x as f32)).collect();
+        let my: Vec<f32> = (0..h).flat_map(|y| (0..w).map(move |_| y as f32)).collect();
+        (
+            Image::<f32, 1>::new(size, mx).unwrap(),
```
edgarriba: in tests we should also propagate errors ; no unwraps; add to the skills/claude.md

crates/kornia-imgproc/src/interpolation/remap.rs
```diff
@@ -3,57 +3,90 @@ use crate::parallel;
...
+    #[cfg(feature = "cuda")]
+    {
+        use crate::cuda::dispatch::{is_device, pair_residency, Residency};
+        if let Residency::Device(exec) = pair_residency(src, dst)? {
+            if !is_device(map_x) || !is_device(map_y) {
+                return Err(ImageError::Cuda(
+                    "remap: map_x and map_y must be device-resident when src/dst are on GPU".into(),
+                ));
+            }
+            let mx = map_x.0.as_cudaslice().ok_or_else(|| {
```
edgarriba: it doesn't work just mal_x.as_cudaslice() ; otherwise it should be implemented

edgarriba (changes requested):

crates/kornia-imgproc/src/calibration/distortion.rs
```diff
@@ -138,15 +138,15 @@ pub fn generate_correction_map_polynomial(
     _new_intrinsic: &CameraIntrinsic,
     distortion: &PolynomialDistortion,
     size: &ImageSize,
-) -> Result<(CpuTensor2<f32>, CpuTensor2<f32>), TensorError> {
-    //// create a grid of x and y coordinates for the output image
```
edgarriba, line 142: dont remove comments

crates/kornia-imgproc/src/calibration/distortion.rs
```diff
@@ -138,15 +138,15 @@ pub fn generate_correction_map_polynomial(
...
-    //// and interpolate the values from the input image.
+) -> Result<(Image<f32, 1>, Image<f32, 1>), kornia_image::ImageError> {
```
edgarriba:
```suggestion
) -> Result<(Image<f32, 1>, Image<f32, 1>), ImageError> {
```

crates/kornia-imgproc/src/calibration/distortion.rs
```diff
@@ -626,7 +626,7 @@ mod tests {
     }
 
     #[test]
-    fn test_undistort_rectify_map_polynomial() -> Result<(), TensorError> {
+    fn test_undistort_rectify_map_polynomial() -> Result<(), kornia_image::ImageError> {
```
edgarriba:
```suggestion
    fn test_undistort_rectify_map_polynomial() -> Result<(), ImageError> {
```

crates/kornia-imgproc/src/interpolation/cuda.rs
```diff
@@ -0,0 +1,140 @@
...
+/// Run the CUDA remap for a device-resident f32 triple (src, dst, maps).
+///
+/// `map_x` and `map_y` are device slices of length `dst_h * dst_w` — the
+/// same size as a flattened single-channel output image.  Bilinear and nearest
+/// are hardware-accelerated; bicubic and lanczos must be handled by the CPU
+/// path (the caller, `interpolation::remap`, falls through for those modes).
+pub(super) fn remap_f32_cuda<const C: usize>(
+    src: &Image<f32, C>,
+    dst: &mut Image<f32, C>,
+    map_x: &CudaSlice<f32>,
```
edgarriba: shouldn't be map_x: &Image<f32, 1>

edgarriba: same for map_y

crates/kornia-imgproc/src/interpolation/cuda.rs
```diff
@@ -0,0 +1,140 @@
...
+use crate::cuda::remap::{launch_remap_bilinear_cuda, launch_remap_nearest_cuda};
+use crate::interpolation::InterpolationMode;
+
+/// Run the CUDA remap for a device-resident f32 triple (src, dst, maps).
+///
+/// `map_x` and `map_y` are device slices of length `dst_h * dst_w` — the
+/// same size as a flattened single-channel output image.  Bilinear and nearest
+/// are hardware-accelerated; bicubic and lanczos must be handled by the CPU
+/// path (the caller, `interpolation::remap`, falls through for those modes).
+pub(super) fn remap_f32_cuda<const C: usize>(
```
edgarriba: this function shouldnt be in remap.rs ?

crates/kornia-imgproc/src/interpolation/mod.rs
```diff
@@ -1,5 +1,7 @@
 mod bicubic;
 mod bilinear;
+#[cfg(feature = "cuda")]
+pub(crate) mod cuda;
```
edgarriba: this seems wrong here

edgarriba (changes requested):

crates/kornia-imgproc/src/cuda/remap.rs
```diff
@@ -0,0 +1,492 @@
...
+                "{w}x{h} {mode:?}: element {i}: cpu {c} ({:#010x}) gpu {g} ({:#010x})",
+                c.to_bits(),
+                g.to_bits()
+            );
+        }
+    }
+
+    /// Identity map: GPU output must reproduce the source exactly and match CPU.
+    #[test]
+    #[ignore = "requires CUDA"]
```
edgarriba, line 434: We shouldn't ignore but be under a feature flag with cuda

kornia-py/src/resize.rs
```diff
@@ -1,6 +1,8 @@
 use pyo3::prelude::*;
 
 use crate::dispatch::cpu_op;
+#[cfg(feature = "cuda")]
```
edgarriba, line 4: I don't think this is correct as the image api should support both cpu and cuda

edgarriba (approved):
Can you explore next a remap function for u8 images? I guess some sort if quantisation during the interpolation might need to be applied

PR #999, opened 2026-07-11
feat(imgproc): add CUDA warp-perspective kernels (bilinear, nearest, bicubic)

Incharajayaram:
## 📝 Description

**Fixes/Relates to:** 

Adds GPU-accelerated warp-perspective (homography transform) kernels alongside the existing warp-affine kernels. Implements inverse-mapping with a homogeneous w-divide: each output thread computes the source coordinate from the inverted 3×3 homography. Architecture decision: implemented as separate fused kernels (not via remap) after benchmarking showed remap NN is 30–51% slower than fused affine.

---

## 🛠️ Changes Made

- [x] `crates/kornia-imgproc/src/cuda/warp_perspective.rs` — new module with:
  - `launch_warp_perspective_bilinear_cuda` — texture-backed bilinear, 3-ch f32
  - `launch_warp_perspective_nearest_cuda` — texture-backed nearest-neighbor, 3-ch f32
  - `launch_warp_perspective_bicubic_cuda` — Keys cubic a=−0.5 via `__ldg`, 3-ch f32
  - `CudaWarpPerspectiveError` with `Cuda`, `SliceTooSmall`, and `SingularHomography` variants
  - Zero-dimension guard + `check_image_slice` for src and dst in all launchers
- [x] `crates/kornia-imgproc/src/warp/perspective.rs` — expose `pub(crate) invert_homography` reusing existing `adjugate3x3` / `determinant3x3` helpers
- [x] `crates/kornia-imgproc/src/warp/mod.rs` — re-export `invert_homography`
- [x] `crates/kornia-imgproc/src/cuda/mod.rs` — register `pub mod warp_perspective`
- [x] `crates/kornia-imgproc/examples/bench_cuda_warp_perspective.rs` — benchmark all three kernels + affine-BL reference

---

## 🧪 How Was This Tested?

- [x] **Build:** `cargo check -p kornia-imgproc --features cuda` — clean
- [x] **Lint:** `cargo clippy -p kornia-imgproc --features cuda` — no warnings
- [x] **Format:** `cargo fmt -p kornia-imgproc` — applied
- [x] **Manual Verification:** Benchmark ran successfully on GTX 1650 (perspective-tilt homography, 200 iters)

| case | bilinear | nearest | bicubic | affine-BL (ref) |
|------|----------|---------|---------|-----------------|
| 512×512 | 0.073 ms | 0.048 ms | 0.079 ms | 0.076 ms |
| 1280×720 | 0.262 ms | 0.157 ms | 0.299 ms | 0.279 ms |
| 1920×1080 | 0.563 ms | 0.350 ms | 0.723 ms | 0.574 ms |
| 3840×2160 | 2.219 ms | 1.390 ms | 2.943 ms | 2.185 ms |

Bilinear overhead vs warp-affine: 1–4% (the w-divide is the only extra cost per pixel). The texture path is otherwise identical.

- [x] **Performance/Edge Cases:**
  - Zero-dimension guard returns `CudaWarpPerspectiveError::Cuda` before any GPU work
  - Singular homography (|det| < 1e-10) returns `SingularHomography` error before kernel launch
  - Degenerate w per pixel (|w| < 1e-10) writes 0 output in kernel without divergent OOB branch
  - Src slice size validated before texture creation

---

## 🕵️ AI Usage Disclosure

---

## 🚦 Checklist

---

## 💭 Additional Context

**Architecture decision:** Warp-perspective is implemented as separate fused kernels (not via remap base) because the remap benchmark showed nearest-neighbor remap is 30–51% slower than fused affine due to the extra `__ldg` map reads. The fused perspective kernel adds only a single FP reciprocal (`rw = 1/w`) vs the affine kernel — essentially free on modern CUDA hardware.

**Bicubic note:** Bicubic uses raw `__ldg` on the source pointer (4×4 tap window). OOB taps within the neighbourhood are clamped to the source border (BORDER_REPLICATE); pixels whose inverse-mapped centre falls outside the source are written as 0 (BORDER_CONSTANT for the output pixel).

### DIFF
--- a/crates/kornia-imgproc/src/cuda/warp_perspective.rs
+++ b/crates/kornia-imgproc/src/cuda/warp_perspective.rs
@@ -395,6 +395,9 @@ fn make_tex(
     src_width: u32,
     src_height: u32,
 ) -> Result<(CudaTexObject, u64), CudaWarpPerspectiveError> {
+    // Safety: DevicePtr::device_ptr returns the raw CUdeviceptr; the _guard
+    // records a stream event on drop (after the texture is destroyed) which is
+    // the correct ordering.
     let (dev_ptr, _guard) = src.device_ptr(stream);
     let tex =
         CudaTexObject::new_pitch2d_border(dev_ptr, src_width as usize * 3, src_height as usize)
@@ -403,6 +406,7 @@ fn make_tex(
     Ok((tex, handle))
 }
 
+#[allow(clippy::too_many_arguments)]
 fn validate_and_invert(
     src: &CudaSlice<f32>,
     dst: &CudaSlice<f32>,
@@ -418,7 +422,7 @@ fn validate_and_invert(
             "image dimensions must be non-zero".into(),
         ));
     }
-    if block_dim.map_or(false, |(bw, bh)| bw == 0 || bh == 0) {
+    if block_dim.is_some_and(|(bw, bh)| bw == 0 || bh == 0) {
         return Err(CudaWarpPerspectiveError::Cuda(
             "block_dim components must be non-zero".into(),
         ));
--- a/crates/kornia-imgproc/src/cuda/warp_perspective.rs
+++ b/crates/kornia-imgproc/src/cuda/warp_perspective.rs
@@ -403,6 +403,31 @@ fn make_tex(
     Ok((tex, handle))
 }
 
+fn validate_and_invert(
+    src: &CudaSlice<f32>,
+    dst: &CudaSlice<f32>,
+    src_width: u32,
+    src_height: u32,
+    dst_width: u32,
+    dst_height: u32,
+    h: &[f32; 9],
+    block_dim: Option<(u32, u32)>,
+) -> Result<[f32; 9], CudaWarpPerspectiveError> {
+    if src_width == 0 || src_height == 0 || dst_width == 0 || dst_height == 0 {
+        return Err(CudaWarpPerspectiveError::Cuda(
+            "image dimensions must be non-zero".into(),
+        ));
+    }
+    if block_dim.map_or(false, |(bw, bh)| bw == 0 || bh == 0) {
+        return Err(CudaWarpPerspectiveError::Cuda(
+            "block_dim components must be non-zero".into(),
+        ));
+    }
+    check_image_slice(src, "src", src_width, src_height)?;
+    check_image_slice(dst, "dst", dst_width, dst_height)?;
+    invert_homography(h).ok_or(CudaWarpPerspectiveError::SingularHomography)
+}
+
 // ── Public launchers ──────────────────────────────────────────────────────────
 
 /// Launch the bilinear warp-perspective kernel for a 3-channel f32 image.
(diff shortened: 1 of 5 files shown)


crates/kornia-imgproc/src/warp/perspective.rs
```diff
@@ -29,6 +29,25 @@ fn adjugate3x3(m: &[f32; 9]) -> [f32; 9] {
...
 }
 
+/// Invert a 3×3 homography matrix.
+///
+/// Returns `None` for degenerate matrices (|det| < 1e-10).  The CUDA
+/// warp-perspective kernel writes zeros for any output pixel whose w-coordinate
+/// falls below the same threshold.
+pub(crate) fn invert_homography(m: &[f32; 9]) -> Option<[f32; 9]> {
+    let det = determinant3x3(m);
+    if det.abs() < 1e-10 {
```
edgarriba: **CPU/GPU threshold divergence + duplication.** This new `invert_homography` rejects `det.abs() < 1e-10`, but the existing `inverse_perspective_matrix` (used by the CPU `warp_perspective`) rejects only `det == 0.0`. Consequences:

1. A near-singular **or simply small-magnitude** homography is now *accepted on CPU but rejected on GPU* as `SingularHomography` — `det` is scale-sensitive, so an unnormalized H with small entries can have `det ≈ 1e-12` while being perfectly invertible.
2. Two near-identical inverse routines now sit in the same file.

Suggest collapsing to one: have `inverse_perspective_matrix` delegate to `invert_homography(...).ok_or(ImageError::CannotComputeDeterminant)`, and normalize H by `h[8]` before the det check so the threshold is scale-invariant. Fixes the divergence and the duplication together.

crates/kornia-imgproc/src/cuda/warp_perspective.rs
```diff
@@ -0,0 +1,684 @@
+//! Native CUDA warp-perspective kernels for `kornia-imgproc`.
+//!
+//! # Algorithm
+//!
+//! Both kernels use **inverse mapping**: each output thread computes the
```
edgarriba: Doc says "**Both** kernels use inverse mapping" but four kernels ship (bilinear, nearest, bicubic, lanczos — per the Public API list below). Reword to "All kernels". Minor: `make_tex` here drops `_guard` without the `// Safety:` comment its `warp_affine::make_tex` twin carries — worth copying for the next reader.

edgarriba (commented):
Automated review pass. Kernels are correct and closely mirror `warp_affine.rs` — nice work. Main item is a CPU/GPU divergence in the two homography-inverse routines (inline). Plus doc/description drift and no unit tests for the new pure-CPU helper.

**Not inlined — negative-w:** none of the four kernels cull `w < 0` (back-projected) pixels; they divide and rely on border/bounds. This *matches* the CPU `transform_point`, so it's consistent, not a regression — noting only that both differ from OpenCV (which culls `w <= 0`).

**Not inlined — tests:** `invert_homography` is pure CPU. Add a `#[cfg(test)]` mod: known H → known inverse, singular → `None`, round-trip `H·H⁻¹ ≈ I`, and a small-magnitude-det H (would surface the threshold issue below).

**Not inlined — description:** title says "(bilinear, nearest, bicubic)" but a 4th Lanczos-3 kernel + bench ship too — update the PR description.

edgarriba (approved):
Verified on Jetson Orin against the CPU `warp_perspective` reference: interior pixels match exactly (bilinear max abs diff 1.9e-5, nearest 0.0). Remaining differences are confined to pixels whose inverse-mapped source coordinate falls outside the source image, which is the documented BORDER_CONSTANT vs CPU border-handling difference and is consistent with the existing warp_affine kernels.

The scale-invariant singularity test in `invert_homography` (normalizing det by |h[8]|^3) is a nice fix — it also tightens the CPU path, which previously only rejected `det == 0.0`.

Note for follow-up (pre-existing, not introduced here): the pitch2d texture path requires `src_width * 3 * 4` bytes to be a multiple of the device texture pitch alignment, so bilinear/nearest fail with `cuTexObjectCreate ... CUDA_ERROR_INVALID_VALUE` for any width not a multiple of 8. `warp_affine` on main has the same limitation; the benchmarks miss it because all benched widths are multiples of 8. I'll open a separate issue.

PR #1001, opened 2026-07-14
fix(imgproc): CUDA warp works at any image width; fix CPU edge-column off-by-one

### LINKED ISSUE #1000: bug(imgproc): CUDA warp bilinear/nearest fail for image widths not a multiple of 8 (texture pitch alignment)
## Summary

The `pitch2d` CUDA texture path used by the warp kernels only works when the source image width is a multiple of 8. For any other width, `cuTexObjectCreate` fails with `CUDA_ERROR_INVALID_VALUE` and the launcher returns an error, so **bilinear and nearest warping is simply unavailable for most image widths**.

This affects `warp_affine` (pre-existing) and `warp_perspective` (inherits the same helper, merged in #999). The bicubic and Lanczos kernels are unaffected — they read the source through `__ldg` on a raw pointer rather than a texture object.

## Root cause

`CudaTexObject::new_pitch2d_border` (`crates/kornia-imgproc/src/cuda/texture.rs:41`) sets:

```rust
pitchInBytes: width * core::mem::size_of::<f32>(),
```

and both warp modules call it with the interleaved element width:

```rust
// crates/kornia-imgproc/src/cuda/warp_affine.rs:401  (and the same line in cuda/warp_perspective.rs)
CudaTexObject::new_pitch2d_border(dev_ptr, src_width as usize * 3, src_height as usize)
```

So `pitchInBytes = src_width * 3 * 4`. CUDA requires `pitchInBytes` to be a multiple of `CU_DEVICE_ATTRIBUTE_TEXTURE_PITCH_ALIGNMENT`, which is **32 bytes** on the Jetson Orin I tested (and on most current devices). `src_width * 12` is a multiple of 32 only when `src_width` is a multiple of 8, so every other width is rejected.

The base pointer also matters: `CU_DEVICE_ATTRIBUTE_TEXTURE_ALIGNMENT` is 512 bytes on the same device. A fresh `alloc`/`clone_htod` allocation satisfies that, but a texture built over a **suballocated or offset** device slice would fail for the same reason, so any fix should validate the pointer alignment too rather than assuming it.

## Reproduction

On a Jetson Orin (`main` @ 0e73e08a), calling the launcher directly with a 64-row image and varying widths:

```
warp_affine bilinear w=128: Ok
warp_affine bilinear w=129: Err(Cuda("cuTexObjectCreate failed: CUDA_ERROR_INVALID_VALUE"))
warp_affine bilinear w=130: Err(Cuda("cuTexObjectCreate failed: CUDA_ERROR_INVALID_VALUE"))
warp_affine bilinear w=132: Err(Cuda("cuTexObjectCreate failed: CUDA_ERROR_INVALID_VALUE"))
warp_affine bilinear w=136: Ok
```

`warp_perspective` bilinear/nearest fail identically (found while validating #999 at 129x97).

## Why CI and the benchmarks miss it

`examples/bench_cuda_warp_affine.rs` and `examples/bench_cuda_warp_perspective.rs` only exercise 512x512, 1280x720, 1920x1080 and 3840x2160 — every one of those widths is a multiple of 8. There is no GPU correctness test for either kernel, and the CUDA CI job only builds (`Check - CUDA features`), so nothing ever launches these kernels against an unaligned width.

## Suggested fix

Some options, roughly in order of preference:

1. **Pad the texture pitch.** Round `pitchInBytes` up to the device's `CU_DEVICE_ATTRIBUTE_TEXTURE_PITCH_ALIGNMENT` (query it, don't hardcode 32) and pass the padded row stride into the kernels as a separate `src_pitch_elems` parameter, so the index math uses the padded stride instead of `src_w * 3`. This requires the source rows to actually be padded in memory, i.e. a pitched allocation.
2. **Fall back to the `__ldg` path** when the pitch is not aligned. Bilinear/nearest can be computed from raw pointers exactly like bicubic/Lanczos already do; correctness is preserved and only the texture-cache speedup is lost on unaligned widths. Cheapest correct fix.
3. **At minimum, fail loudly and early** with a dedicated error variant that names the constraint, instead of surfacing a bare `CUDA_ERROR_INVALID_VALUE` from `cuTexObjectCreate`.

Whichever route: please also add a GPU test that sweeps a few deliberately unaligned sizes (e.g. 129x97, 130x64, 133x101) and compares against the CPU `warp_affine` / `warp_perspective` reference on the interior region. That test is what would have caught this.

@Incharajayaram would you be up for taking this one, since it is adjacent to the warp work you just landed in #999? Happy to help scope it further.

edgarriba:
## 📝 Description

**Fixes #1000.**

The CUDA warp bilinear/nearest kernels failed outright for most image widths, and fixing that surfaced a real CPU bug. Both are fixed here.

### 1. #1000 — warp fails for any width not a multiple of 8

`warp_affine` and `warp_perspective` read the source through a 1-channel pitch-2D texture object, giving `pitchInBytes = src_w * 3 * 4`. CUDA requires that to be a multiple of `CU_DEVICE_ATTRIBUTE_TEXTURE_PITCH_ALIGNMENT` (32 B on Orin), so `cuTexObjectCreate` returned `CUDA_ERROR_INVALID_VALUE` for every width where `src_w % 8 != 0`:

```
w=128: Ok    w=129: Err(CUDA_ERROR_INVALID_VALUE)    w=130: Err    w=132: Err    w=136: Ok
```

The benchmarks missed it because 512/1280/1920/3840 are all 8-aligned, and these kernels had **no tests**.

**Fix:** rewrite bilinear + nearest to read the source via `__ldg` on a raw pointer — exactly what the bicubic and Lanczos kernels in the same modules already do — and delete `cuda/texture.rs`.

Why remove rather than pad the pitch: the source rows are densely packed, so an aligned pitch needs a padded device-to-device copy on every call, which would cost more than the texture ever won. `cuda/resize.rs`'s own module docs report this same pitch-2D trick regressed ~10% vs `__ldg` for interleaved RGB. Measured on Orin: **no throughput regression**.

### 2. Border rules now actually match the CPU

The texture returned 0 for taps falling off the right/bottom edge, which is not what either CPU op does — and the two CPU ops don't even agree with each other. Each kernel now matches its own reference:

- `warp_affine` (`warp/affine.rs`) clamps the coord to `[0, w-1]` and takes `x1 = min(x0+1, w-1)` — per-axis edge clamp.
- `warp_perspective` (`warp/perspective.rs`) goes through `bilinear_interpolation`, which replicates **`val00`** when either axis overflows.

Both launchers now also validate the `src` slice length, which raw-pointer reads require and the affine ones previously omitted (only `dst` was checked).

### 3. CPU bug: a zeroed column at the exact edge

Comparing kernels against the CPU exposed this. `compute_range` derives the valid column span analytically, but the inclusive/strict ends of the interval **swap with the sign of the step**, and the code applied `ceil` to both:

```
d > 0:  x >= -s0/d (incl)  and  x <  (upper-s0)/d (strict)
d < 0:  x <= -s0/d (incl)  and  x >  (upper-s0)/d (strict)   <-- ends swap
```

When the boundary landed on an exact integer, the `d < 0` case dropped the valid column — **zero-filling pixels whose source coordinate sits exactly on the top/left edge**. A plain horizontal flip hits this on every row. Present in both `warp_affine` and `warp_affine_u8`; fixed in both.

---

## 🧪 How Was This Tested?

- [x] **New GPU tests** (`cuda/warp_affine.rs`): CPU/GPU parity for bilinear + nearest, including a width sweep over `128/129/130/132/133/135/136` — the #1000 regression, red before this change.
- [x] **New CPU test** (`warp/affine.rs`): pins the edge-column rule via a horizontal flip; fails without the `compute_range` fix and needs no GPU.
- [x] Parity uses a transform whose *inverse* is dyadic, so CPU and GPU compute bit-identical coordinates. That makes **nearest bit-exact** rather than tolerance-based, and the test asserts that premise (`dyadic_m_inverse_is_exact`) so it can't silently rot.
- [x] `cargo test -p kornia-imgproc --features cuda` — 327 lib + 51 integration, all green, no CPU regressions.
- [x] `cargo clippy --features cuda --all-targets` — clean.

## 💭 Notes for reviewers

- The remaining CPU/GPU divergence is characterized, not papered over: the CPU walks a row accumulating `sx += dsx` while each GPU thread evaluates the map directly, so they differ by an ulp. Under a non-dyadic transform that can flip an exact `.5` rounding tie or a validity-boundary decision. The tests exclude only those provably-ambiguous pixels and assert the count stays small; everything else is strict.
- The CPU edge fix changes `warp_affine` / `warp_affine_u8` output for transforms with an exactly-integer edge boundary (flips, 90° rotations, integer translations): those columns now carry source pixels instead of zeros. That is the intended, correct behavior.

## 🕵️ AI Usage Disclosure

- [x] 🟡 **AI-assisted:** authored with Claude Code; every kernel change verified on-device against the CPU reference.


edgarriba:
Applied the findings from a full multi-agent review of this branch (10 finder angles + adversarial verification + gap sweep), commit fd97c9f0:

**Correctness**
- The same inclusive/strict boundary off-by-one this PR fixed in `warp_affine` was still alive in `warp_perspective_u8` (`apply_ge`/`apply_lt` + `ceil` on both ends). The rounding rules now live in one place — `warp::span::constrain_span`, unit-tested — and `warp_affine`, `warp_affine_u8`, and `warp_perspective_u8` all use it. A u8 horizontal-flip regression test pins the perspective fix.
- `bicubic`/`lanczos` launchers read `src` via raw `__ldg` but never validated its length (OOB device read on a short slice). All four affine launchers now share one `validate()` covering geometry (zero dims, zero `block_dim` components — previously a `div_ceil(0)` panic) and both slice lengths.
- `bilinear`/`nearest` used a panicking NVRTC compile helper while their docs promised `Err` on compile failure — now on the same `Result` path as the rest.
- All eight kernels now do 64-bit index math: `(y*w + x)*3` in `unsigned int` wraps at ~1.43 gigapixels (~18 GB f32 source, real on 80 GB-class GPUs), silently diverging from the `usize` CPU path.

**Structure/docs**
- `make_config` / `try_compile_with_l1` / `check_geometry` hoisted to `cuda/mod.rs`; `warp_perspective.rs` deduplicated onto them (the same helpers also exist in `resize.rs` — left for the follow-up dispatch refactor to avoid widening this PR).
- Stale docs fixed: the affine bilinear kernel comment and launcher rustdoc described the val00-replicate rule the kernel deliberately does not use; the perspective bench header and runtime note still claimed bilinear/nearest are texture-backed.
- New tests: bicubic identity is asserted bit-exact (Keys weights at frac 0 are exactly `[0,1,0,0]`), lanczos near-identity, and per-launcher short-slice/zero-dim/zero-block error tests.

**Benchmarks (Jetson Orin, this device)**
- Texture vs `__ldg` A/B (main vs branch, same bench): bilinear 1080p 1.116 → 0.813 ms, nearest 1024×896 0.622 → 0.356 ms; the rest within run-to-run spread. The texture path was not faster on this hardware, consistent with the ~10% regression `cuda/resize.rs` already documented for the same trick.
- After the 64-bit index change: bicubic/lanczos timings identical (2.632/5.808 ms at 1080p), bilinear/nearest within jitter across repeated runs.
- CPU criterion `bench_warp`: within ±4% of main; `warp_affine_u8`/`warp_perspective_u8` spot timings within run spread.

Full suite green: 335 lib + 51 integration tests, clippy clean.

PR #1009, opened 2026-07-16
feat(imgproc): f32 bicubic + Lanczos-3 across resize/warps/remap, byte-exact CPU/GPU

edgarriba:
## 📝 Description

Unlocks the **last six launcher-only CUDA kernels** (resize/warp-affine/warp-perspective × bicubic/Lanczos-3) through the public API, with matching CPU f32 implementations — `f32::to_bits` parity asserted end-to-end through the public functions for rotations, projective homographies, and non-dyadic resizes. `remap` gets the new modes too.

### Three portability traps found and settled
1. **`sinf` is not portable.** The Lanczos kernels used `__sinf`/`__fdividef` (no host equivalent), and even precise `sinf` rounds differently between the CUDA math library and glibc. Both sides now share a deterministic `sin(πx)`: exact integer reduction + odd Taylor polynomial in plain mul/add — bit-identical under `--fmad=false`. (Naive precise-`sinf` had cost 1.7–1.9× on the separable resize kernels.)
2. **Resize-lanczos weights are per-axis**, so `lanczos_axis` builds the tap/weight tables **once on the host** — the same call feeds the CPU separable pass and the CUDA launcher upload. The kernels become pure gather+FMA: **flat on downscale, 25–27% faster than the old `__sinf` baseline on upscale** (1080p→4K 4.28→3.11 ms).
3. **The warp bicubic kernels still had pre-`fmad=false` plain-op Horner weights** — a 6-ulp parity failure pinned it; they now use explicit `fmaf` matching the CPU's `mul_add`.

### Perf guardrail finding (and a net win)
Putting the now-four-way `interpolate_pixel_fast` inside per-pixel loops deoptimized every hot path (+25–67% CPU resize, +35–55% warps — inline hints didn't save it). Every per-pixel call site now hoists the mode dispatch into one monomorphic loop per mode, which ends up **faster than before this branch**: CPU f32 resize **~30% faster** (1024×896: 1.18→1.02 ms), warp-perspective ~7% faster, warp-affine within jitter.

Also: the warp bicubic/lanczos kernels adopt the same degenerate-axis validity rule as bilinear/nearest (right-angle rotations agree across all modes), and `validate_interpolation` accepts all four modes.

## 🧪 How Was This Tested?

- [x] Jetson Orin, `--features cuda`: 355 lib + 51 integration green; clippy clean; CPU-only build clean.
- [x] New bit-exact parity tests through the **public** functions: resize bicubic/lanczos (up+down, non-dyadic), warp-affine + warp-perspective bicubic/lanczos (37° rotation, projective homography).
- [x] Benchmarks (same-session A/B): GPU lanczos-resize numbers above; GPU bicubic/warp flat; CPU resize/warp numbers above; the four flipped "unsupported interpolation" tests now assert support.

## 🕵️ AI Usage Disclosure

- [x] 🟡 **AI-assisted:** authored with Claude Code; parity and benchmarks verified on-device.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,18 @@ changes early: `cargo add kornia-imgproc@0.1.15-rc.1` or `pip install --pre korn
 
 ## [Unreleased]
 
+**Bicubic and Lanczos-3 for the f32 geometry ops, byte-exact CPU↔GPU.**
+`resize`, `warp_affine`, `warp_perspective`, and `remap` accept
+`InterpolationMode::Bicubic` and `::Lanczos` on the CPU, and device images
+route to the previously launcher-only CUDA kernels — all six now reachable
+from the public API, with bit-identical output between backends. Lanczos
+weights use a shared deterministic sin(πx) polynomial (libm/CUDA `sinf`
+rounding differs between platforms); resize-lanczos builds its per-axis weight
+tables on the host once per call, shared verbatim with the GPU launcher —
+which also made the GPU kernels up to 27% faster than the old `__sinf` path.
+Hot-loop interpolation dispatch is hoisted per mode, making CPU f32 resize
+~30% faster and warp-perspective ~7% faster than before.
+
 **Python: GPU resize and warps.** `kornia_rs.imgproc.resize` / `warp_affine` /
 `warp_perspective` now accept a device `Image` (f32, 3-channel) and run the
 CUDA kernels, returning a device `Image` — output bit-identical to the CPU f32
--- a/crates/kornia-imgproc/src/cuda/resize.rs
+++ b/crates/kornia-imgproc/src/cuda/resize.rs
@@ -254,9 +254,10 @@ extern "C" __global__ void resize_bicubic_3c(
     unsigned int dst_y = blockIdx.y * blockDim.y + threadIdx.y;
     if (dst_x >= dst_w || dst_y >= dst_h) return;
 
-    // Half-pixel centre alignment (matches OpenCV / PIL convention).
-    float sx = fmaxf(fminf(fmaf(ax, (float)dst_x, bx), (float)(src_w - 1u)), 0.0f);
-    float sy = fmaxf(fminf(fmaf(ay, (float)dst_y, by), (float)(src_h - 1u)), 0.0f);
+    // Plain multiply-add (not fmaf): bit-identical to the CPU LUT's `a*x + b`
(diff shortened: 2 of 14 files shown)


edgarriba:
## Benchmark verification (Jetson Orin, same-minute A/B main↔branch, `fast_image_resize` as untouched control)

**No regressions — several wins.** Control library identical between trees (±5%), so the deltas are real:

| CPU f32 (criterion, 256²/512²/1024²) | main | branch | Δ |
|---|---|---|---|
| resize `kornia_par` | 173/553/1884 µs | **87/286/1291 µs** | **~1.9× faster** (hoisted dispatch) |
| warp-affine | 397/1488/5834 µs | 379/1476/5363 µs | flat/faster |
| warp-perspective | 901/3235/9771 µs | 829/2734/9949 µs | flat/faster |
| `fast_image_resize` (control) | 18.7/42.0/126 µs | 19.8/40.4/121 µs | unchanged ✓ |

**GPU:** lanczos-resize up to 27% faster than the pre-branch `__sinf` kernels (host weight tables); bicubic/warp kernels flat (details in the PR body).

## Cross-library head-to-head (1080p→720p resize / 1080p warp, median of 50, this device)

| Op | kornia | cv2 | PIL | VPI (CUDA) |
|---|---|---|---|---|
| resize u8 bilinear (CPU) | **3.52 ms** | 6.08 ms | 24.2 ms | — |
| warp-affine u8 bilinear (CPU) | **9.82 ms** | 17.2 ms | — | — |
| resize f32 lanczos | **2.6 ms GPU** (≈0.6 ms kernel) | 18.0 ms CPU (LANCZOS4) | 63.3 ms (u8, AA) | — |
| warp f32 bilinear (GPU, steady) | **1.05 ms** | 12.8 ms CPU (f32) | — | 3.69 ms (u8!) |
| resize f32 bilinear (GPU, steady) | **1.08 ms** | 3.76 ms CPU (f32) | — | 1.80 ms (u8!) |

kornia is fastest on every comparable op — the GPU paths beat VPI's CUDA backend while moving 4× the bytes (f32 vs u8).

**One honest caveat, filed as follow-up:** the Python device path allocates a fresh destination per call, and Jetson's `cuMemAlloc`/free serialize badly under a naive per-frame `op(); sync()` pattern (8.5 ms/frame vs 1.1 ms steady-state/pipelined; each free ≈0.33 ms). VPI pools its allocations. Fix direction: `out=` support on the device path or a pooled allocator — orthogonal to this PR (the pattern dates from the color ops).

PR #1012, opened 2026-07-16
feat(imgproc): u8 CUDA resize — all modes, byte-exact with the CPU fast paths

edgarriba:
## What

u8 device images now resize on the GPU, across the **entire** CPU u8 cascade — and the output is **byte-for-byte identical** to the CPU (`assert_eq!`, not tolerances):

| CPU branch | New CUDA kernel |
|---|---|
| pyrdown 2× RGB (box, `(sum+2)>>2`) | `resize_u8_pyrdown2x_rgb` |
| pyrup 2× RGB (nested rounding averages) | `resize_u8_pyrup2x_rgb` |
| nearest (f64 half-pixel floor LUT) | `resize_u8_nearest` (pure gather) |
| generic bilinear Q14 (u64 blend, one Q28 round) | `resize_u8_bilinear` |
| separable bicubic/lanczos Q14, AA + non-AA (i16 intermediate) | `resize_u8_sep_h` + `resize_u8_sep_v` |

**How byte-exactness holds:** the u8 CPU paths precompute coordinates in host f64 and run integer fixed-point pixel loops — no device float expression can mirror that. So the kernels are integer-only, and the coordinate/weight tables are built by the *same host functions the CPU uses* (`bilinear_tap`, `nearest_index`, `precompute_contribs` — extracted as shared single-source builders) and uploaded. CPU and GPU cannot drift by construction.

**Device LUT cache:** nsys showed per-call pageable H2D uploads average ~250 µs on Jetson (8 µs median, 8 ms max tail) — dominating the op. Tables now upload once per (device, mode, antialias, geometry) into a bounded cache; the warm path is pure kernel launches and therefore CUDA-Graph-capturable after one warming call.

Also: Rust `resize_fast_u8_aa` residency dispatch (error-never-fallback, cascade mirrored branch-for-branch); Python `kornia_rs.imgproc.resize` accepts u8 1/3/4-channel device images incl. `out=`; resize launchers no longer panic on NVRTC failure (prep commit consolidates the shadowed compile helpers onto `cuda/mod.rs` — fmad-drift risk).

## Benchmarks (Jetson Orin, 1080p→720p u8 RGB, median, out= reuse)

| op | kornia GPU | VPI-CUDA | cv2 CPU | kornia CPU |
|---|---|---|---|---|
| nearest | **0.22 ms** | 1.18 | 0.70 | 0.39 |
| bilinear | **0.38 ms** | 1.26 | 2.39 | 2.40 |
| bicubic (non-AA, = VPI/cv2 semantics) | **1.20 ms** | 1.32 | 5.30 | 4.78 |
| lanczos AA (no VPI equivalent) | **4.12 ms** | — | 9.33 (non-AA) | 8.14 |

Fastest on every VPI-comparable op. CPU A/B vs main (same-minute, alternating, 2 rounds): all deltas within criterion noise — no regression.

## Tests

- 10 Rust parity tests: all 5 branches × C∈{1,3,4} × AA on/off × odd sizes × extreme downscale (4000→50-class, i32 accumulator bound) — `assert_eq!` on raw bytes.
- Python: u8 device vs numpy CPU bit-exact (4 modes × AA), pyr fast paths, `out=` reuse + dtype-mismatch rejection.
- Known-unrelated: `test_mem_probe_is_sensitive` fails identically with a main-built wheel on this Jetson (environmental `mem_get_info` insensitivity, not this PR — will file separately).

## Follow-ups (not this PR)

- 6+ tap separable kernels scale super-linearly with taps; compile-time tap-count kernel variants are the next lever.
- u8 warps (PR 2 of the roadmap).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,10 +20,15 @@ count the conversion). `resize_fast_u8_aa` (Rust) and `kornia_rs.imgproc.resize`
 integer-only CUDA kernels covering the full CPU cascade: exact-2× pyramid fast
 paths, nearest, Q14 bilinear, and antialiased/plain separable bicubic and
 Lanczos-3. The coordinate and weight tables are built by the same host
-functions the CPU uses and uploaded, so device output is byte-for-byte equal
-to host output (asserted with `assert_eq!` across modes, channel counts, odd
-sizes, and extreme downscales). Kernel compile failures now surface as errors
-instead of panics across all resize launchers.
+functions the CPU uses and uploaded once per geometry (a device-side LUT
+cache — Jetson pageable H2D uploads have a ~250 µs average latency tail that
+otherwise dominates, and the warm path is CUDA-Graph-capturable), so device
+output is byte-for-byte equal to host output (asserted with `assert_eq!`
+across modes, channel counts, odd sizes, and extreme downscales). Measured on
+Jetson Orin at 1080p→720p u8 RGB (median, out= reuse): nearest 0.22 ms,
+bilinear 0.38 ms, bicubic 1.20 ms — faster than VPI-CUDA on every comparable
+op (5.4× / 3.4× / 1.1×) and 3–6× ahead of OpenCV CPU. Kernel compile failures
+now surface as errors instead of panics across all resize launchers.
 
 **Python: `out=` and CUDA Graphs for allocation-free frame loops.** The device
 geometry ops accept a preallocated device `out=` image (torch-style, returned
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,18 @@ changes early: `cargo add kornia-imgproc@0.1.15-rc.1` or `pip install --pre korn
 
 ## [Unreleased]
 
+**u8 CUDA resize — every mode, bit-identical to the CPU fast paths.** Camera
+data is u8; until now the GPU only resized f32 (4× the memory traffic once you
+count the conversion). `resize_fast_u8_aa` (Rust) and `kornia_rs.imgproc.resize`
+(Python, including `out=`) now route u8 device images — 1/3/4-channel — to new
+integer-only CUDA kernels covering the full CPU cascade: exact-2× pyramid fast
+paths, nearest, Q14 bilinear, and antialiased/plain separable bicubic and
+Lanczos-3. The coordinate and weight tables are built by the same host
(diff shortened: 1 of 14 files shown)


edgarriba:
CI note: first run failed on (a) an unconditionally-imported `PyImageApi` breaking the CPU-only clippy build, (b) unformatted kornia-py files, both fixed in the latest push — and (c) `gstreamer::tests::gst_resource_capture_n_frames_and_drop` (kornia-io, untouched by this PR): timing-flaky, expected 5 frames got 4. Benchmarks are in the PR body; the u8 nearest/bilinear/bicubic GPU kernels beat VPI-CUDA on every comparable op.

edgarriba:
## Adversarial review (8 finder angles × 12 verified candidates)

**Core math is clean**: all five kernels traced constant-for-constant against the CPU oracles (Q14 shifts, rounding constants, accumulator widths, launch args) — zero wrong-output bugs found.

**3 confirmed issues fixed in the last commit:**
1. **Cross-stream LUT cache race** — async uploads cached without a stream fence; a same-geometry call on another stream could read half-uploaded tables (silent wrong pixels). Now: sync-before-insert (miss-only) + single-entry eviction instead of clear-all.
2. **Cascade drift** — GPU cascade was a hand-copy of the CPU cascade and had already diverged (1×N bilinear: CPU *panicked*, GPU errored; C=2 nearest: CPU ok, GPU errored; different error variants). Both sides now consume one shared `resize_u8_path` selector — routing + errors are residency-independent by construction. Degenerate bilinear is now a typed `InvalidImageSize` on both sides (was a CPU panic).
3. **Latent memory hazard** — `pub` nearest launcher had no channels check while the catch-all kernel arm hardwired 4 bytes/pixel; C=2 via the public API would corrupt memory. Explicit C==4 arm + per-byte generic arm + launcher guards.

**Refuted by verification** (for the record): 1×N OOB-read claim (launcher already guarded), sticky compile-error memoization (deliberate repo-wide policy), ordinal cache keys (cudarc retains the primary context).

**Deferred as follow-ups**: separable H-pass scratch is per-call allocated (blocks CUDA-graph capture of bicubic/lanczos — candidate for a Plan object), kernel-cache boilerplate ×3 → one generic helper, `out=` ColorSpace metadata asymmetry (pre-existing f32 pattern).

363 lib tests green, clippy `-D warnings` clean, new parity/error tests added.

PR #1014, opened 2026-07-16
feat(imgproc): u8 CUDA morphology (dilate/erode), byte-exact + first Python bindings

edgarriba:
## What

Independent of the resize/warp stack (branches off main). `dilate` / `erode` route u8 device images (1/3/4-channel) to a CUDA kernel that is **byte-for-byte identical** to the CPU ops, plus first-ever Python bindings for morphology.

### Byte-exactness

Morphology is pure integer min/max — commutative — so exactness only requires sampling the same value multiset as the CPU: same active tap offsets (`k_data == 1`), same border index math (all five `PaddingMode`s mirrored: constant/replicate/reflect101/reflect/wrap), same empty-kernel default (0). `assert_eq!` parity tests over box/cross/ellipse + even-sized elements, every border mode, C∈{1,3,4}. Non-u8 device pairs (`u16` etc. — the CPU ops are generic over `Ord`) get a typed no-GPU-kernel error instead of a host read of device pointers.

### Per-structuring-element kernel specialization

Tap offsets are **baked into the NVRTC source as literals** (kernel cached per element+op; elements are static per pipeline, so the one-time ~1s compile amortizes to zero). Three designs measured on Orin, 3×3 box 1080p C1:
- runtime tap loop over a global LUT: serialized dependent loads;
- shared-memory tap staging: *worse* — the 8 KB static allocation cut occupancy and fought the `PREFER_L1` carveout;
- baked straight-line stanzas (shipped): every load issues independently; interior threads skip all border logic.

Border constants cached on-device per (device, value, channels) — same Jetson pageable-H2D-tail rationale as the resize LUT cache (#1012).

### Python

New `kornia_rs.imgproc.dilate` / `erode` (`kernel="box"/"cross"/"ellipse"`, `size=(h,w)`, `border=...`, `constant_value=...`) covering numpy u8 CPU and u8 device images. 14/14 tests.

## Benchmarks (Jetson Orin, 1080p C1 u8, median)

| op | kornia GPU | VPI-CUDA | cv2 CPU | kornia CPU |
|---|---|---|---|---|
| dilate 3×3 | **1.20 ms** | 1.43 | 0.62 | 23.8 |
| erode 3×3 | **1.10 ms** | 1.71 | 0.64 | 25.7 |

Faster than the GPU competitor (VPI-CUDA) on both. Honest note: OpenCV's NEON CPU morphology wins on host-resident 3×3 data — for device-resident pipelines the GPU op avoids the D2H/H2D round trip and composes with graphs. Follow-up lever noted in-code: 4-pixel-per-thread `__vmaxu4` vectorization (nsys shows the kernel is latency-bound on per-byte gathers, ~1.8 ms kernel time).

Also of note: the review found a classic Rust deadlock pattern in an earlier draft — `if let Some(x) = mutex.lock().unwrap().get(k)` holds the guard through the `else` branch; caches here use scoped-guard lookups (comment documents it).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,18 @@ changes early: `cargo add kornia-imgproc@0.1.15-rc.1` or `pip install --pre korn
 
 ## [Unreleased]
 
+**u8 CUDA morphology (dilate / erode), bit-identical to the CPU ops, plus
+first-time Python bindings.** `dilate` / `erode` route u8 device images
+(1/3/4-channel) to a CUDA kernel that samples the exact tap multiset the CPU
+does — same active structuring-element offsets, same border index mapping
+for all five padding modes — so min/max output is byte-for-byte equal
+(`assert_eq!` parity tests over box/cross/ellipse and even-sized kernels).
+The structuring element is cached on-device per geometry (Jetson pageable
+H2D tail; warm path is CUDA-Graph-capturable). New
+`kornia_rs.imgproc.dilate` / `erode` Python functions cover the numpy u8 CPU
+path and u8 device images alike. Non-u8 device pairs error with the typed
+no-GPU-kernel message instead of touching device memory from the CPU.
+
 **u8 CUDA warps (bilinear), bit-identical to the CPU fast paths — which are
 now bit-identical to each other too.** `warp_affine_u8` and
 `warp_perspective_u8` route u8 device images (1/3/4-channel) to new
--- a/crates/kornia-imgproc/src/cuda/mod.rs
+++ b/crates/kornia-imgproc/src/cuda/mod.rs
@@ -28,6 +28,10 @@ pub mod warp_perspective_u8;
 /// Native CUDA color-space conversion kernels.
 pub mod color;
 
+/// Native CUDA u8 morphology kernels (dilate / erode, byte-exact with the
+/// CPU ops).
+pub mod morphology;
+
 /// Residency-aware dispatch machinery shared by all device-capable ops.
 pub(crate) mod dispatch;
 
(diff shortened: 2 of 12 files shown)


edgarriba:
Update: added a 4-pixel-per-thread vector kernel for C1 (aligned u32 loads + `__byte_perm` stream normalization + `__vmaxu4`/`__vminu4`), still byte-exact — parity suite extended with full-HD C1 cases.

**Revised numbers (1080p C1 3×3, median):**

| op | kornia GPU | VPI-CUDA | cv2 CPU (NEON) |
|---|---|---|---|
| dilate | **0.275 ms** | 1.43 | 0.61 |
| erode | **0.260 ms** | 1.71 | 0.61 |

Now 2.3× faster than OpenCV CPU and 5–6.6× faster than VPI-CUDA. The earlier caveat about cv2's CPU path winning is obsolete. (These figures still include per-call output allocation + sync; `out=` plumbing for morphology would push further toward the ~20 µs DRAM floor.)

edgarriba:
Second kernel iteration: **4×4 output tile per thread** with horizontal-fold reuse — each input row is loaded + normalized once and one horizontal `__vmaxu4`/`__vminu4` fold per distinct dx-set serves all four output rows (3×3 box: 6 row loads + 6 folds per 16 outputs vs 36 loads before). Still byte-exact; parity suite unchanged.

**Final numbers (1080p C1 3×3, median, per-frame with sync):**

| op | kornia GPU | VPI-CUDA | cv2 CPU (NEON) | vs VPI | vs cv2 CPU |
|---|---|---|---|---|---|
| dilate | **0.237 ms** | 1.43 | 0.66 | 6.0× | 2.8× |
| erode | **0.209 ms** | 1.71 | 0.62 | 8.2× | 3.0× |

Remaining headroom to the ~40 µs floor (25 µs DRAM + launch) is measurement-mode bound — per-frame python sync + per-call allocation account for a large share at this scale; the PR-15 acceptance harness measures device-compute-only and graph-replay modes where this op should land near 10× on both baselines.

PR #1016, opened 2026-07-17
feat(imgproc): FKL-style kernel fusion engine (register-flow stage composition)

edgarriba:
Implements the fusion architecture discussed from the FKL poster (Amoros/Nuñez/Peña, [Libraries-Openly-Fused](https://github.com/Libraries-Openly-Fused)) as the core of roadmap PR 15.

## Design

- **`FusedStage` trait** = FKL's IOp: a host-side `build()` (user-friendly → kernel-friendly params, packed once) + a device `exec` snippet; the composer generates ONE NVRTC kernel where values flow **through registers** between stages — no intermediate global-memory round trips.
- **`__grid_constant__ const` param blob**: all stages' parameters pack into a single 256-byte POD struct read from the constant bank — zero register cost, warp-wide broadcast, so adding stages doesn't inflate register pressure (the poster's key result). Accessor macros bake literal constant-bank offsets.
- **Kernel cached per pipeline *shape***: the blob is data, not code — same-shape pipelines with different means/scales/dims share one compiled kernel (asserted by test).

## Validation

- Composed `resize→normalize→CHW` matches a CPU reference of the same f32 semantics to 1e-4.
- A **novel** chain (`resize→normalize→gray→C1`) composes from the same stage library with no hand-written kernel anywhere.
- Throughput (Orin, locked clocks): **0.187 ms** for 1080p→640×640 CHW-normalize — parity with the hand-written fused preprocessor (0.202 ms), i.e. **zero abstraction overhead** for the measured-73×-vs-cv2-chain pipeline class.

Fused output is a documented higher-precision contract (f32 register intermediates — same semantics as the existing preprocessor); per-stage requantization to mirror op-by-op u8 sequences bit-exactly is a planned opt-in.

Follow-ups (tracked for the full PR 15): batch-N via `blockIdx.z` divergent chains, more stages (crop/letterbox/pad, YUV decode read-stages, u8 sinks), graph-capture integration, python surface.

### DIFF
--- a/crates/kornia-imgproc/src/cuda/fusion.rs
+++ b/crates/kornia-imgproc/src/cuda/fusion.rs
@@ -1,0 +1,685 @@
+//! FKL-style kernel fusion engine.
+//!
+//! Composes per-op device snippets into ONE NVRTC kernel where data flows
+//! through registers between stages — no intermediate global-memory round
+//! trips. The design follows the Fused Kernel Library model (Amoros /
+//! Nuñez / Peña, <https://github.com/Libraries-Openly-Fused>):
+//!
+//! - each stage splits into a host-side **build** step (user-friendly →
+//!   kernel-friendly parameters, done once) and a device-side **exec**
+//!   snippet (the hot path);
+//! - all stages' kernel-friendly parameters are packed into a single POD
+//!   blob passed as **`__grid_constant__ const`** — parameters read from
+//!   the constant bank cost zero registers and broadcast warp-wide, so the
+//!   register footprint of a chain depends on the exec bodies, not on how
+//!   much state the stages carry;
+//! - the composed kernel is compiled once per (pipeline shape) and cached.
+//!
+//! # Value flow
+//!
+//! v1 pipelines are **per-output-pixel** over a destination grid: a source
+//! stage produces a `float3 v` for the output coordinate `(x, y)`, middle
+//! stages map `v -> v`, and a sink stage writes `v` (or a reduction of it)
+//! to the destination. Intermediates stay in f32 registers; fused output
+//! is therefore NOT bit-equal to running the ops separately through u8
+//! buffers — it is a documented, higher-precision contract of its own
+//! (the same one the existing hand-fused preprocessor uses).
+//!
+//! # Parameter blob layout
+//!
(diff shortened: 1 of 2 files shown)


edgarriba:
Added `blockIdx.z` batching (homogeneous DivergentBatchTransformDPP): `build_batched(ctx, stages, dw, dh, batch, out_elems_per_image)` + `launch_batched(stream, &[srcs], dst)`. Per-image src pointers ride inside the `__grid_constant__` blob and a z-prologue re-points `src`/`dst` — **stage snippets unchanged, so every stage batches for free**. Batch-4 output asserted bit-identical to 4 single launches. Measured: batch-4 1080p→640² CHW-normalize = **0.909 ms/batch (0.227 ms/image)**, one launch → one NCHW tensor (~60× the per-image cv2 CPU chain on the roadmap's acceptance workload).

edgarriba:
## Head-to-head vs the real FKL (FusedKernelLibrary, C++)

Built the same pipeline with upstream FKL on this Orin (nvcc sm_87, C++20): `PerThreadRead<uchar3> → Resize<INTER_LINEAR>(640²) → Mul(1/255) → SplitWrite` single + batch-4 via their batched TransformDPP. Same-minute interleaved A/B, 3 cycles, locked clocks:

| pipeline | FKL C++ | kornia FusedPipeline |
|---|---|---|
| single 1080p→640² CHW-norm | 0.121–0.127 ms | 0.121–0.124 ms |
| batch-4 | 0.465–0.471 ms (0.117 ms/img) | 0.467–0.470 ms (0.117 ms/img) |

**Exact parity.** Both engines emit a register-flow fused single-pass kernel and sit on the same bandwidth envelope. The NVRTC-runtime-codegen approach here costs nothing vs FKL's compile-time templates at runtime — while keeping runtime pipeline composition (no nvcc, no C++ recompile per shape).

Caveats: FKL batch-4 writes interleaved-per-plane (their canonical TensorWrite) rather than strict planar NCHW — byte-count identical, layout differs; single-image is a clean apples-to-apples. FKL needed a `bit_cast` shim under CUDA 12.6 (their headers target 13.x libcu++).

Bench source: scratchpad `bench.cu` (FKL side) + the `probe_fused_1080p`/`probe_batch4_1080p` ignored tests in this PR.

PR #1025, opened 2026-07-18
feat(imgproc): CLAHE (CPU+CUDA), byte-for-byte with OpenCV

edgarriba:
## Summary

PR 7 of the GPU-coverage roadmap: CLAHE (Contrast-Limited Adaptive Histogram Equalization) for u8 single-channel images, CPU + CUDA, byte-for-byte with `cv2.createCLAHE(clip, grid).apply()`.

Mirrored cv2 semantics (from `clahe.cpp`, verified against the 4.13 wheel):
- **Tile padding rule**: non-divisible sizes are (virtually) extended right/bottom with `BORDER_REFLECT_101` — including OpenCV's quirk that a divisible dimension still gets one full extra tile of padding when the other dimension is not divisible.
- **Clip limit** `max(1, trunc(clip · tileArea / 256))`, with `clip ≤ 0` disabling clipping; two-phase excess redistribution (uniform batch + residual at stride `max(256/residual, 1)`).
- **LUT quantization**: f32 scale `255/tileArea`, round-half-to-even (`saturate_cast`).
- **Bilinear tile blend with cv2's compiled FMA contraction.** The scalar C++ in cv2 compiles with GCC's default `-ffp-contract=fast` on aarch64, so the weight computation (`txf = fma(x, 1/tw, −0.5)`) and the blend (`res = fma(i1, ya1, i2·ya)`, `i_k = fma(l_k0, xa1, l_k1·xa)`) use true fused multiply-adds. Calibrated empirically per pixel (candidate-pattern grid search on mismatching pixels); implemented with `f32::mul_add` on CPU and explicit `fmaf` in the CUDA kernel (the global NVRTC `fmad=false` only disables implicit contraction). `f32::mul_add` is exact fma on every platform (libm fallback without an FMA unit), so parity is not aarch64-specific.

CUDA path: `clahe_lut_u8` (one block per tile: shared-memory histogram via commutative atomics, thread-0 sequential clip/redistribute/scan, parallel f32 LUT quantize) + `clahe_apply_u8` (per-pixel 4-LUT blend). Textual twins of the CPU stages; single-source `ClaheGeometry` (tile geometry, integer clip, f32 scales computed once on host).

kornia-py: `imgproc.clahe(image, clip_limit=40.0, grid_size=(8,8))` — numpy CPU path + device `Image` path, stubs, parity/leak tests; `cuda_clahe` module cfg-gated (and the no-feature `cargo check -p kornia-py` run this time).

## Byte parity (through the wheel)

**22/22 configurations maxdiff=0** vs cv2 4.13: 7 shapes (1080p, 480p, 43×67, 64×63, 63×64, 9×33, 17×17 — divisible + both padding paths), 5 grids ((1,1)…(16,16), asymmetric), 5 clip limits (0.5…4000 + clipping-disabled 0.0), plus constant / two-level / narrow-range / zeros / low-contrast-gradient content. GPU output equally exact. Rust: CPU↔GPU `assert_eq!` over 36 size/grid/clip combos; 389 lib tests green.

## Performance (Orin, locked clocks, 1080p sustained)

| impl | time | vs cv2 |
|---|---|---|
| cv2 CPU (8T) | 4.00 ms | 1× |
| kornia CPU | 5.46 ms | 0.7× |
| **kornia GPU** | **0.040 ms** | **~100×** |

VPI has no CLAHE op — no VPI baseline exists for this table.

## ncu (locked clocks, `--set basic`)

- `clahe_lut_u8`: occupancy 69%, 40 regs, SM 46% — 64 blocks (one per tile), atomics + sequential thread-0 scan.
- `clahe_apply_u8`: occupancy 83%, 25 regs, SM 67% — gather-4-LUTs blend, L1-resident LUTs.


edgarriba:
**CPU optimization pass added** (same PR, byte-parity preserved):

| stage | before | after |
|---|---|---|
| LUT build | 1.16 ms | ~0.4 ms (interior tiles: branch-free slice walk, reflect only on padded edges) |
| interpolation | 3.55 ms | ~1.1 ms (packed u32 span tables → 1 gather/px; NEON `vfmaq_f32` 8 px/iter) |
| **total (rust)** | 4.72 ms | **~1.55 ms** |

1080p same-minute A/B through the wheel: **kornia CPU 2.19 ms vs cv2 4.45 ms = 2.0×** (cv2 with 8 threads); GPU 0.033 ms = 134×. Byte parity re-verified after the rewrite: 22/22 configs maxdiff=0 (NEON `vfmaq_f32`/`vcvtnq_s32_f32` are bit-identical to the scalar `mul_add`/`round_ties_even`, and the packed tables store exact LUT bytes).

**10×-vs-cv2 on CPU is not reachable on these cores**: interpolation now runs ~9 cycles/px/core on 8 cores with a single L1 gather per pixel; a 0.4 ms target implies ~3 cycles/px — below the load/store floor for value-dependent gathers (each pixel's table index is its own value; no NEON gather instruction exists). The 10× per-op gate is carried by the GPU path (134×).

PR #1026, opened 2026-07-19
feat(imgproc): median blur + bilateral filter (CPU+CUDA), byte-for-byte with OpenCV — median bit-identical to VPI too

edgarriba:
## Summary

PRs 8+9 of the roadmap combined: `median_blur` + `bilateral_filter`, CPU and CUDA in one PR so the twins land together.

**Median** (u8 C1/C3, ksize 3/5): recon showed `cv2.medianBlur` and VPI-CUDA `MedianFilter` are **already bit-identical to each other** (replicate borders, exact medians) — so one implementation matches BOTH. Exact sorting networks (Paeth-9 / Smith-25) kept as const exchange lists; the CUDA kernel source is **generated from the same lists** the CPU walks (single source — the sides cannot drift). NEON CPU path runs the network on 16 u8 lanes at once (min/max are exact; byte-identical).

**Bilateral** (u8 C1): `cv2.bilateralFilter` mirrored end to end:
- cv2's **own SIMD exp polynomial** (`v_exp`, transcribed operation-for-operation with `mul_add` — NOT libm `expf`; the ULP differences are visible in output bytes) for color-table entries 0..251, scalar `expf` for the last 4 (cv2's nlanes split);
- f64-exp space weights over the circular tap mask, reflect_101 borders, `radius = d/2` / `round(1.5σ)` rules, degenerate-σ copy-through;
- per-tap fma accumulation and `rint(sum/wsum)`;
- **cv2's position-dependent tap-summation order**: its unrolled `maxk==13` (d=5) SIMD block sums the window lines in order 1,5,2,4,3 for the 16-pixel-aligned region of each row, while its scalar tail sums sequentially. f32 addition is order-sensitive — both orders and the split point are reproduced (that was the last maxdiff=1 to fall).

VPI's bilateral uses a different formula (maxdiff ~32 vs cv2 measured) — byte parity against both simultaneously is impossible; cv2 is the reference, consistent with the rest of the library.

kornia-py: `imgproc.median_blur(image, kernel_size)` / `imgproc.bilateral_filter(image, d, sigma_color, sigma_space)`, numpy + device `Image` paths, stubs, parity + leak tests.

## Byte parity (through the wheel)

**47/47 configurations maxdiff=0**:
- median vs **cv2** and vs **VPI**: 5 shapes × k∈{3,5}, C3, GPU path — all exact
- bilateral vs **cv2**: 4 shapes × 5 (d, σc, σs) combos incl. d=0 auto-radius, adversarial content (two-level, narrow-range), GPU path — all exact

Rust: CPU↔GPU `assert_eq!` suites; networks verified against naive medians on 2000 random windows; 397 lib tests green; clippy clean; both kornia-py builds checked.

## Performance (Orin, locked clocks, 1080p sustained)

| op | kornia GPU | cv2 CPU 8T | VPI-CUDA | ×cv2 | ×VPI |
|---|---|---|---|---|---|
| median 3×3 | **0.015 ms** | 0.24 | 0.65 | **16×** | **43×** |
| median 5×5 | **0.015 ms** | 2.0 | 1.25 | **132×** | **80×** |
| bilateral d=5 | **0.065 ms** | 9.0 | 0.71 | **139×** | **10.2×** |

All three ops clear the ≥10× gate against BOTH baselines. CPU (NEON): median3 1.85 ms, median5 9.2 ms (11× faster than the scalar walk; cv2's O(1)-histogram CPU median stays faster — GPU carries the gate), bilateral CPU scalar 44 ms (value-indexed gathers; same wall as CLAHE, documented).

## ncu (locked clocks, `--set basic`)

- `median_3x3_c1`: occupancy 88%, 25 regs, SM 91%
- `median_5x5_c1`: occupancy 96%, 40 regs, SM 97% — compute-saturated exchange network
- `bilateral_u8`: occupancy 92%, 21 regs, SM 82%


edgarriba:
**CPU optimization final state** (1080p, same-minute interleaved A/B, `out=` both sides, locked clocks):

| op | kornia CPU | cv2 CPU | verdict |
|---|---|---|---|
| median 5×5 | **1.47 ms** | 1.66 | **KORNIA 1.13×** |
| bilateral d=5 | **7.01 ms** | 8.43 | **KORNIA 1.20×** |
| median 3×3 | 0.284 ms | 0.251 | cv2 by 12% |

Single-threaded, the kernels are at or above cv2's (median3 tied 1.003 vs 1.024 ms; median5 1.3× better) — the 3×3 residual is python-binding overhead + scheduler noise at the 0.28 ms scale, not kernel quality.

Key pieces: rolling sorted-column 5×5 (71-exchange network derived by greedy deletion, zero-one-proven on the 6⁵ sorted-column subspace, pinned by unit test), two-rows-per-pass 3×3 with `vext`-shared column sorts + `stnp` non-temporal stores, 16-lane per-tap bilateral, and per-op-size rayon chunking (static split for sub-ms ops, fine chunks for 10 ms ops — each measured).

Byte parity re-verified after every step: 0 fails across the full cv2+VPI sweep. GPU numbers unchanged (median 0.015 ms, bilateral 0.065 ms — ≥10× both baselines).

PR #1028, opened 2026-07-19
feat(imgproc): connected-component labeling (CPU+CUDA), label-exact with OpenCV SAUF

edgarriba:
## Summary

PR 13: `connected_components` for u8 single-channel masks — i32 labels + count, CPU and CUDA, **label-for-label identical to OpenCV's SAUF** (`cv2.connectedComponentsWithAlgorithm(img, conn, cv2.CV_32S, cv2.CCL_WU)`) for both 4- and 8-connectivity.

Recon finding that fixed the target: cv2's *default* 8-way algorithm (BBDT) numbers components in block-scan order — not reproducible without replicating its decision tree — while its SAUF variant numbers them in raster order of first appearance for both connectivities. Raster-first numbering is exactly what a min-index union-find produces naturally, so SAUF is the byte-exact target; BBDT yields the same partition with different label values (documented).

- **CPU**: run-based min-index union-find (runs collapse to their start for free; two-pointer overlap unions against the previous row), raster-first compaction.
- **CUDA**: `atomicMin` label-equivalence fixpoint with row-run initialization + pointer-jump compression, then a fully-device compaction (parallel per-block scan → offsets → relabel) that renumbers roots in index order — provably the CPU numbering, since a component's root IS its first raster pixel.
- **Infra**: `pair_residency` generalized over destination element type (first mixed-type device pair: u8 → i32); kornia-py `Image` gains full int32 dtype support (device variant, numpy/dlpack/buffer arms) for label maps.

## Label parity (through the wheel)

- CPU: **40/40 configs** (5 shapes × 4 densities 20–85% × both connectivities) label-exact vs cv2 CCL_WU
- GPU: label-exact on the same sweeps + 24 rust CPU↔GPU `assert_eq!` configs + structured-blob content
- 409 rust + 9 py tests green; clippy clean; both kornia-py builds checked

## Performance — honest status

Correctness-complete; speed is deliberately deferred (task #38). 1080p 40% density: kornia CPU ~40 ms, GPU ~11 ms vs cv2 ~5.7 ms. ncu shows exactly why: 14 merge sweeps × 727 µs — the naive label-equivalence fixpoint. The scoped fix is block-local CCL (solve 32×32 tiles in shared memory, then boundary-only global merges → ~2–3 iterations) for GPU and a proper SAUF two-pass for CPU. Numbers this session are also polluted by concurrent load (load avg 5+); task #38 re-measures on a quiet machine.

## ncu (locked clocks)

- `ccl_merge`: occ 77%, 17 regs, ×14 launches (the follow-up's target)
- `ccl_compress`: occ 87% ×14; `ccl_scan_partial`: occ 78%, SM 75%; `ccl_relabel`: occ 75%
- `ccl_init` (row-run): 1.4 µs

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,20 @@ changes early: `cargo add kornia-imgproc@0.1.15-rc.1` or `pip install --pre korn
 
 ## [Unreleased]
 
+**Connected-component labeling (CPU and CUDA), label-exact with OpenCV's
+SAUF.** New `connected_components` for u8 single-channel masks
+(`kornia_rs.imgproc.connected_components` in Python, returning
+`(n, labels)` with int32 labels), matching
+`cv2.connectedComponentsWithAlgorithm(..., cv2.CCL_WU)` label-for-label:
+background 0, components numbered in raster order of first appearance,
+for both 4- and 8-connectivity. (cv2's default 8-way algorithm, BBDT,
+produces the same partition with a different numbering.) The CPU path is
+a run-based min-index union-find; the CUDA path is an atomicMin
+label-equivalence fixpoint with row-run initialization plus a device
+compaction — device labels are identical to the CPU's. Python `Image`
+gains int32 (label map) dtype support. Performance follow-up tracked:
+current implementations prioritize the exactness contract.
+
 **Canny edge detection (CPU and CUDA), byte-for-byte with OpenCV.** New
 `canny` for u8 single-channel images (`kornia_rs.imgproc.canny` in
 Python), matching `cv2.Canny` exactly — the all-integer pipeline (Sobel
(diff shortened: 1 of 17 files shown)


edgarriba:
**Perf follow-through after review feedback** (GPU slower than cv2 CPU was not shippable):

| content | cv2 CPU | kornia GPU before | after | |
|---|---|---|---|---|
| dense 40% noise | 5.8 ms | 11.7 | **4.3** | **1.3× — beats cv2** |
| sparse blobs | 0.93 ms | 4.4 | 2.6 | fixed-overhead-bound |

The fix was algorithmic: the merge/compress host-iteration loop (ncu: 14 sweeps × 727 µs) is replaced by a **single lock-free union pass** (Komura atomicMin-with-retry) + inline root chase in relabel — zero iteration, zero convergence syncs. Vertical unions now emit once per run overlap instead of per pixel. One subtle bug caught by the parity suite along the way: non-atomic path-halving stores can overwrite a concurrent atomicMin link (lost union) — halving must be atomicMin too.

CPU: 40 → 22 ms (stripe-parallel run-based union-find + run-level resolve). Remaining gaps tracked in task #38 (SAUF two-pass CPU; sparse-case op floor for GPU).

Label parity re-verified after every change: **81/81 wheel configs label-exact vs cv2 CCL_WU, 0 fails**; 409 rust + 8 py tests green.

PR #1034, opened 2026-07-19
feat(tensor): UnifiedResource + CudaUnifiedAllocator for CUDA managed memory

Incharajayaram:
## 📝 Description

**Fixes/Relates to:** 

Implements the missing `MemoryDomain::Unified` allocation path. `MemoryDomain::Unified` existed in the type system since the three-state domain enum was introduced but had no allocation backing. This PR closes that gap by wiring `cuMemAllocManaged` into the allocator chain.

On **Jetson** (integrated GPU + CPU sharing physical memory), `cuMemAllocManaged` eliminates the PCIe copy entirely. On discrete GPUs (GTX/RTX), the driver handles demand paging transparently.

---

## 🛠️ Changes Made

- [x] `UnifiedResource`: owning `MemoryResource` backed by `cuMemAllocManaged(CU_MEM_ATTACH_GLOBAL)`, domain `MemoryDomain::Unified { id }`, freed via `cuMemFree` on drop
- [x] `CudaUnifiedAllocator` / `unified_alloc()`: `TensorAllocator` producing `UnifiedResource`, following the same pattern as `PinnedAllocator`
- [x] `zeros_unified<T, N>()`: zero-initialised unified tensor, parallel to `zeros_cuda` / `zeros_pinned`
- [x] All three items re-exported from `kornia-tensor` crate root
- [x] Updated module-level doc comment in `cuda.rs`

---

## 🧪 How Was This Tested?

- [x] **Unit Tests:**
  - `unified_host_access_and_domain`: verifies domain is `Unified`, `as_slice` works on host, GPU kernel increments elements via unified pointer without any H2D copy, CPU reads back without D2H copy
  - `unified_drop_no_double_free`: second `zeros_unified` succeeds after first drops (no CUDA error state after `cuMemFree`)
- [x] **Manual Verification:** `cargo test -p kornia-tensor --features cuda` — all 26 existing tests pass, no regressions; `cargo build` clean with no warnings
- [x] **Performance/Edge Cases:** GTX 1650 (discrete GPU, CUDA 12.4) — allocation and host read/write work correctly; demand paging overhead on discrete GPU is expected and documented

---

## 🕵️ AI Usage Disclosure

---

## 🚦 Checklist

---

## 💭 Additional Context

The `UnifiedResource` drop calls `cudarc::driver::result::memory_free` (wraps `cuMemFree_v2`) rather than the async `cuMemFreeAsync` used by `CudaSlice` — this matches the allocation API (`cuMemAllocManaged` is not a stream-ordered allocation).

The test uses `stream.upgrade_device_ptr` to alias the unified pointer as a `CudaSlice` for the kernel arg, then `std::mem::forget`s the alias to prevent a double-free (the memory is owned by `UnifiedResource`).


crates/kornia-tensor/src/cuda.rs
```diff
@@ -370,6 +377,151 @@ where
...
+/// Unified memory is accessible from **both** the CPU and any GPU stream that
+/// can see the device, so its domain is [`MemoryDomain::Unified`] and
+/// [`as_slice`](crate::storage::TensorStorage::as_slice) works without an
+/// explicit H2D copy.  The CUDA driver migrates pages on demand, so on Jetson
+/// (integrated GPU + CPU on shared physical memory) there is literally no copy;
+/// on discrete GPUs (e.g. GTX/RTX) the driver still handles migration, though
+/// demand-paging overhead makes explicit copies faster for large bulk transfers.
+///
+/// Freed with `cuMemFree` exactly once on drop.
+pub struct UnifiedResource {
```
edgarriba:
```suggestion
pub struct CudaUnifiedResource {
```

edgarriba (commented):
i'd say first test in the Jetson. Would be great to see some sort of numbers

crates/kornia-image/src/cuda.rs
```diff
@@ -120,6 +120,25 @@ where
...
+    /// accessible without explicit copies, e.g. for Jetson or fine-grained UVM
+    /// workflows).
+    ///
+    /// Mirrors [`Self::zeros_cuda`] / [`Self::zeros_pinned`] for the unified
+    /// memory domain.
+    ///
+    /// # Errors
+    ///
+    /// Returns [`ImageError::Cuda`] on allocation failure.
+    pub fn zeros_unified(
```
edgarriba: zeros_cuda_unified

edgarriba: Same for the tensor call

crates/kornia-tensor/src/cuda.rs
```diff
@@ -1376,6 +1528,73 @@ mod tests {
         );
     }
 
+    /// Unified memory is host-accessible, domain is `Unified`, and it round-trips
+    /// through CUDA kernel execution without an explicit H2D copy.
+    #[test]
+    #[ignore = "requires CUDA"]
```
edgarriba: This shouldn't be ignored

crates/kornia-tensor/src/cuda.rs
```diff
@@ -370,6 +377,151 @@ where
...
+pub struct CudaUnifiedAllocator {
+    /// Shared CUDA context (keeps the driver alive and provides the device id).
+    pub ctx: Arc<CudaContext>,
+}
+
+// SAFETY: Arc<CudaContext> is Send + Sync.
+unsafe impl Send for CudaUnifiedAllocator {}
+unsafe impl Sync for CudaUnifiedAllocator {}
+
+impl TensorAllocator for CudaUnifiedAllocator {
```
edgarriba, line 452: This allocator needs test too

edgarriba (changes requested):

edgarriba:
### Code review

Found 3 issues:

1. `CudaUnifiedResource::drop` leaks two `CUevent` handles and an `Arc<CudaStream>` ref per allocation. It frees the pointer directly and never drops the `ManuallyDrop<CudaSlice<T>>`, but `stream.upgrade_device_ptr` in `alloc_managed` builds that slice with `read`/`write` events whenever `ctx.is_event_tracking()` — which defaults to `true` in cudarc 0.19.7 (`core.rs:2501`, `event_tracking: AtomicBool::new(true)`). Nothing ever destroys them, so a per-frame allocation loop (as in `bench_unified_memory.rs`) leaks steadily. The sibling `CudaResource::drop` solves exactly this with `ManuallyDrop::take` + `slice.leak()`; `CudaSlice::leak()` drops `read`/`write`/`stream` and returns the raw pointer without freeing it, and additionally waits on both events first — a sync the current `Drop` skips.

https://github.com/kornia/kornia-rs/blob/d3ab26851085adffd3fe97c6b3e6c67f78515f01/crates/kornia-tensor/src/cuda.rs#L444-L454

2. `CudaUnifiedAllocator`'s doc tells users to pass the handle to `_in` constructors to "create host-and-device-accessible tensors", but an `AllocHandle` is an advisory tag — the `Vec`-based `_in` constructors adopt the vec's own heap memory and never route the allocation through it, so the result is ordinary host memory. The sibling `pinned_alloc` carries exactly that caveat ("they do not route the allocation through it. To get storage that actually lives in pinned memory, use `zeros_pinned`", [cuda.rs#L329-L334](https://github.com/kornia/kornia-rs/blob/d3ab26851085adffd3fe97c6b3e6c67f78515f01/crates/kornia-tensor/src/cuda.rs#L329-L334)); the unified doc drops it.

https://github.com/kornia/kornia-rs/blob/d3ab26851085adffd3fe97c6b3e6c67f78515f01/crates/kornia-tensor/src/cuda.rs#L455-L460

3. `CudaUnifiedResource.len_bytes` stores the padded `n_bytes` (`layout.size().max(1)`) instead of the true `layout.size()`, diverging from `PinnedResource` ([cuda.rs#L322-L324](https://github.com/kornia/kornia-rs/blob/d3ab26851085adffd3fe97c6b3e6c67f78515f01/crates/kornia-tensor/src/cuda.rs#L322-L324)). For a zero-element tensor `len_bytes()` reports 1 instead of 0, so `TensorStorage::layout()` (reconstructed from `owner.len_bytes()`) disagrees with the requested layout. Same in `zeros_cuda_unified` at [L570](https://github.com/kornia/kornia-rs/blob/d3ab26851085adffd3fe97c6b3e6c67f78515f01/crates/kornia-tensor/src/cuda.rs#L569-L571).

https://github.com/kornia/kornia-rs/blob/d3ab26851085adffd3fe97c6b3e6c67f78515f01/crates/kornia-tensor/src/cuda.rs#L486-L490

🤖 Generated with [Claude Code](https://claude.ai/code)

<sub>- If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

edgarriba:
### Design note: couple or decouple `CudaUnifiedResource` from `CudaResource`?

Separate from the correctness review above — a suggestion on the shape, since this is the moment to pick one.

The current design splits at the type level (`CudaResource<T>` vs `CudaUnifiedResource<T>`) and then bridges with per-accessor `.or_else(downcast)` cascades. That bridge is already incomplete. There are 7 downcast sites in `crates/kornia-tensor/src/cuda.rs`; this PR teaches 3 of them about unified memory (`as_cudaslice`, `cuda_stream`, `as_cudaslice_mut`). The other 4 still reject unified tensors:

| fn | result for a `zeros_cuda_unified` tensor |
|---|---|
| [`into_cudaslice`](https://github.com/kornia/kornia-rs/blob/c0168375fbe66aed17006f0719aca8c0fc787cc6/crates/kornia-tensor/src/cuda.rs#L1181-L1187) | `Err(self)` |
| [`to_host`](https://github.com/kornia/kornia-rs/blob/c0168375fbe66aed17006f0719aca8c0fc787cc6/crates/kornia-tensor/src/cuda.rs#L1258-L1263) | `CudaError::NotCudaBacked` |
| [`to_host_into`](https://github.com/kornia/kornia-rs/blob/c0168375fbe66aed17006f0719aca8c0fc787cc6/crates/kornia-tensor/src/cuda.rs#L1316-L1323) | `CudaError::NotCudaBacked` |
| [`to_host_in`](https://github.com/kornia/kornia-rs/blob/c0168375fbe66aed17006f0719aca8c0fc787cc6/crates/kornia-tensor/src/cuda.rs#L1348-L1354) | `CudaError::NotCudaBacked` |

`to_host_into` shows the hazard well: it is now half-taught. Its `self.cuda_stream()` call succeeds for a unified tensor (patched by this PR), but the `downcast_ref::<CudaResource<T>>()` two lines above it does not, so the function still bails. Every future accessor, and every future backing (NVMM, DLPack-managed), pays this tax again.

**Suggested alternative: one resource, backing as a field.**

```rust
pub struct CudaResource<T> {
    slice: ManuallyDrop<CudaSlice<T>>,
    ptr: *mut u8,
    id: i32,
    stream: Arc<CudaStream>,
    backing: Backing,
}

enum Backing {
    /// cudarc owns the allocation; CudaSlice::drop frees it.
    Device,
    /// cuMemAllocManaged; host-dereferenceable; leak the alias, then cuMemFree.
    Managed,
    /// External producer (e.g. DLPack import); leak the alias, drop the keepalive.
    Foreign(Box<dyn Any + Send>),
}
```

- `domain()` becomes a match on `backing` → `Device { id }` / `Unified { id }`.
- `Drop` becomes one match with three arms. Issue 1 in the review above (the leaked `CUevent` pair) becomes structurally impossible — there is no second `Drop` impl to forget `.leak()` in.
- All 7 downcast sites keep working with no diff, and the `as_cudaslice_mut` double-`downcast_mut` workaround goes away.
- Cost is one enum discriminant per allocation, not per element. Nothing on a hot path.

Two things that make coupling the natural fit here rather than a compromise:

1. The safety boundary is already enforced at runtime by domain, not by type — [`TensorStorage::as_slice`](https://github.com/kornia/kornia-rs/blob/c0168375fbe66aed17006f0719aca8c0fc787cc6/crates/kornia-tensor/src/storage.rs#L102-L109) asserts `domain.is_host_accessible()`. Merging the two structs removes no protection that the type split was providing. Meanwhile `MemoryDomain::{Device, Unified}` already encodes this distinction, so the type split encodes the same fact a second time.

2. `CudaResource` is already a backing sum type: [`foreign: Option<Box<dyn Any + Send>>`](https://github.com/kornia/kornia-rs/blob/c0168375fbe66aed17006f0719aca8c0fc787cc6/crates/kornia-tensor/src/cuda.rs#L104-L121) is a second drop strategy encoded as an `Option`, and `Drop` already branches on it. This just makes an implicit two-variant enum explicit and three-variant.

Worth noting that "decouple properly behind a trait" is not really available: `MemoryResource` is used as `Box<dyn MemoryResource>` and is erased over `T`, so a `fn as_cuda_slice(&self) -> Option<&CudaSlice<T>>` accessor cannot live on it without breaking object safety. That is precisely why the code downcasts, and why the type split has no clean bridge.

**Lower-risk fallback** if keeping the two types is preferred: centralise the downcast in a single helper and route all 7 sites through it —

```rust
fn cuda_backing<T: 'static>(owner: &dyn MemoryResource)
    -> Option<(&CudaSlice<T>, &Arc<CudaStream>)>
```

Smaller diff, and it fixes the 4 sites above. It still leaves two `Drop` impls free to drift apart, which is how they drifted in the first place.

Either way, the 4 `NotCudaBacked` paths are worth resolving before merge — otherwise `zeros_cuda_unified` reaches kernels but cannot round-trip back to host through the crate's own API.

🤖 Generated with [Claude Code](https://claude.ai/code)

Incharajayaram:
## Jetson numbers

Measured on a **Jetson Orin Nano Super** (sm_87, L4T R36.4.7, CUDA 12.6), with a **GTX 1650** (sm_75) as the discrete control. Both at `431ac34`, via `examples/bench_unified_memory.rs`:

```
cargo run --release -p kornia-imgproc --features cuda --example bench_unified_memory
```

### End-to-end `resize()` per frame

The number that matters — a real op through the public API. Explicit path is fill → `to_cuda` → `resize` → `to_host_image`; unified path is fill → `resize`, no transfers on either side. Downscale by 2.

| size | Jetson explicit | Jetson unified | | GTX 1650 explicit | GTX 1650 unified | |
|---|---|---|---|---|---|---|
| VGA 640×480 | 2.509 ms | 1.813 ms | **1.38×** | 2.329 ms | 4.053 ms | 0.57× |
| HD 1280×720 | 6.418 ms | 3.879 ms | **1.65×** | 7.759 ms | 11.861 ms | 0.65× |
| FHD 1920×1080 | 12.639 ms | 7.410 ms | **1.71×** | 17.199 ms | 26.041 ms | 0.66× |
| 4K 3840×2160 | 47.318 ms | 27.517 ms | **1.72×** | 67.520 ms | 102.180 ms | 0.66× |

Output is **bit-identical** to the device path on both machines.

This is the split the feature predicts: unified memory wins on the integrated part and *loses* on the discrete one, where the driver demand-pages over PCIe and pays more than the copies it removes. Managed memory is not a general-purpose win, and the benchmark reports `cudaDevAttrIntegrated` so a run says which case it is.

### Where the time goes, per frame

| | Jetson fill | fill_uni | H2D | D2H | explicit | unified |
|---|---|---|---|---|---|---|
| VGA 640×480 | 0.960 | 0.801 | 0.617 | 0.829 | 2.406 | 0.801 |
| HD 1280×720 | 3.018 | 2.408 | 1.656 | 2.153 | 6.827 | 2.408 |
| FHD 1920×1080 | 6.462 | 5.414 | 3.795 | 4.466 | 14.722 | 5.414 |
| 4K 3840×2160 | 25.987 | 21.647 | 16.656 | 28.709 | 71.352 | 21.647 |

`fill` is the CPU producing a frame into pageable memory, `fill_uni` the same work written straight into unified memory. On Jetson writing into unified memory is **not slower** than writing into pageable memory (21.647 vs 25.987 ms at 4K), so the removed H2D+D2H is close to pure profit.

### One-time allocation cost

| size | `zeros_cuda` | `zeros_cuda_unified` |
|---|---|---|
| VGA 640×480 | 0.011 ms | 1.912 ms |
| HD 1280×720 | 0.015 ms | 5.333 ms |
| FHD 1920×1080 | 0.013 ms | 11.232 ms |
| 4K 3840×2160 | 0.013 ms | 42.430 ms |

`cuMemAllocManaged` is roughly 3000× more expensive than a pooled device allocation, so unified buffers must be allocated once and reused — it amortizes after about one frame at 4K, but an allocate-per-frame loop would erase the win entirely.

edgarriba:
Consider the design notes above to couple more the cuda api

crates/kornia-imgproc/examples/bench_opencv_warp_affine.txt
edgarriba, line 1: Do not include this

crates/kornia-imgproc/benches/bench_unified_memory.rs
edgarriba, line 1: Benches should go under benchmark directory or similar; not examples

crates/kornia-imgproc/src/cuda/resize.rs
```diff
@@ -1059,4 +1059,67 @@ mod tests {
...
+        resize(&d_src, &mut d_dst, InterpolationMode::Bilinear)?;
+        let device_out = d_dst.to_host_owned()?;
+
+        // Unified: same op, but the buffers are unified rather than device.
+        let mut u_src = Image::<f32, 3>::zeros_cuda_unified(src_size, &stream)?;
+        let mut u_dst = Image::<f32, 3>::zeros_cuda_unified(dst_size, &stream)?;
+        let u_src_slice = u_src
+            .as_cudaslice_mut()
+            .ok_or("unified source has no device slice")?;
+        stream.memcpy_htod(host.as_slice(), u_src_slice)?;
```
edgarriba: To cuda

edgarriba (changes requested):

crates/kornia-imgproc/examples/bench_opencv_resize.txt
edgarriba, line 1: This artifact should be remove

edgarriba (changes requested):

edgarriba (approved):

PR #1050, opened 2026-07-26
feat(imgproc): CUDA and NEON SIFT, bit-exact against cv::SIFT

edgarriba:
Adds SIFT to `kornia-imgproc`: a CUDA pipeline and a NEON CPU one, both
reproducing `cv::SIFT` **bit for bit**, a brute-force matcher on each, and
Python bindings that dispatch on residency.

## Results

mh01_frame1 (752x480), own process, medians of 11 after warm-up, against
OpenCV 5.0 on all six cores and on one thread.

| backend / config | kp | ms | vs cv2 all-cores | vs cv2 1-thread |
|---|---|---|---|---|
| OpenCV 5.0, all cores | 2515 | 110.0 | 1.0x | 2.0x |
| OpenCV 5.0, 1 thread | 2515 | 224.1 | 0.5x | 1.0x |
| **CUDA exact** | 2515 | **18.6** | 5.9x | 12.0x |
| CUDA exact + budget 500 | 500 | 12.8 | 8.6x | 17.5x |
| CUDA fast | 2516 | 10.0 | 11.0x | 22.3x |
| **CUDA fast + budget 500** | 500 | **8.4** | **13.1x** | **26.7x** |
| CUDA `fo=0`, 4 octaves | 933 | 7.1 | 15.5x | 31.6x |
| NEON exact (6 threads) | 2515 | 79.3 | 1.4x | 2.8x |
| NEON exact + budget 500 | 500 | 52.2 | 2.1x | 4.3x |
| NEON `fo=0`, 4 octaves | 933 | 23.5 | 4.7x | 9.5x |

Pinned with `taskset -c 0-3` — the host had an unrelated job at ~97% CPU. The
GPU rows are unaffected (CUDA exact reads 18.60 here against 18.57 on an idle
box). The CPU rows are conservative: four pinned cores rather than six, so NEON
on an idle machine measures ~72.9 ms exact. Reported as measured rather than
mixing runs.

Read the ratios alongside the keypoint count. Only the 2515-keypoint rows do the
same work as OpenCV; a budget or `fo=0` produces fewer keypoints, so those are
operating points rather than speedups at fixed output. The directly comparable
results are **CUDA exact at 5.4x**, whose output is identical to cv2 on every
column of the matching audit, and **NEON exact at 1.4x**.

`fo=0` is the fastest row and the worst trade for pose — it deletes the finest
octave and scores 59.8% epipolar inlier ratio against 65.3%. A keypoint budget
keeps every octave and cuts only the weakest keypoints.

At the default configuration the CUDA output is **identical to cv2 on every
column** of the matching audit — not merely close:

```
engine          kp     ms  H match  H ok  F match  F inl  inl%   sed
cv2 (1 thr)   2515  225.2     5293  5232      816    533  65.3  0.29
cv2 (6 thr)   2515  103.6     5293  5232      816    533  65.3  0.29
CUDA          2515   19.6     5293  5232      816    533  65.3  0.29
NEON          2515   80.6     5293  5232      816    533  65.3  0.29
```

The two backends agree with each other to the bit as well: joined on
position and angle, **2515/2515 keypoints and 2515/2515 descriptors are
bit-identical**, so residency changes the speed and nothing else.

## Update 2026-07-28 — optimisation audit round (commits `b941e2f4..41890076`)

A three-pass audit (fresh `nsys` profile, four review agents with distinct
lenses, adversarial verification pass) and its implementation. Every change is
gated on the bitwise oracle suite (56/56 across ten images) and the matcher's
pair set was re-verified **identical to cv2's BFMatcher** (816 pairs on mh01).
Full ledger with verdicts: `docs/sift-audit-2026-07-28.md`.

Final numbers, quiet host, medians of 15, pinned:

| config | before | after |
|---|---|---|
| CUDA exact | 18.6 | **16.7 ms** |
| CUDA exact + budget 500 | 12.8 | 12.3 |
| CUDA fast + budget 500 | 8.4 | **8.0–8.8** |
| NEON exact, 6 threads | 79.3 | **68.2** |
| CPU single-thread (criterion, whole NEON round) | ~186 | **164.2** |
| device match, 2515x2447 (both scans + ratio + readback) | — | **12.6 ms** |

Quality re-verified on the final build: exact rows identical to cv2 on every
audit column, and both matchers — dp4a included — reproduce cv2's BFMatcher
pair set exactly (816 pairs).

What landed:

- **u8 + `__dp4a` matcher.** Descriptors are exactly the integers 0..255 in
  f32, so distances become exact i32 arithmetic — the old f32 sums were exact
  integers too (all partials < 2^24), so pairs and tie behaviour are
  *identical*; a representation change, not an approximation. ~4x less
  candidate traffic, ~8x fewer inner-loop ALU ops.
- **Descriptor scatter skips its dead border writes** (both backends,
  bit-exact): the fold reads only interior histogram cells, and 36% of the
  eight trilinear adds per sample landed in the write-only border ring.
- **CPU downsample division removed**: the general rescale index was provably
  `2*x` for the only caller — a non-pipelined 64-bit `udiv` per output pixel
  that sat *between* two stage probes and never appeared in any breakdown.
- Dead code the retain_best reorder had orphaned (18 device-to-device `ranges`
  snapshots/frame written then overwritten), a dead 1.3 MB memset per frame in
  the Python descriptor return, and per-launch `env::var` scans.
- Kernel micro items: reduction scratch sized to the reduced length instead of
  the block (2 KB static smem back per descriptor block), `__launch_bounds__`
  on every kernel, runtime divisions in the two sample loops replaced with
  incremental carries (map-preserving, bit-exact), a one-step `refl101`, and a
  warp-uniform border test that stops the first warp of every blur row
  executing both paths.
- NEON: the scatter's zip staging moved register-to-register; the blur's
  horizontal borders now run through the vector interior over a reflected edge
  buffer (FilterEngine's shape); the upsample's horizontal pass and the
  descriptor's collect loop are vectorised — each with the audit-documented
  contraction traps respected (no FMA introduced anywhere the reference has
  two roundings).

A follow-up NEON round (`214cf31d..c80309ed`) vectorised the blur borders,
the upsample's horizontal pass and the descriptor's collect loop, and added
inline pointer stepping to the blur V interior — each gated on the oracle,
the x86 fallback build, and (for the V pass) a pinned criterion A/B.

Two audit findings were **falsified by implementation** and reverted, with
their numbers recorded so they are not re-derived: 2-row blocking for the
blur V pass (+0.76% — the pass is not load-bound; the prefetcher already
serves adjacent-row re-reads) and per-layer orientation launch ranges. `adjustLocalExtrema` reassigns a
keypoint's layer during refinement, so launch-order slot contiguity does not
partition by the refined layer that selects the Gaussian image — the oracle
budget test caught it at 472 != 471. Recorded in the audit doc so it is not
re-derived.

Still open from the audit (listed with bounds in the doc): deferred count
readbacks + graph capture of the static prefix, and a columnar keypoint
accessor on the Python side.

## API

```python
sift = kornia_rs.imgproc.Sift(contrast_threshold=0.04)
kp, desc = sift.detect_and_compute(device_image)   # CUDA
kp, desc = sift.detect_and_compute(numpy_array)    # NEON
ka, kb, matches = sift.match(image_a, image_b)     # device: descriptors stay on device
```

Shaped like `cv2.SIFT`. The instance owns both backends' scratch and
rebuilds only when the geometry or a parameter changes. `Sift.match` on two
device images never moves descriptors across the bus; mixed residency is
refused rather than silently transferred.

Rust: `cuda::sift::{SiftCuda, SiftMatcher}` and
`features::{sift_detect_with, sift_match_descriptors, ...}`.

## Matching

Brute force, 128-D, Lowe ratio with mutual nearest neighbour. CUDA, NEON and
cv2's own `BFMatcher` return the **identical 816-pair set** on mh01 frame 1
against frame 2. The CUDA kernel uses one warp per query with each lane
owning 4 of the 128 dimensions, so a warp reads a descriptor in one
coalesced 512-byte transaction; the `n1 x n2` score matrix is never
materialised.

A scalar fallback ships alongside the NEON one, always compiled rather than
`cfg`-ed to the platforms that lack NEON, with a test pinning that the two
return the same pair set. `cargo clippy --target x86_64-unknown-linux-gnu`
is part of the checks — the fallbacks did not compile until that was run.

## On bit-exactness

The contract is equality with `cv::SIFT`'s bits, not a tolerance, which
required reproducing things that are properties of the reference's *build*
rather than of SIFT:

* Its `exp`, `atan2` and `magnitude` are overridden on aarch64 by a backend
  built on ARM's *estimate* instructions, so `expf`, `atan2f` and `sqrtf`
  do not reproduce them. CUDA emulates those with baked tables and integer
  exponent assembly; the CPU path calls the instructions directly.
* Its translation unit takes the **SIMD** dispatch, whose accumulation
  rounds differently from its own scalar tail (`w = w*mag` then `+=`, versus
  a contracted `+= W[k]*Mag[k]`), and whose smoothing associates
  differently. Assuming the scalar form left 96 of 1334 angles wrong.
* Sub-pixel refinement solves a near-singular Hessian — entries ~2.5e-3
  against a determinant ~1e-9 — so a single-ULP difference upstream is
  amplified ~1e5 into the result. Every contraction there is load-bearing,
  including one cofactor that must mirror another's spelling and negate,
  because common subexpression elimination evaluates the pair once.

Verified across ten images including adversarial ones (saturated, flat,
pure noise, checkerboard, gradient, all with odd dimensions).

## Benchmarks and docs

`cargo bench -p kornia-imgproc --bench bench_features -- Sift` covers the
four detector configurations and both matcher kernels, grouped with FAST,
Harris and ORB. `docs/benchmark-sift.md` follows the dated System-table
format the existing benchmark docs use, and records the workflows: warm up
before timing CUDA (the first call compiles every kernel, ~1.2 s), `nsys`
rather than `ncu` on this host (no root needed), comparing stage sum against
wall time to find host-side overhead, the oracle harness and every debug env
var, and the chain-diff technique.

## Notes for review

**Bugs fixed here that a tolerance test would not have caught**, several
found only by the assembled pipeline or by cross-backend comparison:

* Descriptors computed in the wrong coordinate frame, zeroing every octave
  but the first.
* A missing ARM zero/infinity special case producing a NaN that silently
  dropped keypoints on flat patches.
* Unclamped counters driving out-of-bounds device writes on dense frames.
* `KORNIA_SIFT_DESC=exact` — the documented validation path — clobbering its
  own output.
* CPU and CUDA computing different layer sigmas from duplicated code that
  had drifted `powi` vs `powf`.
* A matcher ratio of exactly 1.0 rejecting every match on ties.

**Roughly a dozen plausible optimisations were implemented, measured and
rejected**, each recorded in the source with its numbers so they are not
retried: shared-memory band fusion, replicated histograms, run-merged
atomics, padded histogram strides, transposed patch walks, speculative
grids, vectorised tile stores, fast-math primitives, and the rotated-frame
descriptor on CPU — which is worth 2.9x on GPU and a regression on CPU,
because the two backends are bound by different things.

**Structural choices worth a look:** host numerics and input validation are
shared between the backends through one module rather than mirrored, after
the duplicated copies drifted; and the descriptor's final ordering is
decided on the host but applied on device by a gather, so descriptors are
never downloaded just to be shuffled.

Draft while the CPU path's Python surface settles.

---

## Known gaps — please read before reviewing

Three things I would rather state than have a reviewer find:

1. **`max_keypoints` applies only on the CUDA path**, so the same image returns
   a different keypoint count on host and device — contradicting the `.pyi`'s
   claim that residency changes speed and nothing else. Documented rather than
   fixed, because capping the CPU path changes what it computes; a maintainer
   should decide which way it goes.

   *(The `n_features > 0` gap noted in earlier revisions is closed: both
   backends now have a fault-injected Rust test, plus 14 pytest cases covering
   the Python `Sift` class on host and device.)*

2. **Every quality number rests on a single image pair** (mh01 frame1/frame2).
   The homography audit runs on 7 real images, but the epipolar column — the
   only pose-quality metric — has one pair, so figures like "65.3% vs 60.6%
   inliers" are 37 inliers with no error bar. The MH_01 ground-truth trajectory
   is already in-tree at `tests/data/euroc_mh01_gt/data.csv`; only the cam0
   images are missing, and with them this could score absolute pose error
   instead.

3. **`cuda/sift/matcher.rs` casts to `float4` on an alignment its public entry
   point cannot guarantee** (since the dp4a rewrite this sits in the quantise
   kernel; the best2 kernel now reads u32). Every in-tree caller slices from
   zero so it is latent, but an external caller passing a `CudaView` at a
   non-multiple-of-4 offset would hit a misaligned-address launch failure.
   Left unfixed deliberately: the two available fixes are a precondition
   cudarc cannot express portably, or dropping a measured optimisation.
   Related, new with the dp4a matcher: inputs that are not integer-valued
   0..255 are outside the matcher's contract (nothing in this crate produces
   them) — documented in the module header.

Falsified experiments are recorded in the source with their measured penalties
rather than silently dropped — shared-memory band fusion, interleaved gradient
planes, x-tiling the vertical blur, approximate HAL, f16, and sharing the batch
helpers at width four. Please leave those comments in place; several cost a full
measurement cycle to establish.


kornia-py/src/pyutils.rs
```diff
@@ -5,10 +5,28 @@
...
 
 use kornia_algebra::{Mat3F64, Vec2F64};
 
+/// Reshape a flat row-major block into `(len / cols, cols)`.
+///
+/// `PyArray2::from_vec2` derives the column count from the first row, so an
+/// empty result would come back as `(0, 0)` instead of `(0, cols)` and break
+/// any caller that slices a column or stacks the array. Going through a flat
+/// vector also skips the per-row `Vec` allocation and copy.
+pub(crate) fn rows_to_numpy<T: numpy::Element>(
```
edgarriba, line 19: why is this needed ?

kornia-py/src/cuda_ext/mod.rs
```diff
@@ -465,33 +482,62 @@ impl PyTensor {
...
-        let (data, shape): (Vec<f32>, [usize; 4]) = match &*me.inner {
+        let (data, shape): (Vec<f32>, Vec<usize>) = match &*me.inner {
             // Unreachable without `cuda`: `me.is_device()` is always false, so
             // F32 always took the zero-copy view above, and F16's `is_device()`
             // guard below never matches.
             #[cfg(feature = "cuda")]
             TensorInnerEnum::F32(t) => {
                 let host = t.to_host_owned().map_err(err)?;
-                (host.as_slice().to_vec(), t.shape)
+                (host.as_slice().to_vec(), t.shape.to_vec())
```
edgarriba: why to_vec lets avoid copies

crates/kornia-imgproc/src/cuda/sift/plan.rs
```diff
@@ -0,0 +1,927 @@
...
+    }
+
+    /// Detect, orient and describe, leaving the descriptors on device.
+    ///
+    /// `src` is a `width * height` f32 grayscale image in 0..255, matching the
+    /// reference's internal representation. Returns the keypoints; the matching
+    /// descriptor rows are reachable through
+    /// [`SiftCuda::descriptors_device`], so a caller that goes straight on to
+    /// matching never moves them across the bus.
+    pub fn detect_and_compute_device(
```
edgarriba:
```suggestion
    pub fn detect_and_compute(
```

crates/kornia-imgproc/src/cuda/sift/plan.rs
```diff
@@ -0,0 +1,927 @@
...
+                if (ang - 360.0).abs() < f32::EPSILON {
+                    ang = 0.0;
+                }
+                let o = &mut din[pos * DESC_IN_STRIDE..pos * DESC_IN_STRIDE + 4];
+                o[0] = k[0] * scale;
+                o[1] = k[1] * scale;
+                o[2] = (k[2] * scale) * 0.5;
+                o[3] = ang;
+            }
+            stream.memcpy_htod(&din, &mut self.desc_in.slice_mut(0..n * DESC_IN_STRIDE))?;
```
edgarriba: Why not use the ones in device already. This copy should be avoided

crates/kornia-imgproc/src/cuda/sift/plan.rs
```diff
@@ -0,0 +1,927 @@
...
+                    p1 += 1;
+                }
+                if oct < oct_dims.len() && layer < self.pyr[oct].len() {
+                    starts.push(p0 as i32);
+                    groups.push((oct, layer, p0, p1 - p0));
+                }
+                p0 = p1;
+            }
+            let n_i = [n as i32];
+            stream.memcpy_htod(&n_i, &mut self.desc_live.slice_mut(0..1))?;
```
edgarriba, line 596: Reúse data from device

crates/kornia-imgproc/src/cuda/sift/plan.rs
```diff
@@ -0,0 +1,927 @@
...
+            if !starts.is_empty() {
+                let cap = self.ranges.len().min(starts.len());
+                stream.memcpy_htod(&starts[..cap], &mut self.ranges.slice_mut(0..cap))?;
```
edgarriba, line 599: Reúse data from device

crates/kornia-imgproc/src/cuda/sift/plan.rs
```diff
@@ -0,0 +1,927 @@
...
+        self.n_desc = n;
+        Ok(keypoints)
+    }
+
+    /// Detect, orient and describe, returning host-side results.
+    ///
+    /// This is [`SiftCuda::detect_and_compute_device`] plus a single download of
+    /// the ordered descriptor block. A caller that goes straight on to matching
+    /// should use the device form and skip the round trip entirely.
+    pub fn detect_and_compute(
```
edgarriba: This method remove it, in cuda we should return in device all

crates/kornia-imgproc/src/cuda/sift/plan.rs
```diff
@@ -0,0 +1,927 @@
...
+            keypoints,
+            descriptors,
+        })
+    }
+
+    /// The ordered descriptor block from the last call, on device.
+    ///
+    /// Row `i` belongs to keypoint `i` of the returned keypoint list. Valid
+    /// until the next call, which overwrites it.
+    pub fn descriptors_device(&self) -> &CudaSlice<f32> {
```
edgarriba:
```suggestion
    pub fn descriptors(&self) -> &CudaSlice<f32> {
```

crates/kornia-imgproc/src/features/sift/mod.rs
```diff
@@ -0,0 +1,49 @@
...
+//! never drift between them — a one-ULP difference in a single tap shifts every
+//! layer of the pyramid and therefore every keypoint.
+
+// The `pub(crate)` re-exports below exist so `cuda::sift` can name the shared
+// numerics rather than re-declare the literals, which is how the layer sigmas
+// drifted once before (see `SiftCudaConfig::shared_config`). Nothing else in the
+// crate reads them, so they are gated to the build that has that consumer.
+mod params;
+pub(crate) use params::refl101;
+#[cfg(feature = "cuda")]
```
edgarriba: Here we shouldn’t have any cuda stuff

crates/kornia-imgproc/src/features/mod.rs
```diff
@@ -15,3 +15,33 @@ pub use cells::{
...
+    SiftKeypoint, SiftWorkspace, DESCR_LEN, ORI_HIST_BINS, SIFT_IMG_BORDER, SIFT_MAX_INTERP_STEPS,
+};
+
+/// Numerics both backends must agree on bit for bit.
+///
+/// `cuda::sift` names these instead of re-declaring the literals. The copies
+/// drifted once already — see `SiftCudaConfig::shared_config` for the
+/// `powi`/`powf` divergence that motivated the shared module — so a constant
+/// that appears in both pipelines has exactly one definition, here.
+#[cfg(feature = "cuda")]
```
edgarriba: No cuda here

crates/kornia-imgproc/src/features/sift/pipeline.rs
```diff
@@ -0,0 +1,837 @@
...
+/// for _frame in 0..2 {
+///     let feats = sift_detect_with(
+///         &mut ws, &img, w, h, &cfg, FirstOctave::Native, usize::MAX, false,
+///     )?;
+///     assert!(feats.is_empty()); // a flat image has no extrema
+/// }
+/// # Ok::<(), kornia_imgproc::features::SiftConfigError>(())
+/// ```
+#[allow(clippy::too_many_arguments)]
+pub fn detect_and_compute_with(
```
edgarriba: Only one detect and compute, decide which

docs/sift-audit-2026-07-28.md
edgarriba, line 1: Remove this

examples/cuda_camera_sift/src/main.rs
```diff
@@ -0,0 +1,257 @@
...
+            if args.fast { "fast" } else { "exact" },
+            args.ratio
+        );
+
+        // Warm up: the first call JIT-compiles every kernel (~1.2 s). Timing it
+        // would attribute the compile to whichever stage ran first, a mistake
+        // this module has made before.
+        if let Some(frame) = cam.grab_frame()? {
+            luma_to_f32(frame.buffer.as_slice(), &mut gray, w, h);
+            stream.memcpy_htod(&gray, &mut d_gray)?;
```
edgarriba: Use Image to_cuda api

examples/cuda_camera_sift/src/main.rs
```diff
@@ -0,0 +1,257 @@
...
+            let t0 = Instant::now();
+            let Some(frame) = cam.grab_frame()? else {
+                continue;
+            };
+            let buf = frame.buffer.as_slice();
+            if buf.len() < w * h * 2 {
+                continue; // short or corrupt buffer
+            }
+            luma_to_f32(buf, &mut gray, w, h);
+            stream.memcpy_htod(&gray, &mut d_gray)?;
```
edgarriba: To cuda

examples/cuda_camera_sift/src/main.rs
```diff
@@ -0,0 +1,257 @@
...
+            } else {
+                Vec::new()
+            };
+            let t3 = Instant::now();
+
+            // Carry this frame's descriptors forward for the next iteration.
+            if n_cur > 0 {
+                let src = sift.descriptors_device().slice(0..n_cur * DESCR_LEN);
+                let mut dst = d_prev.slice_mut(0..n_cur * DESCR_LEN);
+                stream.memcpy_dtod(&src, &mut dst)?;
```
edgarriba: To cuda

kornia-py/src/cuda_ext/cuda_sift.rs
```diff
@@ -0,0 +1,350 @@
...
+    let inner = if n == 0 {
+        TensorInnerEnum::F32R2(kornia_tensor::Tensor::<f32, 2>::zeros(shape))
+    } else {
+        // SAFETY: uninitialised is fine — the D2D below overwrites every one of
+        // the `n * DESCR_LEN` elements before anything can read the tensor.
+        // (`alloc_zeros` here was a dead 1.3 MB memset per frame.)
+        let mut dst = unsafe { kornia_tensor::uninit_cuda::<f32, 2>(shape, stream) }
+            .map_err(|e| PyValueError::new_err(e.to_string()))?;
+        stream
+            .memcpy_dtod(
```
edgarriba: Why copy here event from d2d, isn’t the tensorrt in cuda already

kornia-py/src/pyutils.rs
edgarriba: It reshapes the flat row-major Vec each op returns into an (N, cols) numpy array in one step. The alternative, PyArray2::from_vec2, needs a Vec<Vec<f32>> (one heap allocation + copy per row) and returns shape (0, 0) instead of (0, cols) for an empty result, which breaks any caller that slices a column. Every 2-D-returning binding (SIFT, ORB, matches) funnels through this one helper so the empty-shape rule lives in one place.

kornia-py/src/cuda_ext/mod.rs
edgarriba: Fixed: the D2H path now moves the downloaded buffer with into_vec() instead of copying it a second time. (The remaining shape.to_vec() is 2-4 usizes — the variants carry different ranks, so the arms have to widen to a common type.)

crates/kornia-imgproc/src/cuda/sift/plan.rs
edgarriba: Done — packing now runs on device (sift_pack_desc) reading the ori_kp rows the device already owns; only the 4-byte row indices cross the bus, and the kernel's expressions match the old host pack bit for bit.

crates/kornia-imgproc/src/cuda/sift/plan.rs
edgarriba: n here is host-derived control data, not device data round-tripped: retain_best and the dedup are host-side sorts, so the survivor count only exists on the host. It is 4 bytes, uploaded once per frame.

crates/kornia-imgproc/src/cuda/sift/plan.rs
edgarriba: Same as above — the group starts come out of a host-side stable sort by (octave, layer), so they originate on the host; ~48 bytes per frame. The per-keypoint payload that used to be packed and uploaded here is now produced on device (see the sift_pack_desc commit).

crates/kornia-imgproc/src/cuda/sift/plan.rs
edgarriba: Removed. detect_and_compute is now the device-resident function (keypoints back, descriptors stay on device via descriptors()); the host-downloading variant is gone and the tests download explicitly where they need bytes to compare.

kornia-py/src/cuda_ext/cuda_sift.rs
edgarriba: The plan has exactly one output buffer and the next detect_and_compute overwrites it, so handing Python a view would silently mutate under a caller holding two frames — which is exactly what frame-to-frame matching does. The D2D is what gives the returned Tensor its own lifetime; it is ~1.3 MB / ~20 us and never touches the host. Making the plan hand over ownership instead needs a borrowing/Arc-backed Tensor variant — happy to do that as a follow-up if you want the copy gone.

examples/cuda_camera_sift/src/main.rs
```diff
@@ -0,0 +1,263 @@
...
+        );
+
+        // Warm up: the first call JIT-compiles every kernel (~1.2 s). Timing it
+        // would attribute the compile to whichever stage ran first, a mistake
+        // this module has made before.
+        if let Some(frame) = cam.grab_frame()? {
+            luma_to_f32(frame.buffer.as_slice(), &mut gray, w, h);
+            let dev = Image::<f32, 1>::from_size_slice(size, &gray)?.to_cuda(&stream)?;
+            let d_gray = dev.0.as_cudaslice().expect("device image");
+            sift.detect_and_compute(&ctx, &stream, d_gray)?;
```
edgarriba: Detect and compute should accept Image

examples/cuda_camera_sift/src/main.rs
```diff
@@ -0,0 +1,263 @@
...
+            if buf.len() < w * h * 2 {
+                continue; // short or corrupt buffer
+            }
+            luma_to_f32(buf, &mut gray, w, h);
+            let dev = Image::<f32, 1>::from_size_slice(size, &gray)?.to_cuda(&stream)?;
+            let t1 = Instant::now();
+
+            // Detect + describe. Descriptors stay on device.
+            let d_gray = dev.0.as_cudaslice().expect("device image");
+            let kps = sift.detect_and_compute(&ctx, &stream, d_gray)?;
```
edgarriba: Why we are not returning descriptors? Something is off

examples/cuda_camera_sift/src/main.rs
edgarriba: Done — detect_and_compute now takes a device Image<f32, 1> directly, validating residency and that the geometry matches the plan. The example (and the Python binding) pass the Image through with no slice extraction.

examples/cuda_camera_sift/src/main.rs
edgarriba: Right — fixed properly. detect_and_compute now returns SiftCudaFeatures { keypoints, descriptors } where descriptors is an OWNED device buffer: the final gather writes each frame into a fresh allocation, so it costs nothing extra and the result has its own lifetime. That also deleted the Python binding's defensive D2D and the example's carry-forward copy — holding the previous frame's result is now enough.

PR #1063, opened 2026-07-31
Fix/colorjitter aliasing and params validation

khushi-070906:
Summary

This PR fixes two correctness issues:

Eliminates undefined behavior in the fused ColorJitter path caused by aliasing a shared (&[u8]) and mutable (&mut [u8]) reference to the same buffer.
Replaces panic paths for malformed Python params dictionaries with proper Python exceptions.
Details
Fix aliasing UB in fused ColorJitter

The fused ColorJitter implementation previously fabricated a &[u8] (current_src) pointing into the same memory as the live &mut [u8] destination buffer. Call sites then passed both references simultaneously to functions such as apply_saturation(input, dst, ...), creating aliased shared and mutable references to the same memory.

Although the implementation only performed read-before-write operations, this violates Rust's aliasing rules and is undefined behavior, with the potential for miscompilation under optimization or auto-vectorization.

This change:

Removes current_src.
Adds apply_saturation_inplace, apply_hue_inplace, and flush_linear_inplace.
Branches on dst_init to invoke either the in-place or out-of-place implementation, avoiding any aliased borrows.
Handle malformed parameter dictionaries gracefully

Malformed params dictionaries supplied from Python previously caused Rust panics:

dict_get used .unwrap() when a required key was missing.
dict_to_params indexed factors[op as usize] without validating the user-provided operation order.

These cases now return Python exceptions instead:

Missing required keys raise PyKeyError.
Invalid operation indices raise PyValueError.
Result
Removes undefined behavior from the fused ColorJitter implementation.
Converts panic paths into proper Python exceptions.
Improves robustness when handling invalid user input from Python.

### DIFF
--- a/kornia-py/src/augmentations.rs
+++ b/kornia-py/src/augmentations.rs
@@ -46,12 +46,20 @@ pub fn set_seed(seed: Option<u64>) {
 
 /// Extract a required key from a params dict.
 ///
-/// Panics on missing key to match the previous `unwrap()` behavior at each call site.
+/// Returns a `PyKeyError` (rather than panicking) if the key is missing,
+/// since `params` dicts come from arbitrary user code and a malformed dict
+/// should surface as a normal Python exception, not a Rust panic.
 fn dict_get<'py, T>(d: &Bound<'py, PyDict>, key: &str) -> PyResult<T>
 where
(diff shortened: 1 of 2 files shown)


edgarriba (changes requested):
This pr seems a mix of things. Please one fix or feature per pr

PR #1064, opened 2026-07-31
Fix/dlpack int32 import

khushi-070906:
Summary

Add the missing K_DL_INT mapping in from_dldatatype to support importing int32 DLPack tensors.

Details

to_dldatatype already supported exporting int32 images via DLPack, but from_dldatatype did not include the corresponding K_DL_INT match arm.

This meant that importing any int32 DLPack tensor—including one exported by this library—would fail with an "unsupported DLPack dtype" error.

This change adds the missing K_DL_INT case to from_dldatatype, making int32 DLPack support symmetric between export and import and enabling successful round-trip conversions.

### DIFF
--- a/kornia-py/src/augmentations.rs
+++ b/kornia-py/src/augmentations.rs
@@ -46,12 +46,20 @@ pub fn set_seed(seed: Option<u64>) {
 
 /// Extract a required key from a params dict.
 ///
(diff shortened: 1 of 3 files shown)


edgarriba:
Pr title and description doesn't have anything to do with the actual code changes

PR #1068, opened 2026-08-03
feat(imgproc): u8 remap with fixed-point quantisation (CPU + CUDA)

Incharajayaram:
## 📝 Description

**Fixes/Relates to:** no issue — this implements the follow-up asked for in review on #998:

> Can you explore next a remap function for u8 images? I guess some sort if quantisation during the interpolation might need to be applied

Adds `remap_u8`, a u8 remap with a CPU path and CUDA bilinear/nearest kernels, dispatched by residency like the f32 `remap`.

### On the quantisation

You were right that it's the crux. Interpolating u8 means computing a weighted sum and landing back in 0..=255, and the choice of how is what decides whether the GPU agrees with the CPU — and whether u8 remap agrees with the u8 warp already in the tree.

This does **not** invent a scheme. It reuses the Q10 fixed-point blend `warp_affine_u8` already uses, so the two u8 ops round identically:

```c
unsigned int fx = (unsigned int)((sxc - (float)x0) * 1024.0f);   // Q10 weight, truncated
...
dst[out + ch] = (unsigned char)((top * fy1 + bot * fy + (1u << 19)) >> 20);  // round-half-up
```

That `(… + (1<<19)) >> 20` is character-for-character `warp_affine_u8`'s blend.

One difference from `warp_affine_u8` worth calling out, since it's the thing I expected to be a problem and wasn't: that kernel carries a Q16 coordinate accumulator anchored at the row span's left edge, because it derives coordinates incrementally along a row. Remap has no span and no accumulator — every coordinate is read straight from `map_x`/`map_y` — so there is nothing to anchor and the Q10 weights come directly off the float coordinate. The exactness argument is correspondingly simpler: both sides truncate the same f32 to Q10 and run the same integer blend.

Nearest needs no quantisation at all: it copies the source byte, so it is exact by construction.

---

## 🛠️ Changes Made
- [x] `remap_u8` public API (`Image<u8, C>` src/dst, `Image<f32, 1>` maps), residency-dispatched like `remap`
- [x] CPU u8 path with Q10 fixed-point bilinear and nearest
- [x] CUDA kernels `remap_bilinear_u8_c{C}` / `remap_nearest_u8_c{C}` — `__ldg` on source and maps, zero border for out-of-range coordinates, channel count specialised at compile time
- [x] Launchers `launch_remap_bilinear_u8_cuda` / `launch_remap_nearest_u8_cuda` and the `remap_u8_cuda` dispatch adapter
- [x] Device-vs-host parity tests, and a benchmark against the f32 kernel

---

## 🧪 How Was This Tested?

- [x] **Unit Tests:** 3 CPU tests (`remap_u8_identity_bilinear`, `remap_u8_bilinear_quantizes_fractional_weights`, `remap_u8_nearest_zeroes_oob_maps`) plus 3 device-vs-host parity tests (`public_remap_u8_device_equals_host`, `public_remap_u8_nearest_device_equals_host`, `public_remap_u8_channels_device_equals_host`). The parity tests are the ones that matter for the quantisation question — they assert the GPU is **bit-exact** against the CPU u8 path, over the cases where rounding actually diverges: fractional map steps (`1.03x − 1.7`, `1.07y − 2.3`) so bilinear weights are never 0 or 1; negative and overshooting coordinates for the border guard; 1, 3 and 4 channels, since the kernel takes the channel count at runtime and each is a distinct indexing path; sizes 65×33, 127×63, 16×16 and 1×1; both modes.

- [x] **Manual Verification:** Full suite on two architectures — **GTX 1650** (sm_75, discrete) and **Jetson Orin Nano** (sm_87, integrated, L4T R36.4.7 / CUDA 12.6). `cargo test -p kornia-imgproc --features cuda --lib` is 467 passed on the GTX and 478 passed on the Jetson, 0 failed on both; the u8 tests are 6/6 on each. Also builds and passes without the `cuda` feature (339 passed).

- [x] **Performance/Edge Cases:** Benchmarked against the f32 kernel with `examples/bench_remap_u8.rs`, 300 iterations after 60 warmup, device-resident, through the public API:

  **Bilinear**

  | size | GTX u8 | GTX f32 | | Jetson u8 | Jetson f32 | |
  |---|---|---|---|---|---|---|
  | FHD 1920×1080 | 0.451 ms | 0.565 ms | 1.25× | 0.682 ms | 0.831 ms | 1.22× |
  | 4K 3840×2160 | 1.818 ms | 2.341 ms | 1.29× | 2.414 ms | 3.133 ms | 1.30× |

  **Nearest**

  | size | GTX u8 | GTX f32 | | Jetson u8 | Jetson f32 | |
  |---|---|---|---|---|---|---|
  | FHD 1920×1080 | 0.262 ms | 0.446 ms | 1.70× | 0.437 ms | 0.719 ms | 1.65× |
  | 4K 3840×2160 | 1.070 ms | 1.848 ms | 1.73× | 1.740 ms | 2.867 ms | 1.65× |

  Bilinear gains less than the 4× the narrower pixel type suggests, and the benchmark prints achieved bandwidth to show why: the two **f32 coordinate maps are the same size either way**, so at 3 channels they are a large fixed share of the traffic and cap the win. Nearest reads one source pixel instead of four, so the maps are a smaller share and it gets closer to the ideal. Sub-HD sizes are launch-bound and show no reliable gain — the Jetson VGA/HD numbers moved between 0.94× and 1.5× across repeat runs, so I would not claim anything below FHD.

---

## 🕵️ AI Usage Disclosure
*Check one of the following:*
- [ ] 🟢 **No AI used.**

---

## 🚦 Checklist

The first two are unchecked because there is no linked issue — this comes from a review comment on an already-merged PR. Happy to open one if you'd prefer the process followed strictly.

---

## 💭 Additional Context

There is deliberately no "convert u8 to f32, remap, convert back" comparison in the benchmark: the only `cast_and_scale` in the tree is CPU-side, so that alternative cannot be measured on-device at all. That absence is part of the case for this kernel — without it, a u8 pipeline has to round-trip through the host to use the f32 remap, which costs far more than the 1.2–1.7× measured here.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -7887,6 +7887,7 @@ dependencies = [
 name = "kornia-imgproc"
 version = "0.1.15-rc.5"
 dependencies = [
+ "argh",
  "criterion 0.8.2",
  "cudarc 0.19.8",
  "half",
--- a/crates/kornia-imgproc/Cargo.toml
+++ b/crates/kornia-imgproc/Cargo.toml
@@ -19,6 +19,10 @@ name = "bench_cuda_remap"
 required-features = ["cuda"]
 
 [[example]]
+name = "bench_remap_u8"
+required-features = ["cuda"]
+
+[[example]]
 name = "dump_cuda_resize"
 required-features = ["cuda"]
 
@@ -114,6 +118,7 @@ thiserror = { workspace = true }
 wide = { workspace = true }
 
 [dev-dependencies]
+argh = { workspace = true }
 criterion = { workspace = true }
 # Deterministic instruction counts, immune to machine load. Needs valgrind and
 # `cargo install iai-callgrind-runner --version 0.14.0`.
--- a/crates/kornia-imgproc/examples/bench_remap_u8.rs
+++ b/crates/kornia-imgproc/examples/bench_remap_u8.rs
@@ -1,0 +1,180 @@
+//! u8 remap vs the f32 remap kernel.
+//!
+//! The u8 kernel exists so a u8 image can be remapped without widening to f32
+//! first. Its case is bandwidth: remap is memory-bound, and u8 moves a quarter
+//! of the bytes per channel, so the kernel should approach a 4x win on the
+//! image traffic — diluted by the two f32 coordinate maps, which are the same
+//! size either way and quickly dominate.
+//!
+//! Both paths are timed through the public API on device-resident images, so
+//! this measures the kernels as callers actually reach them.
+//!
+//! There is deliberately no "convert to f32, remap, convert back" column: the
+//! only `cast_and_scale` in the tree is CPU-side, so that alternative cannot be
+//! measured on-device at all. Which is itself the argument for this kernel — a
+//! u8 pipeline would otherwise have to round-trip through the host to use the
+//! f32 remap.
+//!
+//! ```text
+//! cargo run --release -p kornia-imgproc --features cuda --example bench_remap_u8
+//! ```
+
+use std::sync::Arc;
+use std::time::Instant;
+
+use argh::FromArgs;
+use cudarc::driver::{CudaContext, CudaStream};
+use kornia_image::{Image, ImageError, ImageSize};
+use kornia_imgproc::interpolation::{remap, remap_u8, InterpolationMode};
+
+/// Benchmark the u8 remap kernel against the f32 one.
+#[derive(FromArgs)]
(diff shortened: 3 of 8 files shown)


edgarriba:
For the cpu path include avx kernels to sees real boost vs scalar

edgarriba (approved):

PR #1069, opened 2026-08-04
fix(algebra): Sim3 exp and log were not mutual inverses

edgarriba:
## Problem

`Sim3F32::exp` and `Sim3F32::log` did not round-trip. The error was large, not marginal — about **1.2 in tangent norm** — and it was present even at `sigma = 0`, where Sim(3) must reduce *exactly* to SE(3).

The cause: Sim(3) is not SE(3) with a scale bolted on. Its translation block is

```
W = A·Ω + B·Ω² + C·I
```

whose coefficients `A`, `B`, `C` depend on **both** the rotation angle `theta` and the scale rate `sigma`. The two are coupled, because a point being rotated is simultaneously being scaled — the swept path is a logarithmic spiral rather than a circular arc. The previous implementation used the SE(3) coefficients and applied scale separately, which is only correct when one of the two is zero.

## Fix

Derive `A`, `B`, `C` from the coupled closed form, with series expansions on the degenerate branches (`theta → 0`, `sigma → 0`, and both together) so the small-angle and small-scale limits stay numerically stable rather than dividing by a vanishing denominator.

## Tests

Two new property tests:

- `sim3_exp_log_roundtrip_coupled` — six cases spanning the branch structure: general, pure SE(3) (`sigma = 0`), scale-only (no rotation), negative scale rate at large angle, both near zero, and large angle with large scale. This is the test that fails on `main`.
- `sim3_reduces_to_se3_at_unit_scale` — pins the SE(3) reduction, so the coupled path cannot silently regress the case that already worked.

Both are property tests over closed-form identities: no dataset, no machine dependence.

## Verification

- `cargo test -p kornia-algebra` — 291 passed, 0 failed
- `cargo fmt -p kornia-algebra -- --check` — clean
- `cargo clippy -p kornia-algebra --all-targets` — no warnings

Single file, no public API change, no dependency change.

### DIFF
--- a/crates/kornia-algebra/src/lie/sim3.rs
+++ b/crates/kornia-algebra/src/lie/sim3.rs
@@ -149,6 +149,68 @@ impl Sim3F32 {
         }
     }
 
+    /// The Sim(3) translation matrix `W(sigma, omega)`, shared by [`Self::exp`] and [`Self::log`].
+    ///
+    /// Sim(3) is NOT SE(3) with a scale bolted on. Its translation block is
+    /// `W = A*Omega + B*Omega^2 + C*I` whose coefficients depend on BOTH the rotation angle
+    /// `theta = |omega|` AND the scale rate `sigma` — the two are coupled because a point being
+    /// rotated is simultaneously being scaled, so the swept arc is a logarithmic spiral rather than
+    /// a circle. Using SE(3)'s `V` (which has no sigma in it at all) and dividing by the scale
+    /// afterwards, as this file previously did, is a different function; it agrees only when
+    /// sigma or theta is zero, which is why the existing tests passed — every one of them exercises
+    /// a translation-only, rotation-only or scale-only case.
+    ///
+    /// Returning ONE matrix used by both directions is the point: `log` inverts exactly what `exp`
+    /// applied, so the two are mutual inverses by construction rather than by two derivations
+    /// agreeing. They previously did not: measured round-trip tangent error was ~1.2 on a general
+    /// input, and ~1.2 even on a pure-SE(3) case with `s = 1`.
+    ///
+    /// Follows Strasdat's Sim(3) derivation (the formulation Sophus implements).
+    fn w_matrix(omega: Vec3AF32, sigma: f32) -> Mat3AF32 {
+        let omega_hat = SO3F32::hat(omega);
+        let omega_hat_sq = omega_hat * omega_hat;
+        let theta = omega.length();
+        let scale = sigma.exp();
+
+        let (a, b, c) = if sigma.abs() < SMALL_ANGLE_EPSILON {
+            // Scale rate vanishes: C -> 1 and W degenerates to SE(3)'s V, which is the ONLY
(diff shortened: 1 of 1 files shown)


edgarriba:
## Update: review fixes pushed (`c6519cf8`)

Three defects found in review, all addressed. Correcting one claim in the original description while I'm here.

### The round-trip tests do NOT pin this fix

The description above said the two new property tests pin the behaviour. **They do not, and cannot.** `log` recovers `(omega, sigma)` from `rxso3` alone and then applies the *same* `w_matrix(omega, sigma)` that `exp` applied, so the composition is `W⁻¹W = I` for **any** invertible `W`. The coupled branch had zero value coverage.

Verified by corrupting the coupled-branch denominator:

| test | with `W` deliberately corrupted |
|---|---|
| `sim3_exp_log_roundtrip_coupled` | **still passes** |
| `sim3_exp_matches_matrix_exponential` (new) | **FAILS** |

The new test compares `exp` against a truncated series for `expm` of the sim(3) generator, summed in f64 — a reference independent of the implementation. Cases span every branch, including two straddling the sigma threshold to pin continuity across it.

### 1. `C` dropped its first-order term

`C` was hard-coded to `1` across the whole small-sigma branch; the true value is `(e^s - 1)/s = 1 + s/2 + ...`. With the threshold at `1e-3` that is a **~5e-4 relative step in every translation** at the boundary — 5 cm on a 100 m tangent.

Invisible to every round-trip test, for the reason above: `log` inverts the same wrong `W`.

Now uses `exp_m1`, with a series only where the quotient is genuinely `0/0`.

### 2. `A` and `B` were catastrophic f32 cancellations

Their numerators are `~s²/2` and `~s³/6`, assembled from terms of size 1. Just above the threshold, in f32:

| sigma = 1.1e-3 | computed | true |
|---|---|---|
| `A` | 0.443 | 0.5006 |
| `B` | **33.6** | 0.1668 |

`B` wrong by 200×. Impact was bounded only because `Omega` is tiny in that branch — a latent trap for anyone raising the theta threshold or reusing the coefficients.

The coefficients are now computed in **f64** and cast once. They are three scalars per call, so the width costs nothing, and it removes the cancellation rather than adding more thresholds to tune.

### 3. The singularity comment was false

It claimed `W` "never approaches singularity". `W`'s eigenvalues are `(e^l - 1)/l` for `l` in `{s, s ± i·theta}`, so it is **exactly singular** at `sigma = 0, |omega| = 2*pi` — an input `exp` accepts.

The call site is safe for a different reason worth writing down instead: `RxSO3::log` flips the quaternion when `w < 0`, so `|omega| <= pi` always. The old wording invited reuse of `w_matrix(..).inverse()` on an unbounded exp-side tangent, where it silently yields inf/NaN.

### Verification

`cargo test -p kornia-algebra`: 292 passed, 0 failed. fmt and clippy clean.

The earlier `py3.8-linux` failure was infrastructure, not this change — sccache failed in post-job cleanup (`Mismatch of client/server versions?`) after the build had already succeeded.

PR #1074, opened 2026-08-04
fix(3d): harden the planar-pose solver (order-independent seed, second-solution semantics)

edgarriba:
## Why

`estimate_tag_pose` lived in `kornia-apriltag`, but its entire import list was:

```rust
use kornia_3d::pose::{homography_4pt2d, Pose3d};
use kornia_algebra::linalg::svd::svd3_f64;
```

**Nothing about it is AprilTag-specific.** It is Lu-Hager-Mjolsness orthogonal iteration over 4 coplanar correspondences — a tag is one caller among many (a chessboard cell, a printed rectangle, a door frame all work).

Because it sat there, every consumer of planar pose had to depend on the tag decoder. `kornia-calib` carried a hard dependency on `kornia-apriltag` for **exactly one import** — and since `kornia-apriltag` is excluded from crates.io (large git submodule of test images), that made `kornia-calib` unpublishable too.

That was a packaging accident, not a design constraint.

## What changed

| | |
|---|---|
| `kornia_3d::pose::planar` | new home: `estimate_planar_pose`, `PlanarPose`, `PlanarPosePair`, `PlanarPoseError` |
| `kornia-apriltag::pose` | thin re-export under the old names — existing callers compile unchanged |
| `kornia-calib` | `kornia-apriltag` dependency **dropped entirely** |
| `kornia-calib` | `publish = false` removed; added to `PUBLISH_ORDER` |

Verified the dependency is genuinely gone, not just direct:

```
$ cargo tree -p kornia-calib -e normal | grep -c apriltag
0
```

## A caveat the old docs did not state

The initial estimate comes from a homography fitted between a **canonical unit square** (`±1`, in the given corner order) and the image points. So initialisation is exact only when the object points really are a square in that order.

The orthogonal iteration then refines against the caller's actual `object_pts`, so a non-square coplanar quad still converges — it just starts further away. Callers with strongly non-square targets should raise `n_iters`. `kornia-calib` already relies on this: it passes per-tag centred object frames with no uniform-size assumption.

The module doc now says this. It also documents why **both** solutions are returned rather than one being silently chosen: planar pose is genuinely ambiguous, worst at small tilt and long range, and a caller with a second view or a motion prior can resolve it while a caller without one needs to know a choice was made.

## Verification

- `cargo test -p kornia-3d -p kornia-calib` — 177 + 21 + 13 pass, 0 fail
- the 3 moved tests run under their new home (`pose::planar::tests::*`)
- `cargo fmt --check` on all three crates — clean
- `cargo check -p kornia-3d -p kornia-apriltag -p kornia-calib` — clean

Note `cargo package -p kornia-calib` cannot succeed locally until `kornia-3d 0.1.15-rc.5` is on crates.io — that is the normal release-order constraint every crate in `PUBLISH_ORDER` has, not a remaining blocker from this change.

## Review requested

This changes public API layout across three crates and flips a crate to published — I would rather it be reviewed than fast-tracked.


edgarriba:
## Split: the publish-unblock moved to #1076

Review found three substantive issues in this PR that should not ship, while the relocation underneath them is sound. Rather than hold the whole thing, the mechanical half is now **#1076** — a pure refactor plus build-config fix, no behaviour change and no API break, which is what actually unblocks `kornia-calib` for crates.io.

This PR is retargeted to the solver hardening. Outstanding here:

**1. The homography seed is hardcoded to a canonical square (HIGH).** `Detection::estimate_pose` passes `[TR, BR, BL, TL]` — a cyclic shift — so at 20° tilt the public default (`n_iters = 50`) returns **3.74° rotation error and 9.17 px² reprojection error on perfect, noise-free input**, needing ~2000 iterations to converge. Both in-repo examples use 50. The existing guard only tests the frontal pose, the one configuration where the shift is harmless.

The fix that removes the problem rather than documenting it: seed the homography from the caller's own `object_pts` (centred and scaled) instead of the fixed `tag_norm`. That makes the solver order-independent and the existing caller correct for free.

**2. `PlanarPosePair::second` versus `kornia-calib` (MEDIUM).** Measured, `second` comes back **behind the camera with `error == f64::MAX`** at 0°, 20° and 45° tilt. But `kornia-calib` builds its planar 2-fold disambiguation on `second.pose` with no `z > 0` or error check — `init.rs:62`, `:224`, `:358` — and `:224` **averages** agreeing branches. Either the measurement is wrong or calib is scoring against a mirrored behind-camera pose. This must be resolved before `kornia-calib` is actually released.

Also worth noting the second solution is `R·diag(-1,-1,1)`, a 180° in-plane rotation — for a 2-fold-symmetric target that is not a new hypothesis at all, and not the classical mirror ambiguity. The AprilRobotics reference gates it on positive depth; this port does not.

**3. `SingularHomography` removal is a semver break (MEDIUM).** Splitting the error enum removes the only variant that ever existed, so any downstream `match` stops compiling. Nothing in-repo matches it, so CI stays green and the break is invisible until a user hits it. #1076 deliberately keeps the variant; whatever lands here needs either a deprecated alias for one release or an explicit callout.

**4. Lower priority:** `n_iters == 0` flipping `Ok` → `Err` through the re-export is a behaviour change worth release-noting; the doc sentence "the second is the first with its first two rotation columns negated" describes the *seed*, not the result (both branches run independent iterations and are then sorted); and leftover tag vocabulary in the now tag-agnostic module.

edgarriba:
Superseded by #1087.

#1076 landed the `kornia-apriltag` → `kornia-3d` move that made up the first half of this branch, so only the hardening commit was still relevant. Rather than rebase a branch whose diff is now half redundant, I rebuilt it on top of `main` and resolved the four review findings there:

1. **The homography seed is no longer hardcoded.** It is fitted from the caller's own `object_pts`, so the solver is order-, winding-, shape- and scale-independent. This branch only *documented* the hazard; #1087 removes it, which also fixes `Detection::estimate_pose` (measured 4.16° of rotation error on perfect input at `n_iters = 50`, now 0.0000°).
2. **`PlanarPosePair::second` is now `Option<PlanarPose>`**, cheirality-gated at the source. This branch documented it as degenerate while `kornia-calib` and the `apriltag` example kept consuming it as a real hypothesis; those are fixed too. It also corrects a false claim this branch made — the AprilRobotics reference does *not* gate solution 2 on positive depth; it solves a quartic for the genuine second local minimum, which this port has never done.
3. **The `PlanarPoseError` split is kept and the removal of `SingularHomography` is stated as a breaking change** in the PR body, with the evidence that no released path is affected (`kornia_3d::pose::planar` has never shipped; #1076 already deleted the `kornia-apriltag` path that had).
4. `n_iters == 0` still errors, now flagged as a public behaviour change; the "second is the first with its columns negated" sentence is corrected to describe the seed rather than the result.

Each behavioural fix has a regression test that was verified to fail when that fix alone is reverted. Closing this one.

PR #1087, opened 2026-08-05
fix(3d)!: harden the planar-pose solver — order-independent seed, gated second solution

edgarriba:
Supersedes #1074, which #1076 left stale: that PR already landed the `kornia-apriltag` → `kornia-3d` move, so all that is left here is the hardening, rebuilt on top of `main`.

Four review findings on the moved solver, resolved.

## 1. The homography seed was hardcoded, so the solver was corner-order dependent (HIGH)

`estimate_planar_pose` seeded its initial homography from a FIXED canonical square
`[(-1,-1), (1,-1), (1,1), (-1,1)]` mapped to `image_pts`. The orthogonal iteration then refines
against the caller's *real* `object_pts`, so any consistent correspondence converges eventually —
but only a caller whose winding and phase happened to match that hardcoded square started from the
right place.

Rotation error of `best` on exact, noise-free projections of a 10 cm square at 0.3 m / 20° tilt:

| `object_pts` order | old seed @ 50, 20° tilt | old seed @ 50, 20° tilt + 15° yaw | **new seed @ 50** |
|---|---|---|---|
| canonical `[(-s,-s),(s,-s),(s,s),(-s,s)]` | 0.0000° | 0.0000° | **0.0000°** |
| cyclic shift by one | 3.74° | 4.16° | **0.0000°** |
| reversed winding | 18.47° | 18.47° | **0.0000°** |

Every old-seed case converges by `n_iters = 500`; the problem is purely that 50 is what callers pass.

`kornia-apriltag`'s `[TR, BR, BL, TL]` is itself a cyclic shift of the canonical winding, so it sat
on that middle row. `Detection::estimate_pose` (`crates/kornia-apriltag/src/decoder.rs:318`) passes
it with `n_iters = 50`, so it was returning ~4° of rotation error on *perfect* input. Both in-repo
examples also use 50.

**Fix** — seed from the caller's own `object_pts`
(`crates/kornia-3d/src/pose/planar.rs`, `normalized_object_seed` + `estimate_planar_pose`): take
their x/y, centre on their centroid, rescale uniformly, and fit the homography from *those* to
`image_pts`.

Writing the exact projection as `H = K·[r₁ r₂ t]` and substituting `q = s·(p − c)` gives
`H' = K·[r₁/s r₂/s (t + R·c)]`: the first two columns keep their directions and pick up a *common*
factor, which `homography_to_pose` divides out when it normalises them. So the seed's `R` is the
caller's frame regardless of order, winding, shape or unit, and on exact data the seed already *is*
the answer. `t` is rewritten by the iteration's first step either way.

This fixes the AprilTag caller for free and retires the old "strongly non-square targets need more
iterations" caveat — a 4:1 rectangle and a trapezoid now converge at 50 too.

Honest note on the centring/rescaling: they are the standard DLT conditioning step, not the fix.
Measured — disabling either one on its own leaves every case in
`seed_is_independent_of_corner_order` passing. What is load-bearing is fitting from the caller's
points at all. The module docs say exactly this rather than claiming more.

## 2. `PlanarPosePair::second` was documented as degenerate but consumed as a real hypothesis (MEDIUM)

Measured across depths 0.3 / 1 / 5 m, tilts 0–75°, two yaws, and again with 1 px noise: **41 of 41
configurations returned a second pose behind the camera**, carrying `error == f64::MAX`. The docs
were right; the consumers were not.

Two things about how it was described were wrong, both corrected:

- It is **not** the classical planar mirror ambiguity. It is a re-seed with the first two rotation
  columns negated — a 180° in-plane rotation.
- The claim that "the AprilRobotics C reference gates solution 2 on positive depth" is **false**.
  `apriltag_pose.c` does something different and better: `fix_pose_ambiguities` finds the genuine
  second local minimum of the same objective by solving a quartic, and reports `HUGE_VAL` when
  there is none. It performs no cheirality check. This port has never computed that second minimum
  at all. Porting it is follow-up work and is now stated as such rather than implied by a claim of
  parity with the reference.

**Fix** — `second` is now `Option<PlanarPose>`, `Some` only when all four object points are in front
of the camera under it. That makes the misuse impossible by type rather than by convention, which is
the point: nothing in-repo was checking.

All five in-tree consumers used it with no depth or error check:

- `crates/kornia-calib/src/init.rs:62` — fed into the 2ⁿ branch-combination search.
- `crates/kornia-calib/src/init.rs:230` — `measure_tag_corners` anchors on the camera whose two
  branches disagree *most*, then has every camera contribute the branch closest to that anchor and
  **averages** them. With one branch always behind the camera, "most disagreeing" was arbitrary.
- `crates/kornia-calib/src/init.rs:367` — both branches scored for the board pose.
- `examples/apriltag/src/main.rs:371` — temporal disambiguation could lock a tag's displayed pose
  onto the behind-camera branch whenever its quaternion happened to match the previous frame better.
- `kornia-py/src/apriltag.rs:325` — returned straight to Python as the second element of
  `Detection.estimate_pose`'s tuple.

Each now falls back to `best` (calib) or skips the branch (example, Python binding).

The honest consequence: since `second` is `None` in every configuration measured, `kornia-calib`'s
planar 2-fold disambiguation is now a no-op. It always was — it was choosing between a good pose and
an impossible one. Its test suite passes unchanged.

## 3. `PlanarPoseError::SingularHomography` is removed, not deprecated (MEDIUM)

The old enum had exactly one variant, and it misreported its own cause: an orthogonal-iteration
failure — nothing to do with the homography — returned `SingularHomography`, while a genuine
homography failure had its cause discarded by `map_err(|_| …)`. It is now:

```rust
pub enum PlanarPoseError {
    Homography(#[from] HomographyError),  // the seed could not be fitted; cause preserved
    SingularIteration,                    // (I − avg_F) degenerate: the image rays coincide
    DegenerateObjectPoints,               // object points have no extent in their own plane
    ZeroIterations,                       // n_iters == 0
}
```

**No `#[deprecated] SingularHomography` shim, deliberately.** `kornia_3d::pose::planar` has never
been released: it first appears in `main` via #1076, and the last tag (`v0.1.15-rc.5`) still has the
module at `crates/kornia-apriltag/src/pose.rs`. The only path that ever shipped this variant was
`kornia_apriltag::pose::PlanarPoseError::SingularHomography`, and #1076 deleted that path outright.
A downstream `match` naming it is already broken by #1076 and must be edited regardless — at which
point the new variants are visible. A shim that can never be constructed would add a dead arm, not
a migration path.

**Breaking change (to an unreleased API).** Callers matching exhaustively on `PlanarPoseError` must
handle the new variants; `second` is now an `Option`. On the Python side,
`Detection.estimate_pose` now returns `(TagPose, TagPose | None)`.

## 4. Smaller items (LOW)

- **`n_iters == 0` now errors.** It used to return `Ok` with an unrefined pose carrying a sentinel
  `f64::MAX` error (verified: rotation `I`, translation `(0, 0, 7.07)` for the doctest's inputs).
  This is a behaviour change through the public API — a caller passing 0 now gets
  `Err(ZeroIterations)` where it previously got a useless `Ok`.
- **Doc corrected**: "the second is the first with its first two rotation columns negated" described
  the *seed*, not the result. Both branches run independent iterations and are then sorted by
  reprojection error, so `best`/`second` name outcomes, not branches.
- **De-tagged** the now tag-agnostic module: no more "tag corners", "tag pose estimation", `tag_norm`.
- **Added the `# Example` doctest** AGENTS.md requires for non-trivial public APIs.
- The `1e-10` depth floor is a named `CHEIRALITY_EPS` shared by the error accumulation and the new
  gate, so the two cannot drift apart.

## Tests

Four new tests in `crates/kornia-3d/src/pose/planar.rs`. Each was verified non-vacuous by reverting
*only* its fix and confirming the specific assertion fires:

| test | revert | assertion that fired |
|---|---|---|
| `seed_is_independent_of_corner_order` | restore the hardcoded `tag_norm` seed | `apriltag TR,BR,BL,TL: 4.15856950127979 deg at n_iters = 50` |
| `second_solution_is_gated_on_cheirality` | always return `second` | ``tilt 0 deg / 0.3 m: `second` returned with a point at z = -0.29999999999999916, behind the camera`` |
| `zero_iterations_is_rejected` | drop the `n_iters == 0` guard | `expected ZeroIterations, got Ok(… error: 1.797…e308 …)` |
| `degenerate_object_points_are_rejected` | drop the zero-extent check | `expected DegenerateObjectPoints, got Ok(… rotation: DVec3(NaN, NaN, NaN) …)` |

`seed_is_independent_of_corner_order` covers nine cases: the canonical order, the AprilTag
`[TR, BR, BL, TL]` order, a cyclic shift, a reversed winding, a 4:1 rectangle, a trapezoid, an
object frame whose origin is nowhere near the target, and the same target expressed at 10⁻³× and
10³× scale — the scale cases scale the depth with the target so the image is bit-identical, which
isolates the numerical question from the physical one (a 50 µm target at 0.3 m really is
unsolvable, and should not be dressed up as a scale bug).

`second_solution_is_gated_on_cheirality` asserts in both directions: that any `Some` really is in
front of the camera with a finite error, and — so the first loop cannot pass vacuously — that the
count of `Some` results over the 18 configurations is zero, with a message telling the next reader
to update the docs if that ever changes.

## Self-review corrections (commit 9364221f)

Running `/code-review` against this PR found real defects in it; they are fixed in the third commit:

- **Two doc claims were false.** The "2.56° / 15.58°" figures above were inherited from #1074 and
  shipped without re-measuring — the true values are 3.74° / 18.47°, and the table now states which
  rotation each column uses. Separately, `second_solution_is_gated_on_cheirality`'s doc claimed the
  sweep included "1 px noise" when the test used exact projections only; the noise measurement lived
  in an uncommitted scratch harness. The test now actually sweeps both, 36 configurations, and
  asserts the case count.
- **A degenerate tie in `measure_tag_corners`.** Substituting `best` for a `None` second branch made
  a camera's two branches identical, so the "anchor = the camera whose branches disagree MOST"
  selection saw every distance tied at 0 and `max_by` returned the last element — an arbitrary
  anchor. `Option` is now carried through. `init_poses` had the same substitution (wasteful rather
  than wrong) and is now indexed so only cameras with a real second branch consume a search bit.
- `PlanarPosePair::best` is documented as NOT cheirality-gated, since `second` being gated invites
  the opposite assumption.
- `1.0 / mean_r` is checked for finiteness (a subnormal mean radius overflowed the reciprocal).
- Comments in `kornia-calib` that this PR had invalidated are updated.

## Verification

```
cargo fmt --all -- --check                                         # clean (whole workspace)
cargo clippy -j2 -p kornia-3d -p kornia-calib --all-targets -- -D warnings   # exit 0
cargo test -j2 -p kornia-3d -p kornia-calib                        # 188 + 23 + 14 + 0 pass, 0 fail
cargo test -j2 -p kornia-apriltag --lib                            # 48 pass, 0 fail
cargo test -j2 -p kornia-3d --doc                                  # 14 pass (incl. the new example)
```

Not run on this machine: `kornia-apriltag`'s `parity` integration test, which links the system
`libapriltag` (`image_u8_create_stride` is missing from the aarch64 3.2.0 package here), and
`cargo clippy -p kornia-apriltag`, which trips a pre-existing aarch64-only NEON lint at
`segmentation.rs:881` — identical on `main`, neither touched by this PR.

### DIFF
--- a/crates/kornia-3d/src/pose/planar.rs
+++ b/crates/kornia-3d/src/pose/planar.rs
@@ -588,9 +588,13 @@ mod tests {
     /// metric scale must not change what `n_iters = 50` recovers from exact data.
     ///
     /// This is the test the module lacked: every other test here passes the canonical square in
-    /// the canonical order, which is exactly the ordering the old hardcoded `tag_norm` seed
-    /// happened to match, so none of them could observe the effect. With that seed restored,
-    /// `shifted` lands 2.56 deg out and `reversed` 15.58 deg out at 50 iterations.
+    /// the canonical order, which is exactly the ordering the old hardcoded canonical-square seed
+    /// happened to match, so none of them could observe the effect.
+    ///
+    /// With that old seed restored, under THIS test's pose (`rz(15 deg) * rx(20 deg)` at 0.3 m --
+    /// the right-hand column of the module table, not the plain-20-deg-tilt one),
+    /// `cyclic shift by one` lands 4.16 deg out and `reversed winding` 18.47 deg out at
+    /// `n_iters = 50`.
     #[test]
     fn seed_is_independent_of_corner_order() {
         let camera = test_camera();
--- a/crates/kornia-3d/src/pose/planar.rs
+++ b/crates/kornia-3d/src/pose/planar.rs
@@ -18,11 +18,20 @@
 //! encode; it is conditioning, not correctness.
 //!
 //! This matters because an earlier revision seeded from a FIXED canonical square
-//! `[(-1,-1), (1,-1), (1,1), (-1,1)]`, which was only correct for callers using that exact
-//! winding and phase. Measured on exact projections of a square at 0.3 m / 20 deg tilt, that
-//! seed left a cyclic shift by one 2.56 deg out and a reversed winding 15.58 deg out at
-//! `n_iters = 50` (both converged only by ~500), and left `kornia-apriltag`'s
-//! `[TR, BR, BL, TL]` corner order 4.16 deg out on perfect input.
+//! `[(-1,-1), (1,-1), (1,1), (-1,1)]`, which was only correct for callers whose points shared
+//! that exact winding and phase. Rotation error of `best` against that old seed, on exact
+//! projections of a 10 cm square at 0.3 m, at `n_iters = 50` (every case converges by 500):
+//!
+//! | `object_pts` order | 20 deg tilt | 20 deg tilt + 15 deg yaw |
+//! |---|---|---|
+//! | canonical | 0.0000 deg | 0.0000 deg |
+//! | cyclic shift by one | 3.74 deg | 4.16 deg |
+//! | reversed winding | 18.47 deg | 18.47 deg |
+//!
+//! `kornia-apriltag`'s `[TR, BR, BL, TL]` is itself a cyclic shift of the canonical winding, so
+//! it sat on that middle row: ~4 deg of rotation error on PERFECT input, at the `n_iters = 50`
+//! every in-tree caller passes.
+//!
 //! `seed_is_independent_of_corner_order` pins the current behaviour: every ordering, a 4:1
 //! rectangle, a trapezoid, an off-origin object frame and the same target expressed at 1e-3x and
 //! 1e3x scale all land under 1e-6 deg at `n_iters = 50`.
@@ -43,11 +52,12 @@
 //! (`fix_pose_ambiguities`) and reports `HUGE_VAL` when there is no second minimum. Porting that
 //! is tracked separately.
 //!
-//! Because of that, the second branch is usually not a real hypothesis: across depths of
-//! 0.3--5 m, tilts of 0--75 deg and both noise-free and 1 px-noise inputs, it converged BEHIND
-//! the camera in every configuration measured. It is now gated on cheirality at the source, so
-//! `second` is `Some` only when all four object points lie in front of the camera under it --
-//! which, in practice, means callers get `None` and can stop hand-rolling the check.
+//! Because of that, the second branch is usually not a real hypothesis: across the 36
+//! configurations `second_solution_is_gated_on_cheirality` sweeps (depths 0.3/1/5 m, tilts
+//! 0--75 deg, each noise-free and with 1 px of corner noise) it converged BEHIND the camera in
+//! every one. It is now gated on cheirality at the source, so `second` is `Some` only when all
+//! four object points lie in front of the camera under it -- which, in practice, means callers
+//! get `None` and can stop hand-rolling the check.
 //!
 use crate::camera::PinholeCamera;
 use crate::pose::{homography_4pt2d, Pose3d};
@@ -68,6 +78,13 @@ pub struct PlanarPose {
 #[derive(Debug, Clone)]
 pub struct PlanarPosePair {
     /// Lower-error solution.
+    ///
+    /// NOT cheirality-gated, unlike [`Self::second`]. `best` is whichever branch has the lower
+    /// reprojection error, and that error only *penalises* points behind the camera (a sentinel
+    /// term per point) rather than excluding them — so if BOTH branches put part of the target
+    /// behind the camera, the less-bad one is still returned here. Callers that must not act on
+    /// an impossible pose should check `best.error.is_finite()` and the sign of
+    /// `best.pose.transform_point(..).z`.
     pub best: PlanarPose,
     /// A second local minimum, if the second branch produced a physically possible pose.
     ///
@@ -147,10 +164,12 @@ fn normalized_object_seed(object_pts: &[Vec3F64; 4]) -> Option<[[f64; 2]; 4]> {
         .map(|p| ((p.x - cx).powi(2) + (p.y - cy).powi(2)).sqrt())
         .sum::<f64>()
         * 0.25;
-    if !mean_r.is_finite() || mean_r <= 0.0 {
+    let s = 1.0 / mean_r;
+    // `mean_r` being finite and positive is not enough: a subnormal one still overflows the
+    // reciprocal to infinity, which would put inf coordinates into the homography fit.
+    if !s.is_finite() || s <= 0.0 {
         return None;
     }
-    let s = 1.0 / mean_r;
     Some([0, 1, 2, 3].map(|i| [(object_pts[i].x - cx) * s, (object_pts[i].y - cy) * s]))
 }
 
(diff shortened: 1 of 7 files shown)


edgarriba:
### Code review

Self-review of this PR. Found 4 issues; all are fixed in 9364221f.

1. Two measured claims in the docs were false. The "2.56 deg" cyclic-shift and "15.58 deg" reversed-winding figures were inherited from #1074 and shipped without re-measuring. Reproducing the old canonical-square seed gives **3.74 deg** and **18.47 deg** at 20 deg tilt; the "4.16 deg" AprilTag number only holds under the compound `rz(15)*rx(20)` the test actually uses, not the plain "20 deg tilt" the sentence claimed. AprilTag's `[TR, BR, BL, TL]` is also just a cyclic shift, so it was never a separate measurement.

https://github.com/kornia/kornia-rs/blob/c30ed79ba33aad4fcda74909b4359e68d1e80e3c/crates/kornia-3d/src/pose/planar.rs#L19-L27

2. `second_solution_is_gated_on_cheirality`'s doc claimed the sweep covered "1 px noise", but the test used exact projections only — the noise measurement existed solely in an uncommitted scratch harness, so nothing in CI could falsify it. The test now sweeps each depth/tilt both noise-free and with 1 px of deterministic corner noise (36 configurations) and asserts the case count.

https://github.com/kornia/kornia-rs/blob/c30ed79ba33aad4fcda74909b4359e68d1e80e3c/crates/kornia-3d/src/pose/planar.rs#L676-L681

3. Substituting `best` for a `None` second branch made a camera's two branches identical, feeding a 0-distance tie into the anchor selection three lines below ("the camera whose two branches disagree MOST"). With every distance tied at 0, `Iterator::max_by` returns the **last** element, so the anchor became an arbitrary, iteration-order-dependent camera. `Option` is now carried through, so only cameras that really have a second branch can anchor. `init_poses` had the same substitution — wasteful (2^n identical scorings) rather than wrong — and is now indexed so only those cameras consume a search bit.

https://github.com/kornia/kornia-rs/blob/c30ed79ba33aad4fcda74909b4359e68d1e80e3c/crates/kornia-calib/src/init.rs#L229-L237

4. `PlanarPosePair::best` is not cheirality-gated while `second` is, which invites the assumption that it is — the reprojection error only *penalises* behind-camera points rather than excluding them, so if both branches are partly behind the camera the less-bad one is still returned as `best`. Now documented on the field. Also `1.0 / mean_r` was unchecked for finiteness (a subnormal-but-finite mean radius passes the guard above it and overflows the reciprocal to infinity).

https://github.com/kornia/kornia-rs/blob/c30ed79ba33aad4fcda74909b4359e68d1e80e3c/crates/kornia-3d/src/pose/planar.rs#L79-L82

Not changed: the unguarded `fx`/`fy` divisions flagged by Qodo are pre-existing (identical at `21457087`), and AGENTS.md explicitly permits `.expect()` in test code.

🤖 Generated with [Claude Code](https://claude.ai/code)

<sub>- If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

PR #1090, opened 2026-08-06
fix(imgproc): make CUDA SIFT keypoint selection deterministic

edgarriba:
## The bug

Two runs of the same binary on the same image keep **different keypoints** whenever a frame detects
more extrema than `max_keypoints`.

The extrema kernel appends through `atomicAdd(counter, 1)` and drops whatever lands past the buffer
(`detect.rs:331`). The buffer was sized *at* `max_keypoints`, and `plan.rs` then took the leading
`max_keypoints` entries — of a list ordered by thread arrival. So which keypoints survive is a
property of GPU scheduling.

Not an edge case: on a real clip at the default 4096 budget, one frame produced **16,847 extrema**.

Downstream in an SfM pipeline this was severe — two runs with identical inputs and flags gave
40,878 vs 40,795 points and **11.86 vs 3.28 px reprojection RMSE**, the first rejected by a quality
gate and the second kept.

Note this is *not* the failure that `pairs.sort_unstable()` fixed in the matcher: sorting
canonicalises the **order** of a truncated set, never its **membership**. The CPU path was never
affected — `retain_best_order` has always selected by response there.

## The fix

- Detection buffer gets **8x** `max_keypoints` of headroom, so the detector can report what it
  actually found instead of the surplus vanishing on device where nothing can rank it. 8x covers the
  measured worst case with margin, at ~1.2 MB of device memory at the default budget.
- New `select_best_keypoints` applies the budget on the host by **descending response**, with
  position and size breaking ties so the order is total — the reference's rule, and the same one the
  CPU path already uses.
- Survivors are written back compacted in that canonical order, because orientation consumes
  `kp[0..n]` positionally and its output order feeds matching and then track building, both
  index-sensitive.
- Exceeding even 8x still drops by arrival, and now warns once per process instead of silently.

`total_cmp` rather than `partial_cmp`: a NaN response makes a comparator inconsistent and leaves the
sort order *unspecified*, which would reintroduce the bug exactly when the data is degenerate.

When the budget does not bind, the selection returns early — arrival order is then a permutation of
a fixed set, and canonicalising it would change every under-budget frame's output while fixing
nothing.

## Also included

`matcher_is_deterministic_across_reuse` — a streaming caller reuses one `SiftMatcher` across a whole
clip, so its device buffers carry the previous frame's residue. The test matches A/B, poisons the
buffers with a differently-sized frame, and re-matches, asserting the pair list is unchanged.
Matching A/B twice in a row would pass even with a tail-read bug; the interleaved frame is what
makes it real. It passes as shipped — recorded because it was a live hypothesis for this failure and
it took a test to rule out.

The fixture plants true correspondences rather than using two independent random sets: uniform
random 128-D descriptors sit at nearly the same distance from each other (min d1/d2 = 0.87), so a
0.7 ratio test rejects everything and the assertion would hold vacuously on an empty list.

## Verification

40/40 CUDA SIFT tests pass, including `budget_caps_the_count_and_keeps_the_strongest`. On the
pipeline that surfaced this, the keypoint fingerprint is now identical across runs where it
previously differed on every build.

### DIFF
--- a/crates/kornia-imgproc/src/cuda/sift/matcher.rs
+++ b/crates/kornia-imgproc/src/cuda/sift/matcher.rs
@@ -511,6 +511,64 @@ mod tests {
     }
 
     #[test]
+    /// The same descriptors must match to the same pairs, every time, including when OTHER frames
+    /// are matched in between.
+    ///
+    /// A streaming caller reuses one `SiftMatcher` across a whole clip, so its device buffers carry
+    /// the previous frame's residue. If any kernel reads past the current `nq`/`nt` — or if the
+    /// pairs kernel's `atomicAdd` truncation is ever reachable — the surviving SET depends on what
+    /// ran before, and a reconstruction stops being reproducible. Measured on a real clip: identical
+    /// keypoints and identical descriptors produced DIFFERENT match sets across two runs of the same
+    /// binary, which is what this pins.
+    ///
+    /// The interleaved match with different sizes is the point: matching A/B twice in a row would
+    /// pass even with a tail-read bug, because the residue would happen to be the same data.
+    #[test]
+    fn matcher_is_deterministic_across_reuse() {
+        let stream = default_stream();
+        let ctx = &stream.context();
+        let (n1, n2) = (257usize, 193usize);
+        let a = rand_desc(n1, 999);
+        // Plant true correspondences: uniform random 128-D descriptors all sit at nearly the same
+        // distance from each other (measured elsewhere in this crate: min d1/d2 = 0.87), so a ratio
+        // test at 0.7 would reject every pair and the test would pass vacuously on an empty set.
+        // Copying a's rows with a one-component perturbation gives each of them a genuine nearest
+        // neighbour and a distant runner-up, which is what the ratio test is for.
+        let mut b = rand_desc(n2, 4242);
+        for i in 0..n2 {
+            b[i * DESCR_LEN..(i + 1) * DESCR_LEN]
+                .copy_from_slice(&a[i * DESCR_LEN..(i + 1) * DESCR_LEN]);
+            b[i * DESCR_LEN] += 1.0;
+        }
+        // Deliberately larger, so it leaves residue beyond n1/n2 in every buffer.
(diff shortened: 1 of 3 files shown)


edgarriba:
Review feedback addressed in `1f7a0447`. Summary for whoever picks this up, since CI is currently
backed up repo-wide (18 of the last 20 runs queued, including jobs on `main`) and the earlier red
checks were GitHub Actions failing to download the actions themselves — `Service Unavailable` /
`Failed to resolve action download info` — not this branch.

**The important one:** the new test had been inserted between `matcher_matches_host_brute_force`'s
`#[test]` and its `fn`, so the new function absorbed that attribute and carried its own. The result
was the new test registering twice and the brute-force parity check — the matcher's strongest
correctness test — silently never running. Fixed; the suite now reports 9 tests, each once, with
brute-force parity passing.

Also fixed: the three lints `cargo clippy --tests -- -D warnings` would have rejected
(`duplicate_macro_attributes`, `dead_code`, `unused_mut`); rustfmt; and several doc figures that were
wrong by 2-16x because they assumed a 4096 default when `SiftCudaConfig::max_keypoints` defaults to
8192 (the detection buffer is 2.36 MB at default, and the host round trip is up to 2.36 MB per
octave with a 65k-element sort — the cost note now says so plainly).

The orientation-stage overflow raised in review is real and is NOT fixed here: `orientation.rs`
appends via `atomicAdd` and returns on `slot >= max_out`, with `ori_count` accumulating across all
octaves, so the same arrival-order truncation survives one stage later. It now warns once per
process instead of clamping silently. Fixing it properly needs the same treatment this PR applies to
detection — capacity to report everything, then deterministic host-side selection — and is better as
its own change than bolted onto this one.

Verified locally by running the exact CI commands on this commit:

```
cargo clippy --tests -p kornia-imgproc --features cuda -- -D warnings   clean
cargo fmt --all -- --check                                             clean
cargo test -p kornia-imgproc --features cuda --lib cuda::sift          41 passed
```

Downstream evidence: on the SfM pipeline that surfaced this, keypoints are now bit-identical across
three consecutive builds where they previously differed on every one.

PR #1094, opened 2026-08-07
fix(3d): compare the true robust cost, and damp ellipsoidally, in the Schur BA

edgarriba:
## Summary

Three defects in `bundle_adjust_schur`, all visible from the LM loop, plus the counters that
made them measurable.

### 1. The accept test compared an IRLS surrogate, not the objective

The solver scales the residual and Jacobian rows by `√w`, `w = ρ'(s)`, `s = ‖r‖²`. That part is
right — it makes the normal equations those of the robust problem. But accumulating that scaled
residual gives `½ρ'(s)·s`, which is **not** the cost being minimised. For Huber past the knee the
surrogate is `½k‖r‖` while the loss is `k‖r‖ − k²/2`, so it moves at exactly half the true rate:

```
ρ'(s) = k/√s,  ρ''(s) = −k/(2 s^{3/2})   ⇒   ρ''·s = −ρ'/2
d(½ρ'(s)·s) = ½(ρ' + ρ''·s)·ds = ½ · d(½ρ(s))
```

So every step's measured reduction was halved on downweighted observations, and the `cost` the
solver reports and terminates on was not the function it minimises. Both passes now accumulate
`robust_cost` = `½ρ(s)`, which is what Ceres compares.

`robust_weight` and `robust_cost` are lifted to module level as a pair, so the weight and the loss
it is the derivative of live next to each other — they diverged in the first place because they
didn't.

### 2. LM damping was spherical

`A[i] += λ·I` damps every direction by the same **absolute** amount whatever its curvature. One λ
has to serve parameter blocks in different units — rotation in radians, translation and points in
metres — so it over-damps stiff directions and under-damps soft ones, and λ is not dimensionless.

Now ellipsoidal, as Ceres' `LevenbergMarquardtStrategy` does it: damp with `diag(JᵀJ)` (already
assembled here as the block diagonals), clamped to `[min_lm_diagonal, max_lm_diagonal]` =
`[1e-6, 1e32]`.

### 3. The doc was wrong in both directions

It claimed `BaParams::robust` was unimplemented (it is implemented) and that `gradient_tolerance`
was honoured (it is never read). Both corrected to describe what the solver actually does.

## Tests

**Correction to an earlier version of this description.** It claimed both new tests fail if the
surrogate is reintroduced, "verified by putting it back". That verification was invalid: I
reintroduced the surrogate inside the `robust_cost` helper, which those tests assert on directly, so
they failed for the wrong reason. The real defect was at the **call sites**. Restoring the surrogate
there (`cost += 0.5*(r[0]*r[0]+r[1]*r[1])` and `new_cost += 0.5*w*r_sq`) leaves all tests passing.

So, stated accurately:

- `robust_weight_is_the_derivative_of_robust_cost` — pins `d/ds[½ρ(s)] = ½·w(s)` by finite
  difference across the knee, for Identity / Huber / Cauchy / Tukey.
- `huber_surrogate_moves_at_half_the_true_rate` — pins the factor of 2.
- `robust_cost_is_half_the_shared_kornia_algebra_rho` — pins these helpers against
  `kornia_algebra::optim::losses::RobustLoss::rho`.
- `cauchy_cost_is_accurate_for_small_residuals` and
  `infinite_robust_scale_collapses_to_l2_instead_of_nan` — both verified to fail when their fix is
  reverted; the second is solver-level.

These pin the **helpers**. **No test yet catches the original call-site regression** — that gap is
open and I would rather say so than leave the earlier claim standing. The obstacle is that
`BaResult` exposes no cost, so the objective the solver actually compared is not observable from
outside; closing it properly probably means surfacing the final cost.

193 pass; `cargo fmt --check` and `cargo clippy --all-targets -- -D warnings` clean.

## Measured

**Read these with the following caveat.** They come from a downstream tree that adds a Nielsen
gain ratio to this solver. **This tree computes no gain ratio** — the accept test reads only the
*sign* of the cost change, and λ moves by fixed `/3` and `*10`. In that downstream tree the halved
reduction pinned ρ at **0.4157 for 90 consecutive iterations**, so λ could never decay (Nielsen
shrinks λ only when ρ > 0.5) and LM never handed back to Gauss-Newton.

**In this tree the same defect has a narrower effect**: the cost being compared is not the objective,
so accept/reject can take the wrong branch whenever the surrogate and the true loss disagree in the
*sign* of the change, `cost_tolerance` terminates on a quantity that is not the objective, and the
reported cost is not what is being minimised. Correctness, not the dramatic convergence numbers
below.

| | before | after |
|---|---|---|
| terminal adjustment | 200 iters, `converged=false` | **57 iters, `converged=true`** |
| bundle adjustment total | 451.8 s | **191.2 s** |
| reconstruction phase | 488.4 s | **228.4 s** |

The answer also **improves**, checked against a COLMAP reconstruction of the same clip rather than
assumed:

| vs COLMAP | before | after |
|---|---|---|
| ATE p50 | 0.250 m | **0.116 m** |
| ATE p90 / max | 0.384 / 1.065 | **0.195 / 0.272** |
| out-of-plane jitter | 0.038 | **0.015** (COLMAP's own: 0.015) |
| trajectory flatness | 0.039 | **0.036** (COLMAP's own: 0.045) |

The two changes differ in kind: the gain-ratio fix changes the answer (for the better, above);
the damping change is accuracy-neutral — same minimum, roughly half the iterations to reach it
(ATE 0.117 → 0.116, flatness and jitter identical to three decimals).

## Not changed, having checked Ceres

The Triggs curvature correction does **not** apply to a Huber problem. Ceres' `Corrector` sets
`alpha = 0` whenever `rho[2] <= 0` — *"applying the curvature correction when rho'' < 0, which is
the case when we are in the outlier region slows down the convergence of the algorithm
significantly"* — and Huber has `ρ'' ≤ 0` everywhere. Plain `√ρ'` IRLS, as here, is what Ceres does
for Huber; the true robust cost is what it pairs that with.

---

## Review round

*(An adversarial review ran against an earlier revision that also carried the counters; the counter-specific fixes moved with them.)*

An adversarial review of this PR found 15 issues; fixes are pushed. The ones worth a maintainer's
attention:

- **Cauchy cost lost precision in f32.** `(1.0 + r_sq/s2).ln()` quantises `1.0 + x` before the log —
  28.6% low at s2=5.99/r_sq=1e-6, and exactly 0.0 for any `r_sq/s2 < 6e-8`, i.e. precisely the
  converged regime the accept test has to resolve. Now `(r_sq/s2).ln_1p()`.
- **`robust_scale_sq = f32::INFINITY` (the documented default) produced NaN cost** under
  Cauchy/Tukey, so every step was rejected and the solver returned its input poses with `Ok`.
  `ba::build_robust_loss` already guards this; `ba_schur` did not. Now collapses to Identity, with a
  solver-level test.
- **Ellipsoidal damping silently redefines `initial_lambda`** from absolute to relative, and
  `ba::bundle_adjust` reads the *same* `BaParams` while still damping by `λ·I`. Documented in the
  function docstring. A maintainer may prefer a separate field or a units change — happy to follow.
- **The λ floor `1e-8` is a bit-exact no-op in f32 under relative damping** (`x + 1e-8*x == x` for
  every magnitude tested), so LM degenerated to undamped Gauss-Newton after ~11 accepted steps.
  Raised to `1e-7`.
- **`MAX_LM_DIAGONAL = 1e32`** is Ceres' *f64* constant; `1e32_f32 * 1e7_f32` overflows to `inf` and
  poisons the solve with NaN. Reduced to `1e24`.
- **Depth residual used `/sigma` in one pass and `*inv_sigma` in the other**, so the two evaluations
  of the same objective were not bit-identical, with a one-sided bias toward accept. Pre-existing,
  but inside the function this PR rewrites. Now symmetric.

**Known and deliberately not addressed here**, flagged for your call:

- `RobustKernelKind::Tukey` silently runs **Cauchy** in these helpers, while `ransac/kernels.rs`
  dispatches the same enum to a real hard redescender. Fixing it changes results for existing Tukey
  callers, so it is documented rather than changed.
- These helpers duplicate `kornia_algebra::optim::losses::RobustLoss::{weight, rho}`. The right fix
  is static-dispatch `rho()` on the shared trait plus `pub(crate) build_robust_loss`; `Arc<dyn>` in
  the innermost LM loop would be a measurable regression on Jetson-class hardware. That is a design
  call I did not want to make unilaterally.
- The counters and phase timers that were previously the second commit here are now a **separate,
  stacked PR**, per `AGENTS.md` "one concern per PR". This PR is one concern: the cost the accept
  test compares, and the metric it damps with.

### DIFF
--- a/crates/kornia-3d/src/ba.rs
+++ b/crates/kornia-3d/src/ba.rs
@@ -163,6 +163,16 @@ pub struct BaResult {
     pub iterations: usize,
     /// Whether the optimizer converged.
     pub converged: bool,
+    /// Value of the objective at the returned solution — `Σ ½ρ(‖r‖²)` over every residual the
+    /// solver scored, with `ρ` the robust loss selected by [`BaParams::robust`] (identity when
+    /// none). For an unrobustified problem this is the familiar `½Σ‖r‖²`.
+    ///
+    /// This is the quantity the optimiser minimises and the quantity its accept test compares,
+    /// so it is the only way a caller can check that the two agree. That matters more than it
+    /// sounds: this solver previously scored an IRLS surrogate `½ρ'(s)·s` instead of `½ρ(s)`,
+    /// which is half the true value on every downweighted observation, and the discrepancy was
+    /// invisible from outside because nothing reported the cost.
+    pub final_cost: f32,
 }
 
 // ---------------------------------------------------------------------------
@@ -806,6 +816,7 @@ pub fn bundle_adjust(
         points: out_points,
         iterations: result.iterations,
         converged,
+        final_cost: result.final_cost,
     })
 }
 
--- a/crates/kornia-3d/src/ba_schur.rs
+++ b/crates/kornia-3d/src/ba_schur.rs
@@ -33,9 +33,14 @@
 //!   * Point parameters are the 3-dim world coordinates.
 //!   * z is clamped to `MIN_Z` to handle mid-iteration cheirality flips.
 //!
-//! Currently supports: identity loss only, fixed-pose anchors, fixed-point
-//! gauge (motion-only BA). Robust kernels and full LM-with-backtracking
-//! are TODO.
+//! Supports fixed-pose anchors, fixed-point gauge (motion-only BA), optional
+//! per-observation depth residuals, optional per-pose translation priors, and
+//! the robust kernels in `BaParams::robust` via IRLS (see
+//! [`bundle_adjust_schur`]). LM damping is ellipsoidal — `λ·diag(JᵀJ)`, as
+//! Ceres does it — so `λ` is dimensionless here, unlike
+//! [`crate::ba::bundle_adjust`], which damps by `λ·I`. Full
+//! LM-with-backtracking (a gain-ratio / trust-region update) is still TODO:
+//! the step test is the sign of the cost change only.
 
 use faer::prelude::Solve;
 use faer::Mat;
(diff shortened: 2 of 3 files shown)


edgarriba:
Note: the counters and phase timers that were originally the second commit here are now split out into #1095, stacked on this branch, per `AGENTS.md` "one concern per PR". This PR is now a single commit and a single concern.

PR #1102, opened 2026-08-13
feat(calib): a sequential-capture preset for the triangulation gate

edgarriba:
`CalibConfig::sequential()` lowers `min_parallax_deg` for video and walkthrough capture, where the general-purpose default is not merely suboptimal but fatal.

## The failure

The bootstrap pair is chosen by **most shared tracks**. For a rig of cameras pointed at a common subject that is a fine proxy for a well-conditioned pair. For forward-motion video it selects two **adjacent frames** — the smallest baseline in the entire sequence, because adjacency is exactly what maximises shared tracks.

`min_parallax_deg` then rejects nearly every triangulation from that pair, so the seed cloud comes back near-empty. The growth loop's candidate test needs an unplaced camera sharing >= 4 triangulated points; with an empty seed cloud none qualifies, so it breaks on its first iteration having placed only the anchor, and BA fails with `NoFreeVariables`.

Measured downstream on a 459-keyframe phone walkthrough, and reproduced at 120 keyframes.

## Why the gate is right in general and wrong here

Small-parallax points genuinely have unconstrained depth — the gate earns its place. It is wrong specifically **for the bootstrap**, where the alternative to an ill-conditioned scaffold is no scaffold at all. Registration only needs something good enough to place the next camera, and bundle adjustment refines it afterwards.

A caller who wants a clean *output* cloud should filter `Reconstruction::points` by triangulation angle at the point of consumption — a separate decision from what the solver is allowed to register against.

## Test

`sequential_preset_rescues_a_forward_walk` builds six cameras marching along -Z with landmarks well ahead, so consecutive views share the most tracks and have the least baseline. It asserts the preset places >= n-1 views AND that the **default config places strictly fewer** — so it cannot pass on a no-op preset.

35 lib tests + 2 doctests pass; clippy clean.

## Not addressed here

The deeper fix is the seed-pair heuristic itself: for sequential input, "most shared tracks" is close to an anti-heuristic for conditioning. Choosing a pair with adequate baseline among those with sufficient shared tracks would remove the need for this preset. That is a larger change and wants its own measurement.

### DIFF
--- a/crates/kornia-calib/src/sfm.rs
+++ b/crates/kornia-calib/src/sfm.rs
@@ -2054,4 +2054,81 @@ mod tests {
             refined.reproj_rmse_px
         );
     }
+
+    /// A forward-walking sequence is exactly the case the general defaults mishandle.
+    ///
+    /// Cameras march along -Z looking forward, so consecutive views share the most tracks AND have
+    /// the smallest baseline — the bootstrap picks the worst-conditioned pair available. At the
+    /// default `min_parallax_deg` every triangulation from it is rejected, growth finds nothing to
+    /// register against, and the solve dies with `NoFreeVariables`. `sequential()` is the
+    /// difference between a reconstruction and no reconstruction, not a quality tweak.
+    #[test]
+    fn sequential_preset_rescues_a_forward_walk() {
+        let n_cams = 6;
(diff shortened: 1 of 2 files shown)


edgarriba:
Folded into #1103, which contains both commits — the preset belongs on `ReconstructionConfig` (a rig config can never be sequential), so splitting the config and adding the preset are one change, not two.

PR #1108, opened 2026-08-14
feat(3d): point-to-plane ICP and an RGB-D odometry front end

edgarriba:
`registration` has vanilla point-to-point ICP. This adds the point-to-plane variant and the RGB-D odometry pipeline built on it. Both have been running in anger for handheld odometry on a Jetson Orin, at roughly **25 ms per solve on CPU**.

## Why point-to-plane

Point-to-point minimises the distance between corresponded points, which resists the tangential motion that indoor scans are full of: two views of the same wall are related by a slide the metric fights. Point-to-plane minimises distance along the target normal instead, letting surfaces slide over each other, so plane-dominated scenes converge in far fewer iterations.

## What's here

| file | |
|---|---|
| `registration/icp.rs` | the solver — normal estimation, correspondence rejection, Huber weighting, a 6x6 Cholesky on the normal equations, and an optional rotation prior contributing `1/sigma^2` to the rotation block alone |
| `registration/rgbd.rs` | depth-image front end: unproject, pyramid, per-level solve |
| `registration/synth.rs` | test-only synthetic scenes shared by both modules' tests |
| `examples/icp_bench.rs` | benchmark |
| `examples/icp_debug_pair.rs` | single-pair probe for debugging a bad alignment |

Additive alongside `icp_vanilla`; nothing existing changes behaviour.

## Tests

13 new (9 in `icp.rs`, 4 in `rgbd.rs`), on synthetic scenes with known ground-truth transforms. `cargo test -p kornia-3d` passes 215 + 14; clippy and rustfmt clean. **No new dependencies.**

### DIFF
--- a/crates/kornia-3d/examples/icp_bench.rs
+++ b/crates/kornia-3d/examples/icp_bench.rs
@@ -1,0 +1,201 @@
+//! Times one full 320x180, 3-level projective point-to-plane ICP solve.
+//!
+//! Renders a synthetic corner-and-sphere scene from two nearby camera poses
+//! (mirroring the in-crate test renderer), builds both pyramids, then times
+//! `icp_projective_plane` alone. Run in release:
+//!
+//! ```sh
+//! cargo build --release -p kornia-3d --example icp_bench -j1
+//! ./target/release/examples/icp_bench
+//! ```
+
+use std::time::Instant;
+
+use kornia_3d::registration::{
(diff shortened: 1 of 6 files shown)


edgarriba:
Converting to draft. This code is still in-flight as part of an RGB-D odometry effort that has not finished its live validation, so the API may still move; it should not land until its owner is done with it. Opened prematurely on my side — the branch had no PR because it wasn't ready, not because it had been forgotten. The port onto `main` is clean and the tests/lints pass, so it can be marked ready whenever that work settles.

PR #1109, opened 2026-08-14
feat(calib,3d): log-depth residual, sparse reduced camera system, track completion, and solve instrumentation

edgarriba:
The depth residual is absolute — `r = (z - m) / σ`, with `σ` in metres. A monocular depth network's error is not shaped like that; it is roughly proportional to depth. So the same fractional error costs ten times more at 10 m than at 1 m, and far structure dominates the solve.

This adds the log form

```
r = ln(z / (s·m)) / σ
```

where `σ` is a **fraction** of depth and the residual is comparable near and far, together with the per-view scale `s` that a relative residual needs to be meaningful.

**Off by default** (`depth_log_residual: false`). Existing callers are bit-unchanged — the 216 pre-existing `kornia-3d` tests pass untouched.

## API

Three `BaParams` fields, all inert unless the flag is set:

| field | |
|---|---|
| `depth_log_residual` | select the log form |
| `depth_scale_prior` | weight `λ` of the shrinkage toward `s = 1`; **negative freezes** the scales at their seed |
| `depth_scales_init` | starting value only — a robust median fit is the intended seed |

`BaResult` gains `depth_scales`, empty unless log mode ran.

## Two details that are easy to get wrong

**The `z ≤ 0` branch is the log residual's own first-order expansion about `z = s·m`**, not a bare `z - s·m`. Matching value *and* slope at the changeover keeps the cost C¹ there, so a point crossing the image plane mid-iteration doesn't hand LM a step discontinuity. `depth_residual_log_branches_match_at_changeover` pins both.

**The prior shrinks toward `s = 1`, deliberately not toward the seed.** Callers typically re-fit the seed against the current geometry before every solve, so shrinking toward the seed means "stay near wherever the geometry already is" — the prior tracks drift instead of resisting it, and the scales float free exactly as if `λ` were 0. Measured on a 365-keyframe walk, seed-anchored `λ = 1` doubled the map (54.8 m → 108 m of trajectory, vertical extent 5.0 → 13.4 m) **while every scale-invariant metric improved**: the signature of a reconstruction whose shape is fine and whose gauge has come loose. An anchor recomputed from the thing it anchors is not an anchor. `schur_ba_scale_prior_anchors_absolutely_not_to_seed` is the regression test.

## Why the scale is a closed form rather than another LM variable

The log residual is **linear in `ln s_i`**, so each view's block is a one-variable weighted least squares whose exact minimiser is a shrunk weighted mean. Alternating that exact block update with the LM step is block-coordinate descent on the joint objective: it reaches the same stationary point as folding `s_i` in as a 7th camera DOF, without widening every 6×6 block in the Schur reduction.

## Also

- Dedups the mid-iteration cheirality clamp, previously inlined identically at both depth sites, into `clamped_z`.
- Routes the trial-cost pass through the same `depth_residual` call as the linearisation pass, so accept/reject scores one objective. (These had been separate expressions; even `x / s` vs `x * (1.0 / s)` differs by an ulp in f32 — one-sided — which biases the accept test.)

219 tests pass (3 new), clippy and rustfmt clean, no new dependencies.

---

# Second commit: sparse reduced camera system by default in SfM

The sparse reduced camera system landed in #1105 and **`reconstruct` has never used it.** `sfm.rs` builds its `BaParams` with `..Default::default()`, and that default is `false` — correct for a rig, exactly wrong for the long sequential capture the sparse path was written for.

Two cameras occupy a nonzero 6×6 block of the `6P × 6P` reduced system only if they share a point. On a walkthrough, cameras a few hundred frames apart share nothing, so the dense path pays `O(dim²)` memory and `O(dim³)` factorisation for structure that isn't there. At 900 keyframes that's a dense 5400×5400 matrix, ~233 MB, ~5×10¹⁰ flops per iteration.

Sequential synthetic, 10 LM iterations, **identical cost to 5 significant figures and identical iteration counts**:

| views | dense | sparse | |
|---|---|---|---|
| 100 | 0.11 s | 0.02 s | 5.5× |
| 300 | 1.20 s | 0.08 s | 15× |
| 600 | 8.34 s | **0.13 s** | **64×** |

The gap widens with camera count — dense grows roughly cubically, sparse roughly linearly — and sparse was never slower at any size measured, down to 8 views.

Exposed as `ReconstructionConfig::sparse_reduced_system` (default `true`) rather than hardcoded, so the knob is discoverable where callers already look, documented with the measurement and with the one reason to turn it off: a genuine rig, where every camera sees the same subject and the block pattern really is dense. `CalibConfig`'s rig path and `BaParams::default()` are both untouched.

`sparse_reduced_system_does_not_change_the_reconstruction` pins the end-to-end result — same points to 1e-9, same rmse to 1e-6, same registration. It asserts equality rather than comparable quality, because this is a different storage of the same matrix: anything that makes them disagree is a bug in the sparse assembly, not a trade.

---

# Third commit: re-admit filtered observations once the poses improve

COLMAP's `CompleteTracks` — the other half of the filter this crate already has.

Filtering judges a sighting against the geometry of the moment, and the first filter fires three cameras in. So a sighting rejected against an early, raw PnP pose was gone for the rest of the run, **even after the solve that would have vindicated it**. Every global BA judges every sighting again and the decision only ever goes one way, so track length erodes monotonically over a long capture.

That is the mechanism behind a severed map: `cut_min` — tracks crossing the weakest keyframe cut — counts exactly the long tracks that lose a sighting first. Measured on a 459-keyframe walkthrough, **zero** tracks crossed the weakest cut, leaving the two halves of the map joined by nothing and their relative scale and rotation unconstrained, while every residual-based metric read healthy (0.82 px).

This also makes the filter's deletions **recoverable**, which is how COLMAP has always worked and how this file did not: `DeleteObservation` cuts the `Point3D`↔`Point2D` link while the `Point2D` survives in the correspondence graph, precisely so completion can re-admit it. `reconstruct` now keeps that graph — one clone of the observations at entry — instead of deleting evidence outright.

Re-admission **appends**, so each touched track is re-sorted into ascending camera order, permuting `norm` and `norm_depth` together since they're index-paired. `tracks.rs` publishes that order as contract and `Reconstruction::observations` is emitted in it, so the test asserts the ordering rather than trusting the append.

`ReconstructionConfig::complete_tracks`, default `true`; off restores the previous delete-only behaviour at the cost of the retained copy.

---

All three commits are the same subsystem — the incremental SfM solve — and share `sfm.rs`, so they review and land together.


edgarriba:
**Note on the two track-repair commits, before review.**

`complete_tracks` and `merge_tracks` are ports of COLMAP's `CompleteTracks` and `MergeTracks`. On COLMAP they repair a real incompleteness: COLMAP builds tracks incrementally at `max_transitivity = 1`, so a track can genuinely be missing observations that a later pose makes admissible, and two tracks can genuinely be the same landmark reached by different match chains.

**`kornia-calib`'s `build_tracks` does not create that incompleteness.** It emits the *full transitive closure* of the verified match graph (`tracks.rs`), so:

- `complete_tracks` can only re-admit observations the reprojection filter itself removed — which is a real and useful case, and is what its test covers. It is *not* recovering the transitivity COLMAP's version recovers, because none is missing.
- `merge_tracks`' candidate set is close to empty by construction: two points reached by a verified correspondence are already one component, so they are already one track. Measured on a 643-keyframe walkthrough, enabling it moved the point count the *wrong* way (111,919 → 113,112) and left `obs_per_point` flat at 4.65–4.68 — consistent with it finding almost nothing to fuse.

Both are correct code with tests, and `complete_tracks` earns its place by making the filter's deletions recoverable. But nobody should read this PR as delivering COLMAP's track-repair benefits on a closure-based track builder, and nobody should port the third piece (`Retriangulate`) expecting a different outcome without first changing how tracks are built.

The interesting question that falls out — worth its own issue rather than this PR — is whether the closure is the better choice at all. It cannot lose transitivity, but a single bad match chains two physically distinct points into one track, and `filter_points` then deletes the whole thing rather than splitting it. COLMAP's incremental construction plus `LORANSAC` triangulation with a recursive split spawns a second landmark where we delete the contradiction.

PR #1110, opened 2026-08-14
feat(calib): factorise the reduced camera system sparsely by default in SfM

edgarriba:
The sparse reduced camera system landed in #1105 and **`reconstruct` has never used it.** `sfm.rs` builds its `BaParams` with `..Default::default()`, and that default is `false` — correct for a rig, and exactly wrong for the long sequential capture the sparse path was written for.

The reduced system is `6P × 6P`, and two cameras occupy a nonzero 6×6 block only if they share a point. On a walkthrough, cameras a few hundred frames apart share nothing, so the dense path pays `O(dim²)` memory and `O(dim³)` factorisation for structure that isn't there. At 900 keyframes that's a dense **5400×5400** matrix, ~233 MB, ~5×10¹⁰ flops per iteration.

## Measured

Sequential synthetic, 10 LM iterations, **identical cost to 5 significant figures and identical iteration counts**:

| views | dense | sparse | |
|---|---|---|---|
| 100 | 0.11 s | 0.02 s | 5.5× |
| 300 | 1.20 s | 0.08 s | 15× |
| 600 | 8.34 s | **0.13 s** | **64×** |

The gap widens with camera count — dense grows roughly cubically, sparse roughly linearly. Sparse was **never slower at any size measured, down to 8 views**.

## Exposed, not hardcoded

`ReconstructionConfig::sparse_reduced_system`, defaulting to `true`, documented with the measurement above and with the one case for turning it off: a genuine rig, where every camera sees the same subject, the block pattern really is dense, and the sparse bookkeeping saves nothing.

`CalibConfig`'s rig path is untouched, and so is `BaParams::default()` — this changes what `reconstruct` asks for, not what the solver defaults to.

## Tests

`sparse_reduced_system_does_not_change_the_reconstruction` pins the end-to-end result a caller receives: same points to 1e-9, same rmse to 1e-6, same registration count. It asserts *equality*, not comparable quality, because this is a different storage of the same matrix — anything that makes them disagree is a bug in the sparse assembly, not a tuning trade.

`kornia-3d` already pins the lower triangle as bit-identical at the matrix level. The other 40 tests in this crate previously exercised the dense path and are unchanged, which is itself the regression evidence.

### DIFF
--- a/crates/kornia-calib/src/sfm.rs
+++ b/crates/kornia-calib/src/sfm.rs
@@ -1104,6 +1104,7 @@ fn global_ba(
             } else {
                 0.0
             },
+            sparse_reduced_system: config.sparse_reduced_system,
             ..Default::default()
         },
         up_priors(poses, a0, config).as_deref(),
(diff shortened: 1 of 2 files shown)


edgarriba:
Folded into #1109 so the two SfM changes review and land as one. Same commit, unchanged; closing this in favour of that.

PR #1112, opened 2026-08-17
feat(3d): intrinsics in bundle adjustment — radial term, then free focal and k1 as a bordered block

edgarriba:
Step one of making intrinsics a bundle-adjustment parameter block.

The residual is a bare pinhole — `u = fx*(x/z) + cx` — so a focal or distortion error can only be absorbed by **bending the structure**, and no reprojection residual reveals it. The post-hoc correction that exists downstream cannot recover it either: it fits the residuals of a reconstruction already built with the wrong camera.

This adds the distortion model and the intrinsic Jacobian columns **without freeing anything**. `k1` is fixed at zero, the reduced camera system is untouched, `BaObservation` is unchanged. Behaviour is identical; the value is that the next step becomes safe to attempt.

## Bit-identity, established rather than assumed

At `k1 = 0` the new code must reduce exactly to the old formulas.

- ~95M randomised f32 comparisons of the projection coefficients and residual → zero differences.
- **384 end-to-end solver configurations** fingerprinted bit-for-bit against `f999fe3d`: 4 seeds × 4 geometry regimes (including `MIN_Z`, behind-camera, 1e16 magnitudes) × depth × Identity/Huber/Cauchy × log-depth × pose priors. **All 384 identical**, including the 60 identical `CholeskyFailed` errors.

Two documented exceptions, both narrow: a signed-zero case that is value-identical and cannot change any accumulation, and an `r²`-overflow NaN confined to a region where the old objective was already `+inf`.

## Why `k1` is a parameter and not a `const`

This is the substance of the commit, not a refactor.

While `BA_K1` was a `const`, **no test could evaluate the residual at nonzero `k1`** — so the composed chain rule, the part the derivation is easiest to get wrong in, was unreachable. Measured: **seven deliberately wrong derivatives passed the entire 207-test suite**:

| mutant | caught before? |
|---|---|
| `u` missing its `* rad.d` | no |
| `a0` built from `dyd_dyn` | no |
| `b0` scaled by `fx` | no |
| `a2` cross-term sign flipped | no |
| `b2` cross using `pc.y` | no |
| `du/dfx = xn` instead of `xd` | no |
| `dv/dk1` scaled by `fx` | no |

Only mutations *inside* `radial_term` were caught, because only those changed the `k1 = 0` value.

`k1` is now a parameter of `residual_and_jacobians_k1`, with a thin wrapper supplying `BA_K1` to every existing call site — no call site changes.

## The test

`every_jacobian_matches_central_differences_at_nonzero_k1` checks every point and intrinsic Jacobian against central differences at `k1 = -0.3`, `r² ≈ 1`, under a general rotation so the cross terms cannot vanish.

The regime is deliberate: at a small coefficient near the optical axis the distortion terms sit inside the finite-difference tolerance and two of those seven mutants slip back through. Confirmed by injecting two of them — `du/dfx` analytic 0.851 vs difference 0.616; `dv/dk1` analytic −245.5 vs −241.4. Both fail now, both passed before.

## Note for step 2

Linearisation and the trial-cost pass share one `camera: &PinholeCamera`. The moment `fx`/`k1` are free, **the trial pass must receive the trial intrinsics**, or `new_cost < cost` scores the old model against a new step — the same failure this file already warns about for `depth_residual`.

Also: a shared intrinsic block couples every camera to every other, so the reduced system gains a dense border. That must be a bordered/arrowhead solve, not a densified RCS, or the sparse path from #1105 (measured 64× at 600 views) is thrown away.

208 tests pass, clippy clean, no new dependencies.

### DIFF
--- a/crates/kornia-3d/src/ba.rs
+++ b/crates/kornia-3d/src/ba.rs
@@ -335,6 +335,46 @@ pub struct BaParams {
     ///
     /// Only read by [`crate::ba_schur`]; ignored by [`bundle_adjust`].
     pub sparse_reduced_system: bool,
+    /// Refine the focal length as a SHARED parameter across every camera.
+    ///
+    /// The free variable is a dimensionless scale `α` on the seed focals, not an absolute `fx`:
+    /// `fx = α·fx₀`, `fy = α·fy₀`, seeded at `α = 1`. That fixes the aspect ratio (COLMAP's
+    /// `SIMPLE_RADIAL` carries one focal for both axes) and stays meaningful for callers that
+    /// bundle-adjust in NORMALISED coordinates, where `fx₀ = 1` and an absolute focal is not a
+    /// focal. `cx`/`cy` stay fixed — the principal point is the least observable intrinsic, and
+    /// `SIMPLE_RADIAL` fixes it too.
+    ///
+    /// SHARED means exactly that: [`BaObservation`] carries no camera index, so every observation
+    /// is assumed to come from the one physical camera named by the solver's `camera` argument.
+    ///
+    /// `α` is clamped to `[0.1, 10]` — COLMAP's `min_focal_length_ratio` / `max_focal_length_ratio`
+    /// band, read as a ratio to the SEED, since no image dimensions are available here and the
+    /// band therefore cannot be expressed against the sensor.
+    ///
+    /// Use this WITH A GAUGE. With every pose free and no depth or pose prior, focal trades
+    /// against map scale — the classic focal/depth ambiguity — so the fit is only as determined as
+    /// the fixed poses, pose priors or depth measurements make it. On a degenerate segment (pure
+    /// rotation, a single plane) the intrinsic step falls back to zero rather than failing.
+    ///
+    /// Default `false`, and with both this and [`Self::refine_k1`] false the solve is bit-identical
+    /// to one run before either parameter existed.
+    ///
+    /// Only read by [`crate::ba_schur`]; ignored by [`bundle_adjust`].
+    pub refine_focal: bool,
+    /// Refine the radial distortion coefficient `k1` as a SHARED parameter across every camera.
+    ///
+    /// Seeded from `camera.k1` when this is set — and ONLY when it is set. With this `false` the
+    /// solver pins `k1 = 0` as it always has, whatever the camera carries; that is what keeps
+    /// existing callers bit-identical.
+    ///
+    /// `k1` is clamped so `1 + k1·r²` stays positive over the largest `r²` seen in the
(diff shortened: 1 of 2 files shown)


edgarriba:
**Step 2 pushed** (`7cc311de`) — this PR now contains both halves.

`BaParams::refine_focal` and `refine_k1`, **both default false**. Off, the solver is bit-identical: 162 end-to-end configurations, 35,910 bit patterns, digest `0xac49186c81ee3861` on this tree and on the branch base.

**The border rides the existing factorisation.** A shared intrinsic block couples every camera to every other, so densifying the reduced camera system to absorb it would have discarded #1105's sparse path. Instead the intrinsic elimination reuses the same `C⁻¹` and `BC⁻¹` the pose path already computes, and the pose solve widens its RHS to `dim × (1+q)` so one factorisation yields both `M⁻¹m` and `M⁻¹V`. Measured at 600 views / 30k observations: **sparse 0.218 s → 0.243 s**, against 2.24 s dense.

**Accept/reject.** Linearisation and the trial pass previously shared one `&PinholeCamera` — with intrinsics free that is a live bug, since the trial must be scored at the *trial* intrinsics or LM accepts garbage. `BaIntrinsics` is a value type, so a stale evaluation is a type error. Two mutation studies: 19/20 and 13/13 deliberate defects caught.

## What it recovers

| | |
|---|---|
| synthetic, seeded 0.863× truth | fx exact clean, **+0.17% ± 0.55%** at 0.5 px noise |
| k1 = 0.036, seeded 0 | recovers **0.03600** |
| 5× 7-Scenes seeded −14.53% | recovers **62%**; the incumbent linear fit recovers **0.00%** |

## Where it does not

Two exact motion degeneracies, proven flat valleys rather than local minima: **pure axial translation (−7.1%)** and **lateral translation under constant rotation (−7.3%)**. Neither a fixed gauge nor depth measurements rescue them. **Planar scenes are not a failure mode** — recovered exactly at 0.00 m relief.

## Two things a reviewer should push back on

1. **`refine_focal: true` with `refine_k1: false` silently discards a caller's calibrated `k1`** and lets the fitted focal absorb it. Measured on a real 459-keyframe clip seeded 15% low: focal-only sheds **21%** of the cloud, focal+k1 sheds **2.5%**, and interior min-cut goes 13 → 18 → **50** across off / focal / focal+k1. Free them together or not at all. Not fixed in this PR; the combination should probably be rejected outright.
2. **The `[0.1, 10]` focal-ratio guard is not a safety net.** It clamps silently with `converged = true`, and being per-solve it chains through a re-seeding incremental loop.

**Do not enable `refine_focal` on calibrated cameras** — on 7-Scenes, where the focal is exactly 585, freeing it wanders −7.14%…+5.33% and costs mean ATE +4.5%.

PR #1121, opened 2026-08-20
perf(imgproc): register-tile the SIFT vertical blur and warp-aggregate descriptor atomics

Incharajayaram:
## 📝 Description

**Fixes/Relates to:** 

Two independent optimizations to the CUDA SIFT path, both bit-exact against the
existing reference tests.

**1. Register-tiled vertical Gaussian blur** (`src/cuda/sift/kernels.rs`, `src/cuda/sift/pyramid.rs`)

`blur_v_tiled_src(kernel, q, dog)` emits a kernel computing `q` output rows per
thread (default 2, `KORNIA_SIFT_TILE_Q` to sweep). Adjacent output rows share
most of their column taps, so the `n + q - 1` loads are hoisted into registers
instead of each thread re-issuing its own `__ldg` per tap. The border test also
moves to warp granularity.

This replaces `blur_v_dog_src`, which produced the fused DoG variant by string
replacement over `blur_v_src`'s generated source. That was a documented footgun
— its own comment warned that a signature edit in one and not the other
silently emits an unrenamed kernel. Fused DoG is now just a parameter of the
same generator.

**2. Warp-aggregated descriptor histogram atomics** (`src/cuda/sift/descriptor.rs`)

The trilinear splat has lanes of a warp repeatedly hitting the same histogram
bin and serializing on the atomic unit. `warp_atomicAdd` combines the lanes
sharing an address via shuffles so the leader issues one `atomicAdd` instead of
up to 32.

Bit equality holds because the bins are fixed point (#1092) and unsigned
integer addition is associative — the reordered sum is identical to what the
per-lane atomics produce in any order. The same reordering on floats would not
be reproducible. Guarded on `__CUDA_ARCH__ >= 700` (`__match_any_sync` is
Volta+) with the plain per-lane atomic on older devices, so the crate's arch
floor is unchanged.

Reference implementation: OpenCV's SIFT, `modules/features2d/src/sift.dispatch.cpp`
and `sift.simd.hpp` — the same source this module already documents itself
against in `src/cuda/sift/mod.rs`. Neither change touches the expression trees
that bit equality depends on: the tiling only changes *which thread* evaluates a
tap, and the atomic aggregation only changes the *order* of an integer sum.

---

## 🛠️ Changes Made
- [x] `kernels.rs`: `blur_v_tiled_src(kernel, q, dog)` replaces `blur_v_src` + `blur_v_dog_src`
- [x] `pyramid.rs`: `TILE_Q` (default 2) + `KORNIA_SIFT_TILE_Q` override; launchers use the tiled grid height
- [x] `descriptor.rs`: `warp_atomicAdd` for the 8 histogram accumulations, sm_70 guarded
- [x] `benches/bench_cuda_sift.rs`: end-to-end `detect_and_compute`, unified memory vs explicit copies
- [x] `benches/bench_opencv_sift.py`: `cv2.SIFT_create` vs kornia-rs host and CUDA bindings
- [x] `benchmarks.md`: results for both, GTX 1650 and Jetson Orin Nano

---

## 🧪 How Was This Tested?
- [x] **Unit Tests:** existing SIFT suite, all 43 pass with `--features cuda` on a GTX 1650 (sm_75). The gates that matter here are `fused_blur_v_dog_matches_unfused_bitwise`, `dog_pyramid_matches_reference_bitwise`, `full_gaussian_pyramid_matches_reference_bitwise`, `separable_blur_matches_reference_bitwise`, `end_to_end_matches_reference_keypoints` and `pipeline_is_deterministic_across_runs` — all bitwise/determinism checks, so neither optimization changes output. `blur_v_dog_generates_fused_kernel` updated for the new generator.

```
$ cargo test -p kornia-imgproc --features cuda --lib cuda::sift
test cuda::sift::pyramid::tests::fused_blur_v_dog_matches_unfused_bitwise ... ok
test cuda::sift::pyramid::tests::dog_pyramid_matches_reference_bitwise ... ok
test cuda::sift::pyramid::tests::full_gaussian_pyramid_matches_reference_bitwise ... ok
test cuda::sift::pyramid::tests::separable_blur_matches_reference_bitwise ... ok
test cuda::sift::plan::tests::end_to_end_matches_reference_keypoints ... ok
test cuda::sift::plan::tests::pipeline_is_deterministic_across_runs ... ok

test result: ok. 43 passed; 0 failed; 0 ignored; 0 measured; 449 filtered out; finished in 5.13s
```

- [x] **Manual Verification:** `cargo fmt --all -- --check` and `cargo clippy -p kornia-imgproc --no-deps --all-targets --features cuda -- -D warnings` both clean. CPU-only `cargo check -p kornia-imgproc` (no `cuda` feature) unaffected.
- [x] **Performance/Edge Cases:** border rows still take the `refl101` reflected-index path; the fast path is only entered when the whole `q`-row group is interior, so partial groups at the bottom edge fall to the guarded path. `q` is validated to 1..=16. On pre-Volta devices `warp_atomicAdd` compiles to the plain atomic.

`bench_cuda_sift` on a GTX 1650 (30 iters, 10 warmup), per frame end-to-end:

| Resolution | Explicit copies | Unified memory | Unified speedup |
| --- | ---: | ---: | ---: |
| VGA 640×480 | 2.451 ms | 5.114 ms | 0.48x |
| HD 1280×720 | 8.086 ms | 13.715 ms | 0.59x |
| FHD 1920×1080 | 19.902 ms | 25.013 ms | 0.80x |

Unified memory loses on a discrete GPU (pageable buffer, per-page PCIe faults);
it wins only on Jetson at VGA (1.13x) and OOMs above that. Recorded in
`benchmarks.md` as a negative result — the unified path is not yet a general win.

---

## 🕵️ AI Usage Disclosure
*Check one of the following:*
- [ ] 🟢 **No AI used.**
- [ ] 🔴 **AI-generated:**

---

## 🚦 Checklist

---

## 💭 Additional Context

The two changes are independent and can be split if you'd rather review them
separately. Both are pure kernel-generation changes behind the `cuda` feature;
no public API moves.

---

## 🧹 CI fix carried in this branch

`Check - CUDA features` is the only job in `rust_lint.yml` that resolves its own
toolchain (`dtolnay/rust-toolchain@stable`); every other job runs through pixi,
which pins rust to 1.96. When stable moved to 1.98 the new
`clippy::chunks_exact_to_as_chunks` lint began firing on 43 pre-existing call
sites across `kornia-io` and `kornia-imgproc` and, with `-D warnings`, failed
that job on every open PR with no source change.

This branch pins the job to 1.96.0 so it runs the same compiler as the `Clippy`
job. One line, and none of the 43 flagged call sites needed touching — verified
by running the job's three commands under pixi's toolchain against an otherwise
unmodified tree:

```sh
pixi run cargo clippy --tests -p kornia-tensor  --features cuda -- -D warnings
pixi run cargo clippy --tests -p kornia-image   --features cuda -- -D warnings
pixi run cargo clippy --tests -p kornia-imgproc --features cuda -- -D warnings
```

all three clean. The pin should be bumped deliberately alongside pixi's rust
pin, so a compiler release cannot turn the queue red on its own again.

### DIFF
--- a/.github/workflows/rust_lint.yml
+++ b/.github/workflows/rust_lint.yml
@@ -69,7 +69,12 @@ jobs:
     steps:
       - uses: actions/checkout@v6
 
-      - uses: dtolnay/rust-toolchain@stable
+      # Pinned to match the toolchain pixi provides for the `Clippy` job above
+      # (`rust = ">=1.76"` currently resolves to 1.96). Tracking `stable` meant a
+      # compiler release could introduce a lint and fail this job on every open
+      # PR without any source change - Rust 1.98's `chunks_exact_to_as_chunks`
+      # did exactly that. Bump this deliberately, together with pixi's rust pin.
+      - uses: dtolnay/rust-toolchain@1.96.0
         with:
           components: clippy
 
--- a/crates/kornia-imgproc/Cargo.toml
+++ b/crates/kornia-imgproc/Cargo.toml
@@ -46,6 +46,11 @@ required-features = ["cuda"]
 
 [[bench]]
 harness = false
+name = "bench_cuda_sift"
+required-features = ["cuda"]
+
+[[bench]]
+harness = false
 name = "bench_sift_instr"
 
 [[bench]]
--- a/crates/kornia-imgproc/benches/bench_cuda_sift.rs
+++ b/crates/kornia-imgproc/benches/bench_cuda_sift.rs
@@ -1,0 +1,137 @@
+use std::time::Instant;
+
+use argh::FromArgs;
+use cudarc::driver::CudaContext;
+use kornia_image::{Image, ImageSize};
+use kornia_imgproc::cuda::sift::{FirstOctave, SiftCuda, SiftCudaConfig};
+
+/// Benchmark unified memory against explicit device copies for SIFT.
+#[derive(FromArgs)]
+struct Args {
+    /// CUDA device ordinal (default 0)
+    #[argh(option, default = "0")]
+    device: usize,
+
+    /// timed iterations per case (default 10)
+    #[argh(option, default = "10")]
+    iters: usize,
+
+    /// untimed warmup iterations per case (default 3)
+    #[argh(option, default = "3")]
+    warmup: usize,
+
+    /// accepted and ignored: `cargo bench` passes `--bench` to the harness
+    #[argh(switch)]
+    #[allow(dead_code)]
+    bench: bool,
+}
+
+const SIZES: &[(&str, usize, usize)] = &[
+    ("VGA    640x480", 640, 480),
+    ("HD    1280x720", 1280, 720),
+    ("FHD  1920x1080", 1920, 1080),
+];
+
+fn timed<F>(warmup: usize, iters: usize, mut f: F) -> Result<f64, Box<dyn std::error::Error>>
+where
+    F: FnMut() -> Result<(), Box<dyn std::error::Error>>,
+{
+    for _ in 0..warmup {
+        f()?;
+    }
+    let start = Instant::now();
+    for _ in 0..iters {
+        f()?;
+    }
+    Ok(start.elapsed().as_secs_f64() * 1e3 / iters as f64)
+}
+
+fn fill_ramp(buf: &mut [f32]) {
+    let inv = 255.0 / buf.len() as f32;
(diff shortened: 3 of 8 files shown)


edgarriba (approved):
Fix ci

Incharajayaram:
yes fixed the ci

PR #1032, opened 2026-07-19
test(cuda): OpenCV pixel-level correctness check for resize and warp-affine

Incharajayaram:
## 📝 Description

**Fixes/Relates to:** 

Adds Python + Rust tooling to compare GPU kernels against OpenCV pixel-by-pixel. Covers all four interpolation modes for both resize and warp-affine.

---

## 🛠️ Changes Made

- **`examples/dump_cuda_resize.rs`** — Rust binary that runs a GPU resize for a given mode/dimensions and prints pixel data as JSON. Updated to pass `PixelMapping::HalfPixel` after the API change that added the `PixelMapping` enum parameter to all resize launchers.
- **`examples/dump_cuda_warp_affine.rs`** — Rust binary that runs a GPU warp-affine (rotation at a given angle) and prints pixels + the 2×3 forward matrix as JSON.
- **`examples/check_correctness_cuda.py`** — Orchestrating Python script: calls both Rust binaries, re-creates the same image in NumPy, runs the equivalent cv2 operation, and compares pixel-by-pixel.
- **`Cargo.toml`** — Registers both new examples with `required-features = ["cuda"]`.

   ### VPI correctness + benchmark (runs on Jetson / JetPack-equipped x86)

   - **`examples/check_correctness_vpi.py`** â Pixel-by-pixel comparison against VPI CUDA backend for resize, warp_affine, and warp_perspective. Exits cleanly with a JetPack install hint if VPI is not available.
   - **`examples/bench_vpi.py`** â Wall-clock throughput benchmark (VPI CUDA backend) at 1920x1080 and 3840x2160 for all three ops. Reports mean, stddev, and Mpix/s. VPI ships with JetPack SDK only â numbers to be collected on Jetson hardware.

### Result classification

| Case | Verdict | Reason |
|---|---|---|
| resize bilinear (down + up) | ✅ PASS ≤6e-8 | Identical half-pixel alignment and BORDER_REPLICATE |
| resize nearest | ⚠ INFO | Convention difference: our `floor((x+0.5)*s)` vs cv2's `floor(x*s)` |
| resize bicubic | ⚠ INFO | OOB 4×4 taps: BORDER_REPLICATE vs cv2's BORDER_REFLECT_101 |
| resize lanczos | ⚠ INFO | Different kernel: Lanczos-3 (6-tap) vs cv2's Lanczos-4 (8-tap) |
| warp bilinear identity | ✅ PASS ≤8e-9 | No fractional-weight edge tap |
[...]

PR #1122, opened 2026-08-20
feat(tensor): element-wise and reduction ops with CPU/GPU dispatch on MemoryDomain

Incharajayaram:
## 📝 Description

**Fixes/Relates to:** _TODO — issue link pending, will add before review._

Supersedes #1036, which was auto-closed by the stale bot on 2026-08-14 before
anyone reviewed it (the Copilot pass failed on a quota limit, so there is no
prior feedback to carry over).

Adds `kornia_tensor::ops` — unary, binary and reduction operations over
`Tensor<f32, N>` that choose their backend from the tensor's `MemoryDomain` at
runtime. This is the same shape as the routing `kornia-imgproc` already uses in
`cuda::dispatch::pair_residency`: the public entry point is backend-agnostic,
host operands take the CPU loops, device operands launch a kernel, and nothing
transfers implicitly.

- `UnaryOp` — `Abs`, `Relu`, `Neg`, `Clamp { min, max }`
- `BinaryOp` — `Add`, `Sub`, `Mul`, `Div`, `Min`, `Max`
- `ReduceOp` — `Sum`, `Mean`

`Host` and `Unified` both take the CPU path — unified memory is host-accessible
(`MemoryDomain::is_host_accessible`), so the CPU loops stay valid for it.
`Device` launches NVRTC kernels compiled once and cached in a `OnceLock`.
Without the `cuda` feature a device tensor returns `OpsError::CudaNotEnabled`
rather than failing to compile, so CPU-only builds are unaffected.

Available as free functions (`apply_unary`, `apply_binary`, `reduce`) and as
inherent methods on `Tensor<f32, N>`.

**What changed since #1036.** That branch was 78 commits behind main and carried
a lot that is now obsolete: `PLAN_remaining.md` (a planning doc that does not
belong in the repo), four example scripts already merged in #1032, and a
`UnifiedResource` / `CudaUnifiedAllocator` implementation superseded by #1034.
All dropped. This branch is 2 files and +770 lines against current main, versus
11 files and +2490 before.

Operation semantics follow PyTorch's elementwise/reduction ops (`torch.abs`,
`torch.relu`, `torch.clamp`, `torch.sum`, `torch.mean`); the parity tests below
pin the GPU kernels against the CPU loops rather than against an external
reference, since both sides are implemented here.

---

## 🛠️ Changes Made
- [x] `crates/kornia-tensor/src/ops.rs` (new): op enums, `OpsError`, CPU implementations, NVRTC GPU kernels, domain dispatch, tests
- [x] `crates/kornia-tensor/src/lib.rs`: `pub mod ops` + re-exports of the op enums and free functions
- [x] `crates/kornia-tensor/benches/bench_tensor_ops.rs` (new): CPU vs GPU benchmark with a kernel / round-trip split, plus a summation-accuracy probe

---

## 🧪 How Was This Tested?
- [x] **Unit Tests:** 9 CPU tests covering every op and the shape-mismatch error, plus 3 CPU/GPU parity tests gated behind `cuda` that run each op on both backends and compare.

```
$ cargo test -p kornia-tensor --features cuda --lib ops
running 14 tests
test ops::tests::cpu_unary_abs ... ok
test ops::tests::cpu_unary_relu ... ok
test ops::tests::cpu_unary_neg ... ok
test ops::tests::cpu_unary_clamp ... ok
test ops::tests::cpu_binary_add ... ok
test ops::tests::cpu_binary_mul ... ok
test ops::tests::cpu_reduce_sum ... ok
test ops::tests::cpu_reduce_mean ... ok
test ops::tests::cpu_shape_mismatch_is_error ... ok
test ops::tests::cuda_parity::parity_unary_all_ops ... ok
test ops::tests::cuda_parity::parity_binary_all_ops ... ok
test ops::tests::cuda_parity::parity_reduce_sum_and_mean ... ok

test result: ok. 14 passed; 0 failed; 0 ignored; 0 measured; 103 filtered out; finished in 0.98s
```

Run on a GTX 1650 (sm_75), CUDA 12.4.

- [x] **Manual Verification:** `cargo check -p kornia-tensor` (no `cuda` feature) builds clean, confirming CPU-only consumers are unaffected. `cargo fmt --all -- --check` and `cargo clippy -p kornia-tensor --no-deps --all-targets --features cuda -- -D warnings` both clean.
- [x] **Performance/Edge Cases:** shape mismatch between operands is a typed `OpsError::ShapeMismatch` before any launch. A device tensor in a CPU-only build returns `OpsError::CudaNotEnabled` instead of panicking.

**Benchmark** — GTX 1650 (sm_75), minimum of 20 iterations after 3 warmups:

```sh
cargo bench -p kornia-tensor --bench bench_tensor_ops --features cuda
```

| op | size | CPU ms | GPU kernel ms | GPU round-trip ms | kernel | round-trip |
| --- | ---: | ---: | ---: | ---: | ---: | ---: |
| unary relu | 1M | 2.711 | 0.062 | 10.868 | 43.8x | 0.2x |
| binary add | 1M | 4.465 | 0.078 | 16.076 | 57.1x | 0.3x |
| binary mul | 1M | 4.272 | 0.079 | 16.871 | 54.4x | 0.3x |
| reduce sum | 1M | 1.794 | 0.111 | 3.869 | 16.2x | 0.5x |
| unary relu | 10M | 29.583 | 0.477 | 148.118 | 62.0x | 0.2x |
| binary add | 10M | 53.523 | 0.695 | 183.818 | 77.0x | 0.3x |
| binary mul | 10M | 30.693 | 0.693 | 170.349 | 44.3x | 0.2x |
| reduce sum | 10M | 19.922 | 0.938 | 38.435 | 21.2x | 0.5x |
| unary relu | 100M | 253.036 | 4.698 | 1418.901 | 53.9x | 0.2x |
| binary add | 100M | 401.012 | 6.893 | 1790.781 | 58.2x | 0.2x |
| binary mul | 100M | 467.344 | 6.850 | 1798.547 | 68.2x | 0.3x |
| reduce sum | 100M | 220.927 | 9.314 | 352.570 | 23.7x | 0.6x |

Kernel-only is 16-78x across every op and size. The round trip is a **loss**
everywhere (0.2-0.6x) — the same conclusion the imgproc sweep reached: for
memory-bandwidth-bound elementwise work PCIe dominates, and the device path
only pays off when the tensor is already resident and stays resident across
several ops. That is exactly the usage this dispatch enables, but the numbers
should stop anyone expecting a win from a one-shot upload-compute-download.

Two caveats stated plainly. The CPU baseline is the scalar single-threaded loop
in `ops.rs` — `rayon` is already a dependency of this crate and is not used
here, so a parallel host path would narrow these ratios considerably. And the
host timings are noisy on a loaded machine (2-6x spread run to run), which is
why the harness reports the minimum per-iteration time rather than the mean;
GPU kernel times were stable to a few percent.

---

## 🕵️ AI Usage Disclosure
*Check one of the following:*
- [ ] 🟢 **No AI used.**
- [ ] 🔴 **AI-generated:**

---

## 🚦 Checklist

---

## 💭 Additional Context

**Summation accuracy — flagging this for a maintainer's call.** The benchmark
includes a probe that sums 10M elements interleaving `1e7` and `1e-3`. The two
backends disagree by 2.8%:

| | Sum | Relative error |
| --- | ---: | ---: |
| True value (f64) | 50000000004999.8 | — |
| Host path (sequential f32) | 48614172786688 | 2.77% |
| GPU path (block tree + atomicAdd) | 49982858067968 | 0.034% |

The host figure reproduces an explicit sequential f32 accumulation bit for bit
(cross-checked against numpy), so this is textbook naive summation rather than a
bug — but it does mean `reduce` returns materially different answers depending
on where the tensor lives, and the *host* path is the inaccurate one. The GPU's
tree reduction is roughly 80x more accurate here.

The existing parity tests use benign input and a tolerance, so they pass. If you
would like the host path to use pairwise or Kahan summation to close the gap, I
am happy to add it here or in a follow-up — it is a change to CPU behaviour, so
I did not make that call unilaterally.

Placement note: this lives in `kornia-tensor` rather than `kornia-tensor-ops`
because the GPU path needs `CudaKernel` / `CudaLaunchBuilder` / `CudaError` from
`kornia-tensor`'s own `cuda` module, and dispatch keys off `MemoryDomain`, also
`kornia-tensor`'s type. Hosting it in `kornia-tensor-ops` would mean re-exporting
the NVRTC internals across a crate boundary and adding a `cuda` feature there.
Happy to move it if you would rather elementwise/reduce sit next to the existing
`TensorOps` algorithms.

---

## 🧹 CI fix carried in this branch

`Check - CUDA features` is the only job in `rust_lint.yml` that resolves its own
toolchain (`dtolnay/rust-toolchain@stable`); every other job runs through pixi,
which pins rust to 1.96. When stable moved to 1.98 the new
`clippy::chunks_exact_to_as_chunks` lint began firing on 43 pre-existing call
sites across `kornia-io` and `kornia-imgproc` and, with `-D warnings`, failed
that job on every open PR with no source change.

This branch pins the job to 1.96.0 so it runs the same compiler as the `Clippy`
job. One line, and none of the 43 flagged call sites needed touching — verified
by running the job's three commands under pixi's toolchain against an otherwise
unmodified tree:

```sh
pixi run cargo clippy --tests -p kornia-tensor  --features cuda -- -D warnings
pixi run cargo clippy --tests -p kornia-image   --features cuda -- -D warnings
pixi run cargo clippy --tests -p kornia-imgproc --features cuda -- -D warnings
```

all three clean. The pin should be bumped deliberately alongside pixi's rust
pin, so a compiler release cannot turn the queue red on its own again.

---

## ✅ Review follow-up

> We should also probably support in place ops `t.apply_unary_(Abs)`? — @edgarriba

Added in `a0260df1`: `apply_unary_inplace` / `apply_binary_inplace` free
functions plus `Tensor::apply_unary_` / `Tensor::apply_binary_` methods,
following PyTorch's trailing-underscore convention. They dispatch on
`MemoryDomain` exactly as the out-of-place forms do and skip the destination
allocation, which matters most on the device path where the out-of-place form
needs a second buffer of the same size.

The CUDA path reuses the existing kernels with one buffer bound as both source
and destination. That aliasing is safe because the mapping is strictly
element-wise — thread `i` reads index `i` and writes index `i`, so no thread
observes another thread's write.

Tests: `cpu_unary_inplace_matches_out_of_place` and
`cpu_binary_inplace_matches_out_of_place` check every op against the
out-of-place result, `cpu_binary_inplace_shape_mismatch_is_error` covers the
error path, and two CUDA parity tests confirm the aliased device launch agrees
with the CPU reference. Both methods carry doc examples that run as doctests.
`cargo test -p kornia-tensor --features cuda --lib ops` now reports 19 passed,
`--doc` reports 28.

### DIFF
--- a/crates/kornia-tensor/src/ops.rs
+++ b/crates/kornia-tensor/src/ops.rs
@@ -1,0 +1,762 @@
+//! Element-wise and reduction operations on [`Tensor<f32, N>`] with automatic
+//! CPU / GPU dispatch based on the tensor's [`MemoryDomain`].
+//!
+//! # Operations
+//!
+//! | Group | Ops |
+//! |-------|-----|
+//! | Unary | [`UnaryOp::Abs`], [`UnaryOp::Relu`], [`UnaryOp::Neg`], [`UnaryOp::Clamp`] |
+//! | Binary | [`BinaryOp::Add`], [`BinaryOp::Sub`], [`BinaryOp::Mul`], [`BinaryOp::Div`], [`BinaryOp::Min`], [`BinaryOp::Max`] |
+//! | Reduce | [`ReduceOp::Sum`], [`ReduceOp::Mean`] |
+//!
+//! # Dispatch
+//!
+//! [`apply_unary`], [`apply_binary`], and [`reduce`] inspect the tensor's
+//! [`MemoryDomain`] at runtime:
+//!
+//! - [`MemoryDomain::Host`] / [`MemoryDomain::Unified`] → CPU loops (unified
+//!   memory is host-accessible so the CPU path is always valid).
+//! - [`MemoryDomain::Device`] → NVRTC GPU kernels (requires `cuda` feature;
+//!   returns [`OpsError::CudaNotEnabled`] otherwise).
+
+use crate::{
+    resource::MemoryDomain,
+    tensor::{Tensor, TensorError},
+};
+
+// ── Op enums ──────────────────────────────────────────────────────────────────
+
+/// Element-wise unary operation.
+#[derive(Clone, Copy, Debug)]
+pub enum UnaryOp {
+    /// Absolute value: `|x|`
+    Abs,
+    /// Rectified linear unit: `max(0, x)`
+    Relu,
+    /// Negation: `-x`
+    Neg,
+    /// Clamp to `[min, max]`
+    Clamp {
+        /// Lower bound (inclusive).
+        min: f32,
+        /// Upper bound (inclusive).
+        max: f32,
+    },
+}
+
+/// Element-wise binary operation.
+#[derive(Clone, Copy, Debug)]
+pub enum BinaryOp {
+    /// `a + b`
+    Add,
+    /// `a - b`
+    Sub,
+    /// `a * b`
+    Mul,
+    /// `a / b`
+    Div,
+    /// `min(a, b)`
+    Min,
+    /// `max(a, b)`
+    Max,
+}
+
+/// Reduction operation (collapses all elements to a scalar).
+#[derive(Clone, Copy, Debug)]
+pub enum ReduceOp {
+    /// Sum of all elements.
+    Sum,
+    /// Arithmetic mean.
+    Mean,
+}
+
+// ── Error type ────────────────────────────────────────────────────────────────
+
+/// Error type for tensor element-wise and reduction operations.
+#[derive(Debug, thiserror::Error)]
+pub enum OpsError {
+    /// Shape or element-count mismatch between operands.
+    #[error("shape mismatch: {0}")]
+    ShapeMismatch(String),
+
+    /// Operation requires device-accessible memory.
+    #[error("tensor is not device-accessible (domain={0:?})")]
+    NotDeviceAccessible(MemoryDomain),
+
+    /// Tried to run a GPU op but the `cuda` feature is not enabled.
+    #[error("CUDA feature not enabled — rebuild with --features cuda")]
+    CudaNotEnabled,
+
+    /// Underlying CUDA driver error.
+    #[cfg(feature = "cuda")]
+    #[error("CUDA error: {0}")]
+    Cuda(#[from] crate::cuda::CudaError),
+
+    /// Wrapped tensor error.
+    #[error("{0}")]
+    Tensor(#[from] TensorError),
+}
+
+// ── CPU implementations ───────────────────────────────────────────────────────
+
+fn cpu_apply_unary<const N: usize>(
+    input: &Tensor<f32, N>,
+    output: &mut Tensor<f32, N>,
+    op: UnaryOp,
+) -> Result<(), OpsError> {
+    if input.shape != output.shape {
+        return Err(OpsError::ShapeMismatch(format!(
+            "{:?} vs {:?}",
+            input.shape, output.shape
+        )));
+    }
+    let src = input.as_slice();
+    let dst = output.as_slice_mut();
+    match op {
+        UnaryOp::Abs => {
+            for (d, s) in dst.iter_mut().zip(src) {
+                *d = s.abs();
+            }
+        }
+        UnaryOp::Relu => {
+            for (d, s) in dst.iter_mut().zip(src) {
+                *d = s.max(0.0);
+            }
+        }
+        UnaryOp::Neg => {
+            for (d, s) in dst.iter_mut().zip(src) {
+                *d = -*s;
+            }
+        }
+        UnaryOp::Clamp { min, max } => {
+            for (d, s) in dst.iter_mut().zip(src) {
+                *d = s.clamp(min, max);
+            }
+        }
+    }
+    Ok(())
+}
+
+fn cpu_apply_binary<const N: usize>(
+    a: &Tensor<f32, N>,
+    b: &Tensor<f32, N>,
+    out: &mut Tensor<f32, N>,
+    op: BinaryOp,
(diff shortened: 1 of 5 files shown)


crates/kornia-tensor/src/ops.rs
```diff
@@ -0,0 +1,762 @@
...
+    use super::*;
+    use crate::tensor::Tensor;
+
+    // ── CPU tests ─────────────────────────────────────────────────────────────
+
+    #[test]
+    fn cpu_unary_abs() {
+        let t = Tensor::<f32, 1>::from_shape_vec([4], vec![-1.0, 2.0, -3.0, 0.0]).unwrap();
+        let mut out = Tensor::<f32, 1>::zeros([4]);
+        t.apply_unary(&mut out, UnaryOp::Abs).unwrap();
```
edgarriba, line 786: We should also probably support in place ops t.apply_unary_(Abs)?

edgarriba (approved):
Fix ci

crates/kornia-tensor/src/ops.rs
Incharajayaram: yes i added, Tensor::apply_unary_ / apply_binary_ (plus apply_unary_inplace / apply_binary_inplace free functions), following the PyTorch trailing underscore convention.

Incharajayaram:
fixed the ci

edgarriba (approved):

PR #1111, opened 2026-08-16
Huber pnp refine

Astik-2002:
## 📝 Description
Fix for the issue [https://github.com/kornia/kornia-rs/issues/1099]
**⚠️ Issue Link Required**: This PR must be linked to an approved and assigned issue. See [Contributing Guide](CONTRIBUTING.md#pull-request) for details.

Added huber kernel in PnP pose refinement
---

## 🛠️ Changes Made
- [ ] Added robust loss in refine.rs
- [ ] Added testcases for testing huber loss

---

## 🧪 How Was This Tested?
- [✓] **Unit Tests:** (List new/updated tests)
- [✓] **Manual Verification:** (Describe the steps you took)
- [ ] **Performance/Edge Cases:** (How does this handle nulls, large data, etc.?)

---

## 🕵️ AI Usage Disclosure
*Check one of the following:*
- [ ] 🟢 **No AI used.**
- [✓] 🟡 **AI-assisted:** I used AI for boilerplate/refactoring but have manually reviewed and tested every line.
- [ ] 🔴 **AI-generated:** (Note: These PRs may be subject to stricter scrutiny or immediate closure if the logic is not explained).

---

## 🚦 Checklist
- [✓] I am assigned to the linked issue (required before PR submission)
- [ ] The linked issue has been approved by a maintainer
- [ ] This PR strictly implements what the linked issue describes (no scope creep)
- [ ] I have performed a **self-review** of my code (no "ghost" variables or hallucinations).
- [ ] My code follows the existing style guidelines of this project.
- [ ] I have commented my code, particularly in hard-to-understand areas.
- [ ] I have added tests that prove my fix is effective or that my feature works.
- [ ] (Optional) I have attached screenshots/recordings for UI changes.

---

## 💭 Additional Context
Add any other context or screenshots about the pull request here.

PR #1124, opened 2026-08-23
fix: Clean up #1111 scaffolding: TEMP test name/comment and lightweight_vio r

### LINKED ISSUE #1120: Clean up #1111 scaffolding: TEMP test name/comment and lightweight_vio reference in public rustdoc
## 📝 Description

#1111 merged with some development scaffolding still in place in `crates/kornia-3d/src/pnp/refine.rs`. Two items, both reaching a public surface.

**1. Temporary test left in the suite.** The test is a genuinely useful regression check — it demonstrates that Huber suppresses a 300px outlier (`no_robust_vs_clean_rmse=81.2` vs `huber_vs_clean_rmse=1.49`) — but it is named and commented as disposable:

```rust
// TEMP verification test (Claude): checks that `robust: Huber` actually
// suppresses a gross outlier's influence on the LM solve, vs. plain L2.
// Not part of the requested scaffold — delete or keep at your discretion.
#[test]
fn temp_huber_suppresses_outlier_influence() -> Result<(), PnPError> {
```

Suggested: keep the test, drop the comment, rename to something durable such as `huber_suppresses_gross_outlier_influence`, and consider replacing the bare `println!` with the assertion messages already present.

**2. Reference to an unrelated external codebase in published rustdoc.** The doc comment on `LMRefineParams::robust` reads:

```rust
/// M-estimator kernel applied per-residual (pixel-space, unlike
/// lightweight_vio's chi-square-whitened residuals — this crate's
/// `ReprojectionFactor` residual is raw pixel error). `Identity`
/// collapses to plain L2 (today's behavior, so this field is
/// backward-compatible).
```

`lightweight_vio` is not part of this workspace and means nothing to a reader of kornia's docs; this text renders into published rustdoc for a public field. The useful half of the comment is the units statement, which is worth keeping on its own terms — `ReprojectionFactor`'s residual is raw pixel error, so `robust_scale_sq` is in px², unlike the whitened-residual convention used elsewhere. `BaParams::robust` at `ba.rs:236` is a good model for the phrasing.

Also minor, same file: the two adjacent `use kornia_algebra::optim::{...}` statements introduced by the PR can be merged into one.

## ✅ Expected behaviour

Public docs and test names in `main` should not carry authoring artifacts or point at codebases outside the workspace.

## 📎 Context

Follow-up to #1111. Companion issues: #1117 (NaN loss parameter), #1118 (shared step tolerance), #1119 (Tukey inconsistency).

VedantMadane:
## Summary

Clean up #1111 scaffolding: TEMP test name/comment and lightweight_vio reference in public rustdoc

## Changes

- crates\kornia-3d\src\pnp\refine.rs: rename TEMP test; clean scaffolding comments

Fixes #1120

### DIFF
--- a/COMMITMSG
+++ b/COMMITMSG
@@ -1,0 +1,7 @@
+fix: Clean up #1111 scaffolding: TEMP test name/comment and lightweight_vio
+
+- crates\kornia-3d\src\pnp\refine.rs: rename TEMP test; clean scaffolding comments
+
+Fixes #1120
+
+Signed-off-by: Vedant Madane <6527493+VedantMadane@users.noreply.github.com>
--- a/crates/kornia-3d/src/pnp/refine.rs
+++ b/crates/kornia-3d/src/pnp/refine.rs
@@ -32,7 +32,7 @@ pub struct LMRefineParams {
     /// Initial damping factor (lambda).
     pub initial_lambda: f32,
     /// M-estimator kernel applied per-residual (pixel-space, unlike
-    /// lightweight_vio's chi-square-whitened residuals — this crate's
+    /// pnp refine's chi-square-whitened residuals — this crate's
     /// `ReprojectionFactor` residual is raw pixel error). `Identity`
     /// collapses to plain L2 (today's behavior, so this field is
     /// backward-compatible).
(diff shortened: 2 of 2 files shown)


COMMITMSG
edgarriba, line 1: This file should go out

edgarriba (approved):

PR #1126, opened 2026-08-26
feat(kornia-3d): StereoRectifier rectifies through remap_u8 on CPU and CUDA

edgarriba:
## Summary

`StereoRectifier` now has one sampler for every backend: `rectify_left/right` route through kornia-imgproc's residency-dispatched `remap_u8` (CPU↔CUDA byte-exact by its tested contract), and a new `cuda` feature adds `to_cuda(&stream) → CudaStereoRectifier` with device-resident maps.

Design (from a three-agent design review):
- **Dispatch stays in `remap_u8`** — kornia-3d holds maps in both residencies and forwards; zero kernel/launch/unsafe code in this crate.
- **Maps stored planar** (`Image<f32,1>` x/y pairs) — what every consumer actually wants; the #1125 interleaved accessors become `left_maps()/right_maps()`.
- **Warm-up in `to_cuda`**: kernels compile lazily via nvrtc, so without a warm-up a broken CUDA stack passes construction and fails on frame 1 with the caller's CPU fallback already forfeited. The constructor runs one full rectify.
- **`_device` primitives + `_into` conveniences**: zero-copy async for device-resident pipelines; H2D/D2H with retained scratch (no per-frame allocation) for host-bytes drivers.
- **No implicit anything**: no context creation, no lazy uploads, typed errors on mid-run failures — demote-to-CPU policy belongs to the caller, and is now SAFE because the bytes match.

## Behaviour change (deliberate)

Rectified bytes shift: the `[w-1, w)` border band (formerly black) now clamp-samples, and blending is Q10 fixed point (±1 LSB anywhere). In exchange **CPU and CUDA rectification are identical**, so live-GPU and replay-CPU pipelines produce the same pixels — previously they could not.

## Tests

- All existing kornia-3d tests green with the remap_u8 CPU path (row-alignment, MH01 baseline, size mismatch).
- New `cuda_rectify_matches_cpu_byte_exact` — passes on Jetson Orin (Vulkan/CUDA 8.7).
- Default build stays cudarc-free; feature is additive.

First consumer: flux-xlerobot's OAK source deletes its hand-rolled 107-line CudaRectifier (13.7× measured on Orin, 4.09→0.30 ms/eye).

### DIFF
--- a/crates/kornia-3d/Cargo.toml
+++ b/crates/kornia-3d/Cargo.toml
@@ -32,7 +32,14 @@ kiddo = { workspace = true }
 kornia-algebra = { workspace = true }
 kornia-image = { workspace = true }
 kornia-imgproc = { workspace = true }
-cudarc = { workspace = true, optional = true }
+cudarc = { version = "0.19.0", default-features = false, features = [
+  "cuda-version-from-build-system",
+  "driver",
+  "fallback-dynamic-loading",
+  "fallback-latest",
+  "nvrtc",
+  "std",
+], optional = true }
 log = { workspace = true }
 nalgebra = { workspace = true }
 rand = { workspace = true }
--- a/crates/kornia-3d/Cargo.toml
+++ b/crates/kornia-3d/Cargo.toml
@@ -21,14 +21,7 @@ name = "bench_twoview"
 
 [dependencies]
 bincode = { workspace = true }
-cudarc = { version = "0.19.0", default-features = false, features = [
-  "cuda-version-from-build-system",
-  "driver",
-  "fallback-dynamic-loading",
-  "fallback-latest",
-  "nvrtc",
-  "std",
-], optional = true }
+cudarc = { workspace = true, optional = true }
 faer = { workspace = true }
 kiddo = { workspace = true }
 kornia-algebra = { workspace = true }
--- a/crates/kornia-3d/Cargo.toml
+++ b/crates/kornia-3d/Cargo.toml
@@ -19,19 +19,8 @@ name = "bench_linalg"
 harness = false
 name = "bench_twoview"
 
-[features]
-# CUDA-resident rectification: uploads the remap tables once and serves device frames through
-# kornia-imgproc's residency-dispatched remap (byte-exact with the CPU path by its tested
-# contract). cudarc appears only for the stream type in `to_cuda`'s signature.
-cuda = ["dep:cudarc", "kornia-imgproc/cuda", "kornia-image/cuda"]
-
 [dependencies]
 bincode = { workspace = true }
-faer = { workspace = true }
-kiddo = { workspace = true }
-kornia-algebra = { workspace = true }
-kornia-image = { workspace = true }
-kornia-imgproc = { workspace = true }
 cudarc = { version = "0.19.0", default-features = false, features = [
   "cuda-version-from-build-system",
   "driver",
@@ -40,6 +29,11 @@ cudarc = { version = "0.19.0", default-features = false, features = [
   "nvrtc",
   "std",
 ], optional = true }
+faer = { workspace = true }
+kiddo = { workspace = true }
+kornia-algebra = { workspace = true }
+kornia-image = { workspace = true }
+kornia-imgproc = { workspace = true }
 log = { workspace = true }
 nalgebra = { workspace = true }
 rand = { workspace = true }
@@ -52,3 +46,9 @@ approx = { workspace = true }
 criterion = { workspace = true }
 kornia-io = { workspace = true }
 kornia-tensor = { workspace = true }
+
+[features]
+# CUDA-resident rectification: uploads the remap tables once and serves device frames through
+# kornia-imgproc's residency-dispatched remap (byte-exact with the CPU path by its tested
+# contract). cudarc appears only for the stream type in `to_cuda`'s signature.
+cuda = ["dep:cudarc", "kornia-image/cuda", "kornia-imgproc/cuda"]
--- a/crates/kornia-3d/Cargo.toml
+++ b/crates/kornia-3d/Cargo.toml
@@ -19,6 +19,12 @@ name = "bench_linalg"
 harness = false
 name = "bench_twoview"
 
+[features]
+# CUDA-resident rectification: uploads the remap tables once and serves device frames through
+# kornia-imgproc's residency-dispatched remap (byte-exact with the CPU path by its tested
+# contract). cudarc appears only for the stream type in `to_cuda`'s signature.
+cuda = ["dep:cudarc", "kornia-imgproc/cuda", "kornia-image/cuda"]
+
 [dependencies]
 bincode = { workspace = true }
 faer = { workspace = true }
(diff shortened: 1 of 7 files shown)


crates/kornia-3d/src/stereo/rectify.rs
```diff
@@ -235,33 +237,229 @@ impl StereoRectifier {
...
-        // `out` starts black; only pixels whose source lands inside the raw
-        // image get written, so out-of-frame borders stay black.
-        let mut out = vec![0u8; self.width * self.height];
-        for (dst, &[fx, fy]) in out.iter_mut().zip(map.iter()) {
-            if let Some(v) = sample_bilinear(src, sw, sh, fx, fy) {
-                *dst = v;
-            }
-        }
-        Ok(Image::from_size_slice(
+        let mut out = Image::from_size_val(
```
edgarriba: That should be part of the signature

crates/kornia-3d/src/stereo/rectify.rs
```diff
@@ -15,6 +15,10 @@
 use crate::camera::PinholeCamera;
 use kornia_algebra::{Mat3F64, Vec3F64, SO3F64};
 use kornia_image::{Image, ImageError, ImageSize};
+use kornia_imgproc::interpolation::{remap_u8, InterpolationMode};
+
+#[cfg(feature = "cuda")]
+use std::sync::Arc;
```
edgarriba: Why cuda feature here ?

edgarriba:
Addressed all four review comments in the latest push: rectify_left/right are into-style (dst in the signature), the CUDA half moved into a cfg'd submodule (no feature-gated top-level imports), the raw-slice length error is now InvalidImageSize, and the _device methods check residency explicitly. 224+14 tests green including the CUDA byte-exact test on Orin. Note for downstream: flux-vio/cufx-source migrate to into-style at repin.

edgarriba:
Pushed e6d417b1 — second review round (adversarial, compiler-verified) found two real blockers plus three medium issues, all fixed:

- **`CudaStereoRectifier` was unnameable downstream** — re-exported from the *private* `rectify` module only, never from `stereo/mod.rs`, so `to_cuda` returned a type you could not put in a struct field. Now re-exported, and the CI clippy matrix gains `-p kornia-3d --features cuda` so a dangling cfg'd export can never pass again (rustc had been flagging it as an unused import all along — nothing compiled that path in CI).
- **Default clippy job failed** on `[&*src, dst]` (`borrow-deref-ref` under `-D warnings`); `[src, dst]` coerces identically.
- Wrong-length raw slices now return a typed `StereoError::RawLengthMismatch` instead of a transposed `InvalidImageSize` — matters for demote-to-CPU policies that classify caller bugs vs CUDA failures.
- `check()` now really guards cross-device use: it compares the frame's context ordinal against the retained stream (the old comment claimed remap_u8 caught this; it only compares src-vs-dst, so multi-GPU misuse was an illegal memory access).
- The duplicated geometry fields/accessors collapsed into one shared `RectifiedGeometry` copied whole in `to_cuda` — the CPU and CUDA rectifiers can no longer drift.

Plus: warm-up runs on the retained device scratch (no host Vecs; its synchronize still fences the map uploads, now documented as load-bearing), `_device` methods take `&self`, `build_map` moves its Vecs instead of copying 5.8 MB, cudarc pinned to the minimal feature set kornia-image uses, umbrella `kornia`'s `cuda` feature now forwards here, and CHANGELOG records the feature + the into-style break.

Verified on Orin: `cargo test -p kornia-3d --lib --features cuda` — 224 passed, incl. `cuda_rectify_matches_cpu_byte_exact`; clippy clean under both default and cuda feature sets.

crates/kornia-3d/Cargo.toml
```diff
@@ -19,13 +19,27 @@ name = "bench_linalg"
...
+cuda = ["dep:cudarc", "kornia-imgproc/cuda", "kornia-image/cuda"]
+
 [dependencies]
 bincode = { workspace = true }
 faer = { workspace = true }
 kiddo = { workspace = true }
 kornia-algebra = { workspace = true }
 kornia-image = { workspace = true }
 kornia-imgproc = { workspace = true }
+cudarc = { version = "0.19.0", default-features = false, features = [
```
edgarriba: Use workspace

crates/kornia-3d/src/stereo/rectify.rs
```diff
@@ -154,114 +192,320 @@ impl StereoRectifier {
...
+            // for correctness — it fences the four async map uploads above, so the maps are
+            // quiescent for every later call.
+            remap_u8(
+                &dev.scratch_in,
+                &mut dev.scratch_out,
+                &dev.left_map_x,
+                &dev.left_map_y,
+                InterpolationMode::Bilinear,
+            )?;
+            stream
```
edgarriba: Synchronisation should be called by users. Add a comment for the user or agent to known it

crates/kornia-3d/src/stereo/rectify.rs
```diff
@@ -154,114 +192,320 @@ impl StereoRectifier {
...
+                &self.right_map_x,
+                &self.right_map_y,
+                InterpolationMode::Bilinear,
+            )?;
+            Ok(())
+        }
+
+        /// Host-bytes convenience for the driver loop: H2D into retained scratch, kernel, blocking
+        /// D2H into `out` (resized once). No per-frame allocation.
+        pub fn rectify_left_into(
```
edgarriba: This method not needed

crates/kornia-3d/src/stereo/rectify.rs
```diff
@@ -154,114 +192,320 @@ impl StereoRectifier {
...
+            &mut self,
+            raw: &[u8],
+            out: &mut Vec<u8>,
+        ) -> Result<(), StereoError> {
+            self.host_roundtrip(raw, out, true)
+        }
+
+        /// See [`rectify_left_into`](Self::rectify_left_into).
+        pub fn rectify_right_into(
```
edgarriba: This method not needed

crates/kornia-3d/src/stereo/rectify.rs
```diff
@@ -154,114 +192,320 @@ impl StereoRectifier {
...
+            &mut self,
+            raw: &[u8],
+            out: &mut Vec<u8>,
+        ) -> Result<(), StereoError> {
+            self.host_roundtrip(raw, out, false)
+        }
+
+        fn host_roundtrip(
```
edgarriba: This method not needed

edgarriba:
Addressed the round-3 comments in the latest push:

- **`rectify_*_into` / `host_roundtrip` deleted** — the CUDA rectifier is device-only now; host staging is the caller's business. The retained scratch and the raw-length error variant went with them, and the byte-exact test drives `rectify_*_device` directly (still passing on Orin: 224 tests green).
- **No library-side synchronize** — `to_cuda` enqueues the map uploads and warm-up launch on the given stream and returns; the doc states that synchronization belongs to the caller, and names the one ordering hazard (using the rectifier from a *different* stream before syncing the upload stream). nvrtc compile failures still surface synchronously at construction, so the CPU-fallback story is unchanged.
- **`cudarc = { workspace = true }`** restored in kornia-3d's Cargo.toml.

cjpurackal (approved):

Issue #915, opened 2026-05-23
[Feature]: Port kornia-algebra to PyO3 (Python bindings for matrices/vectors/Lie groups)


edgarriba:
### 🚀 Feature Description

Port `kornia-algebra` to PyO3 inside `kornia-py` so downstream Python users (and Python bindings of higher-level crates like `kornia-slam`) can construct, inspect, and convert the foundational linear-algebra types — `Mat2/3/4F32/F64`, `Vec2/3/4F32/F64`, `Quat*`, and the Lie group types (`SO2/3`, `SE2/3`, `RxSO3`, `Sim3`) — without each downstream crate having to reimplement the numpy<->Rust marshalling.

### 📂 Feature Category

Python Bindings (kornia-py)

### 💡 Motivation

`kornia-py` already exposes a wide PyO3 surface for image ops, two-view geometry, PnP, BA, etc., but `kornia-algebra` itself isn't bound. That means every downstream Python binding that touches a `Pose3d`, `Mat3F64`, or `Vec3F64` has to do its own ad-hoc conversion to/from numpy.

Concrete example: in the brand-new [`kornia-slam-py`](https://github.com/kornia/kornia-slam/pull/34) bindings (PyO3 for ORB-SLAM), I had to hand-roll:
- `Pose3d -> ndarray (4,4) f64` (extracting `r.x_axis/y_axis/z_axis` columns + translation)
- `PinholeCamera::intrinsic_matrix` -> ndarray (3,3)
- The skeletons of `PinholeCamera`/`Pose3d` as `#[pyclass]` wrappers

All of this would be downstream of a single canonical Python representation if `kornia-algebra` types were already bound. As more crates grow Python bindings (slam, sensors, future SfM/MVS work), each one will rewrite the same marshalling unless we centralize it.

### 💭 Proposed Solution

Add a new module — e.g. `kornia-py/src/algebra.rs` — that exposes the most-used algebra types as PyO3 classes with idiomatic numpy interop:

- **Vectors:** `Vec2/3/4F32/F64` — constructed from a tuple/list or 1-D numpy array; `.as_numpy()` returns a fresh `ndarray` of the matching dtype/shape.
- **Matrices:** `Mat2/3/4F32/F64`, `DMatF32/F64` — constructed from a 2-D numpy array, with `.as_numpy()` returning a `(rows, cols)` ndarray.
[...]

PR #1138, opened 2026-09-18
feat(py): add Vec3F32 and Vec3F64 bindings

Hrithikcrick:
##  Description

Relates to #915

Adds PyO3 bindings for `Vec3F32` and `Vec3F64` with NumPy interoperability.

##  Changes Made

- Added `Vec3F32` and `Vec3F64` Python bindings
- Added `from_numpy()` and `as_numpy()`
- Added Python type stubs
- Added tests for construction, NumPy round-trip, invalid length, and dtype mismatch

##  How Was This Tested?

```text
pixi run python -m pytest kornia-py/tests/test_algebra.py -v

8 passed in 0.11s

### CODE REVIEW

**Comment by Hrithikcrick** [comment]
## What this enables

This PR exposes the existing Rust `kornia-algebra` 3D vector types to Python.

Python users can now:

- Create `Vec3F32` and `Vec3F64` objects
- Access their `x`, `y`, and `z` components
- Convert NumPy arrays into Kornia vector objects
- Convert Kornia vectors back to NumPy arrays
- Preserve the correct `float32` / `float64` dtype during conversion
- Receive clear errors for invalid vector lengths or dtype mismatches

The bindings wrap the existing `kornia-algebra` vector implementations, so the vector math is not reimplemented in Python.

File: kornia-py/tests/test_algebra.py [hunk 1]
```diff
@@ -0,0 +1,70 @@
+import numpy as np
+import pytest
+
+import kornia_rs
+
+
+@pytest.mark.parametrize(
+    ("vector_type", "dtype"),
+    [
+        (kornia_rs.Vec3F32, np.float32),
```
edgarriba: We should add a sub module kornia_rs.algebra.Vec3F32

edgarriba (approved):

kornia-py/tests/test_algebra.py
Hrithikcrick: Thanks! Updated the bindings to expose `Vec3F32` and `Vec3F64` under `kornia_rs.algebra`, and updated the tests and type stubs accordingly. The focused tests still pass.

